PR #600, opened 2014-05-07
Remove CORS headers from fullResponse (can conflict with restify.CORS)

rprieto:
The new `CORS` middleware has great support for cross-origin requests, including:

- defining a list of allowed origins
- defining a list of custom exposed headers

The `fullResponse` middleware also does some of this, but unfortunately:

- not as configurable
- sometimes conflicts with choices made by the CORS middleware

For example, it will not override `Access-Control-Allow-Origin` if it's already set ([see code](https://github.com/mcavage/node-restify/blob/master/lib/plugins/full_response.js#L35)), which is good. However the CORS module might decide **not** to add the header, for example if the request doesn't match the allowed origins ([see code](https://github.com/mcavage/node-restify/blob/master/lib/plugins/cors.js#L94)). In this case setting it as `*` later is probably not the right behaviour.

There might be ways to arrange the middlewares & configs to avoid these clashes, but I think having a single full-featured `CORS` middleware is preferable to having CORS support added in a few places. Thoughts?

### DIFF
--- a/lib/plugins/full_response.js
+++ b/lib/plugins/full_response.js
@@ -5,48 +5,12 @@ var crypto = require('crypto');
 var httpDate = require('../http_date');
 
 
-///--- Globals
-
-var ALLOW_HEADERS = [
-    'Accept',
-    'Accept-Version',
-    'Content-Length',
-    'Content-MD5',
-    'Content-Type',
-    'Date',
-    'Api-Version',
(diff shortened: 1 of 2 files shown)


DonutEspresso:
Closing this now that we're rebuilding CORS from the ground up. If you're interested, please follow over at restify/plugins#10

PR #609, opened 2014-05-23
Use matched origin for non-credential CORS.

### LINKED ISSUE #608: Cannot return `Access-Control-Allow-Origin: *` with CORS plugin
The [CORS](http://www.w3.org/TR/cors/#access-control-allow-origin-response-header) spec say that the `Access-Control-Allow-Origin` header can be `*` but the plugin doesn't support this.

We require this functionality in web APIs that are going to be loaded from multiple websites where the response itself is cached by the browser.

longhotsummer:
This means that we can return `Access-Control-Allow-Origin: *` for non-authenticated requests. Fixes #608

### DIFF
--- a/lib/plugins/cors.js
+++ b/lib/plugins/cors.js
@@ -97,9 +97,8 @@ function cors(opts) {
         }
 
         function corsOnHeader() {
-            origin = req.headers['origin'];
             if (opts.credentials) {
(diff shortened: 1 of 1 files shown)


DonutEspresso:
I've tried to incorporate some of this work in the new plugin. If you're interested, please hop over to restify/plugins#10. We're removing CORS from restify core and moving it there.

PR #624, opened 2014-06-11
Expose the Request and Response

TheThing:
With these changes, one can check if the request object or response object are instanceof.

Assuming I have a (stupid) helper function like so:

``` javascript

exports.isLocalhost = function(ip, cb, req) {
  if (ip === '127.0.0.1') {
    if (req instanceof restify.Request) {
      req.log.warn('Request was made localhost');
    }
    return true;
  }
  return false;
}
```

One could unit test the isLocalhost helper without worrying about faking the request object.


serv:
`require('restify/lib/request');` doesn't work now with Restify 4.0.4 as the module isn't exported.

PR #628, opened 2014-06-17
Support URLs that have a path component.

calebbrown:
Rather than this:

```javascript
var client = restify.createJsonClient({url: 'https://api.twitter.com'});

client.get('/1.1/statuses/user_timeline.json');
client.get('/1.1/users/show.json');
```

This Pull Requests lets you do this:

```javascript
var client = restify.createJsonClient({url: 'https://api.twitter.com/1.1/'});

client.get('/statuses/user_timeline.json');
client.get('/users/show.json');
```

It uses `path.join()` to join the two fragments so you don't need to think about slashes.

This will break existing code if people have a URL with a path set in the client.

### DIFF
--- a/lib/clients/http_client.js
+++ b/lib/clients/http_client.js
@@ -6,6 +6,7 @@ var fs = require('fs');
 var http = require('http');
 var https = require('https');
 var os = require('os');
+var path = require('path');
 var querystring = require('querystring');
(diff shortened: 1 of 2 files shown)


micahr:
Closing in favor of #843 which combines this with #633

PR #630, opened 2014-06-18
Fix issue with Server toString method

OiNutter:
Safety check for this.versions being an array, don’t try and join it if
it’s not. This addresses mcavage/node-restify#629

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -544,10 +544,6 @@ Server.prototype.toString = function toString() {
         return (s);
     }
 
-    function isArray(arr) {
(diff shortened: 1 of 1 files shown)


yunong:
@OiNutter Thanks for doing this. In order for us to merge, could you please: 
1) add unit tests, prefixing the names with 'gh #630'
2) make sure `make prepush` runs clean

micahr:
Closing in favor of #840 which includes all the changes from this PR, plus tests and doc fixes.

PR #633, opened 2014-06-23
url.path + path

nakosung:
Fixed erroneous path generation like below:
```
url[http://localhost:3000/something] + get[/foo] 
  ->  (should be)     http://localhost:3000/something/foo
  -x->(should not be) http://localhost:3000/foo
```


yunong:
@nakosung Thanks for doing this. Could please
1) add unit tests, and name them with the prefix 'gh #633'
2) make sure `make prepush` runs clean.

micahr:
Closing in favor of #843 which combines this with #628

PR #635, opened 2014-06-24
Adding req.matchedVersion()

nathanpeck:
When a route is defined as responding to multiple version numbers then if the client supplies a version range that it can accept such as `accept-version: <2.2.0` it can be useful for the route to know exactly which version number from its array of supported versions matched the clients version specification.

req.version() already exposes the clients version string but it is often a range such as <2.2.0 or 2.*
req.matchedVersion() allows a route which matches versions 2.0.0, 2.1.0, and 2.2.0 to know that a client who specified `accept-version: <2.2.0` reached the route because 2.1.0 matched the request header.

This pull request includes documentation, and a test.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -367,6 +367,7 @@ Router.prototype.find = function find(req, res, callback) {
         }
     }
 
+    var maxV;
     if (!r) {
         // If upload and typed
         if (typed) {
@@ -390,7 +391,6 @@ Router.prototype.find = function find(req, res, callback) {
         }
 
         if (versioned) {
-            var maxV;
             candidates.forEach(function (c) {
                 var k = c.r.versions;
                 var v = semver.maxSatisfying(k, req.version());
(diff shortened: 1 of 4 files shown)


yunong:
Hey @nathanpeck , thanks so much for doing this. In order to merge, could you 
1) prefix the unit test name with 'gh #635'
2) make sure `make prepush` runs clean.

lib/router.js
```diff
@@ -367,6 +367,7 @@ Router.prototype.find = function find(req, res, callback) {
         }
     }
 
+    var maxV;
```
micahr: Can you move this up to the top of the function, with the rest of the variable declarations?

test/server.test.js
```diff
@@ -1076,6 +1076,34 @@ test('gh-283 maximum available versioned route matching', function (t) {
     });
 });
 
+test('routes matching multiple versions should know what the max matched version was', function (t) {
```
micahr: please add `gh #635` to the beginning of your test case, and this should be good to go!

micahr:
@nathanpeck Almost there, just a couple small changes (and merge conflict resolutions) and this is good to go!

micahr:
Closing in favor of #842 which fixes the merge conflict and JSCS errors

PR #671, opened 2014-08-22
Add 'server.route' alias for multiple methods

Autarc:
A route can be defined which delegates to the proper methods, preventing repetitive declaration of the path. In addition got the static plugin a modification: the directory path is just relative and decoupled from the actual route path without getting prefixed.

### DIFF
--- a/docs/index.restdown
+++ b/docs/index.restdown
@@ -257,6 +257,22 @@ Routes can also accept more than one handler function. For instance:
         }
     );
 
+### Multiple Methods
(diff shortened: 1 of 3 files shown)


yunong:
This is just syntactic sugar. I appreciate your efforts, but this is not something that we're interested in picking up at this time. It adds a lot of unneeded complexity within the lib, for not too much gain. You are however, welcome to write a plugin for Restify that does this :)

PR #678, opened 2014-08-31
Adding before event as a general solution to 2.8.2 breaking socket.io 

cmawhorter:
Went ahead and created what I was suggesting in #669 

For some reason 'request' was being emitted without the handleRequest callback.  Seems like it was being emitted from outside restify maybe?

Instead, I just added a before event to match the 'after' that already exists.

Very limited testing...

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -248,9 +248,15 @@ function Server(options) {
 
     this.server.on('request', function onRequest(req, res) {
         self.emit('request', req, res);
+        var handleRequest = function() {
+            self._setupRequest(req, res);
+            self._handle(req, res);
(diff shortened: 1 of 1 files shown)


yunong:
I'm going to close this for now, as we'll be working on adding support for http2. https://github.com/mcavage/node-restify/milestones/Add%20support%20for%20http2

cmawhorter:
I'm not sure this would be resolved by implementing http2.  This pr was about restify not playing nice with other libs sharing the same server.

What this pr aimed to do was allow you to tell restify not to handle a connection and bypass the inherent problem.

PR #688, opened 2014-10-01
Fix throttle using and/or reporting incorrect rate (if overrides present) in response

brycekahle:
Throttle could use an incorrect default rate if a request utilizing
an override occurred as the first request.

Throttle could also report the wrong rate in the response message if a
request utilizing an override has occurred since start.

Also removed the outdated comment.


yunong:
@brycekahle Sorry for dropping the ball on this. Thanks for adding the tests. Could you please run a `make prepush` and ensure the linter and styler pass?

brycekahle:
@yunong updated.

yunong:
Merged, thanks @brycekahle

PR #698, opened 2014-10-22
Docs typo, replace ``` with <code>

chmac:
Guessing this was missed in the conversion from markdown to HTML.

This pull request is a single commit to fix the issue, apologies for the mess in the previous PR.


yunong:
Hi @chmac, thanks for doing this, but we'll need to fix this in the base index.restdown file. Would you mind fixing it there?

chmac:
This was fixed in 7b04a8d669a63e69a2ba34ebfd1db2eeb7df080c. :+1:

PR #702, opened 2014-11-01
Support multiple response header values.

thizzle:
Multiple response headers will be stored as an array, and passed
to the Node.js `ServerResponse`.

See http://nodejs.org/api/http.html#http_response_setheader_name_value
and
https://github.com/joyent/node/issues/2750


yunong:
merged

PR #704, opened 2014-11-07
Allows partial regex for named parameters

gchauvet:
Hi,

Please find attached my first contribution on this project. This pull request allows partial regular expression for named parameters, like this : 
/foo/:bar([0-9]+) is parsed as :
req.params.bar = value that match the pattern.
In other case, route is considered as a "missing route".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ examples/todoapp/node_modules
 *.tgz
 build
 docs/*.json
+/nbproject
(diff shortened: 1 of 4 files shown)


yunong:
Hi, thanks so much for doing this. Could you please run `make prepush` and ensure lint and style runs clean. Also, if could you please prefix your unit tests with 'gh #704'?

gchauvet:
See #763 (because this merge point out a "unknown repository"

gchauvet:
See #764

PR #719, opened 2014-12-07
Update static.js to allow for serving static files that do not use the route as a path

Wavewash:
When using restify moving from express, I found sending a static file of which I knew the direct path to difficult when I was trying to map a route to it which did not contain the filename in it.

For example in express:

app.get('/cool', function (req, res) {
  res.sendFile("../public/coolindex.html");
}

the proposed change:
By sending a "file" option now with the serveStatic call you can serve a file which does not have its' filename present in the req.path there by mapping routes to static files.

Example usage:
app.get('/logins',
    restify.serveStatic({
      directory: './public',
      file: 'login.html'
    })
  );

### DIFF
--- a/lib/plugins/static.js
+++ b/lib/plugins/static.js
@@ -25,6 +25,7 @@ function serveStatic(opts) {
     assert.optionalNumber(opts.maxAge, 'options.maxAge');
     assert.optionalObject(opts.match, 'options.match');
     assert.optionalString(opts.charSet, 'options.charSet');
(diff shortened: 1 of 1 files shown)


yunong:
Hi, thanks for doing this. Could you please add a test for this? Prefix the test name with 'gh #719'. Also, please run a `make prepush` to ensure the linter and styler checks pass.

micahr:
@Wavewash This change looks fairly small. Could you provide a quick test case to validate the feature?

PR #722, opened 2014-12-10
Normalise route names when chaining routes

cysp:
Route names appear to be normalised when they're added to the router but during route chaining the route name requested is used for the lookup verbatim.

This has the upshot that it is non-obvious how to chain to a route named `'test-route'`.


yunong:
Thanks. @pfmooney this looks good, could you make sure it prepushes clean and merge?

PR #726, opened 2014-12-18
Enable {agent: false} option override per request

g-murray-sreams:
Allow override of agent option to ```false``` for each request. 

Currently you can override an agent to true, or to an Agent object, but overriding to ```false``` will shortcut (via ```||```) to whatever the global setting is (```true```, ```false```, ```Agent```, or even ```undefined```).

In short, this pull request will make a per request ```false``` setting override the global setting.

### DIFF
--- a/lib/clients/http_client.js
+++ b/lib/clients/http_client.js
@@ -514,7 +514,7 @@ HttpClient.prototype._options = function (method, options) {
 
     var self = this;
     var opts = {
-        agent: options.agent || self.agent,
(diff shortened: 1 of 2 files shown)


yunong:
Looks good. Slightly nitty, could you please prefix the test name with the PR #?. 

e.g.
'PR-726 Enable {agent: false} option override per request '

g-murray-sreams:
No worries, done.

PR #729, opened 2014-12-22
Added typeof check to the jsonp callback

mbjordan:
When getting data via a jsonp request, its useful to have a check that ensures the callback function is defined before calling it. This prevents a reference error from being thrown in the browser if a function name is not defined when data is returned.


yunong:
Hi, thanks for contributing! Could you please add a unit test for this? Please prefix the name with "gh #729".

mbjordan:
The test has been added, however the tests failed due to using version 3.0.0 on Travis and my PR is from 2.8.4.

Thinking I should close this PR and open a new one with the most recent version.

PR #741, opened 2015-01-21
Add ability to listen for error events

yunong:
This is especially helpful for server 5XX or 4XX error pages. You can listen for say, `InternalServerError`, and then decorate that error with information globally, without having to do that every time the error is served.

### DIFF
--- a/docs/index.restdown
+++ b/docs/index.restdown
@@ -473,6 +473,23 @@ Alternatively, restify 2.1 supports a `next.ifError` API:
       });
     });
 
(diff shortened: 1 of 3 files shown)


vickvu:
@mcavage @yunong 
This is a breaking change for existing events (i.e. 'NotFound' ,'MethodNotAllowed', 'VersionNotAllowed' and 'UnsupportedMediaType') which expect the listener to have the signature of
`function (request, response, cb) {}`

PR #744, opened 2015-01-29
fix typo in Request.isUpload

jcreigno:
There is an obvious typo on the patch method in `Request.isUpload`.

This pull request fix it and add tests on the method.


pfmooney:
Would you be opposed to implementing the `request.isUpload` validity tests as part of the server.test.js suite?  (Testing the result in a more integrated setting rather than calling the function directly.)

jcreigno:
Would you prefer that I change the '<method> ok' tests cases or add a new one dedicated to `isUpload` ?

pfmooney:
@jcreigno Either way should be fine.

jcreigno:
I've changed the 'ok' tests cases. I removed the assertion on OPTIONS as I'm not sure it is valid there.

pfmooney:
Merged in 614a34731c9b17a46c0513fec91b7f3bb9d3a043.
Thanks

PR #747, opened 2015-02-03
Call the next() function at the right place, remove the redundant 'utf8' encoding.

lygstate:


yunong:
Hi, thanks for doing this, could you give me some context about this PR?

lib/plugins/multipart_parser.js
```diff
@@ -71,25 +72,31 @@ function multipartBodyParser(options) {
...
-                            'utf8',
                             function (ex, data) {
+                                i = i + 1;
                                 if (ex) {
                                     return (false);
                                 }
                                 req.params[f] = data;
-                                return (true);
+                                if (i === length) {
+                                  return (next());
```
micahr: I believe next could be called multiple times if you move it in here.

`i === length` will be subject to a race condition depending on the speed of the filesystem. If `files` is an array of 10, and you read all 10 files, you will be checking a `length` of 10, to 1 - 10, but conceivably you could read the first file, before the next forEach loop is run, which would be length === 1 and i === 1.

If you are looking for a way to ensure that you call next after all files have been read, then the check should be more resilient to race conditions.

yunong:
no longer an issue since it's fixed as part of #847

Issue #751, opened 2015-02-14
Server creation fails (bad server.url) on Travis CI for NodeJS v11 and io.js v1.2.0


ArminTamzarian:
Myy project [node-misfit](https://github.com/ArminTamzarian/node-misfit) uses node-restify to build a [mock service](https://github.com/ArminTamzarian/node-misfit/blob/master/test/mock-api.misfitwearables.com.js) for unit testing. While running just fine locally (OSX - NodeJS v0.10.29) when I push the current in progress code to Github Travis CI tests the build with what appears to be an invalid URL detected.

Specifically my code for creating the server is as follows:
```javascript
var mock = function (options, callback) {
          var server = restify.createServer();
          server.use(restify.bodyParser());

          // routing logic...

          server.listen(options.port, function() {
            console.log(server); // for debugging this issue
            console.log(server.url); // for debugging this issue
            callback();
        });
```

While I am testing on NodeJS v0.11.x, v0.10.x and io.js v1.2.x my tests only pass on NodeJS v0.10.x (which makes since as the current version of io.js was forked from the NodeJS v0.11 branch). On startup I am outputting the URL created by the server startup, and the results are as follows:

* **NodeJS v0.11** - server.url -> `http://:::8000`
* **NodeJS v0.10** - server.url -> `http://0.0.0.0:8000`
* **io.js v1.2** - server.url -> `http://:::8000`

For more information see the Travis CI output for the referenced build: https://travis-ci.org/ArminTamzarian/node-misfit/builds/50702849

So, I'm not sure if there is an issue between restify and the current configuration of Travis CI or if there is a bug between restify and the NodeJS v0.11 branch (or if I am doing something egregiously wrong). Any assistance would be appreciated.

ArminTamzarian:
No sooner did I report this than I installed nvm and tested locally with v0.11. Same issue as Travis CI, but I drilled down into the issue and found that the call to `server.listen(port, callback)` tried to bind to an nonexistent IPv6 address `:::` by default whereas on v0.10 the same call binds to the IPv4 address `0.0.0.0` (all addresses).

[...]

PR #752, opened 2015-02-16
Addresses URL property for servers created with IPv6 addresses

ArminTamzarian:
Fix for #751 with associates unit tests.


ArminTamzarian:
Fair enough comments- was assuming that anything beyond IPv6 would have the same addressing scheme. However, since IPv8 shouldn't be an issue for the next few millennia I'll take your suggestion as well as refactoring tests within server.tests.js.

pfmooney:
Merged in 457a0d27aeed163f49a6653ffa258aa4177dd5c9 (squashed and with style fixes).

Thanks

Issue #691, opened 2014-10-02
Posting with text/csv content type crashes Restify.


nathanpeck:
A request with the content type set to `text/csv` causes a fatal crash.

Example:

```
Uncaught exception on request POST /foo/bar HTTP/1.1
content-type: text/csv
host: localhost:8090
connection: close
user-agent: Paw 2.0.9 (Macintosh; Mac OS X 10.9.5; en_US)
content-length: 63454
accept-version: 3.8.0

TypeError: object is not a function
TypeError: object is not a function
    at Server.parseBody (/Users/nathanpeck/Documents/Code/StoryDesk/api/node_modules/restify/lib/plugins/body_parser.js:78:13)
    at next (/Users/nathanpeck/Documents/Code/StoryDesk/api/node_modules/restify/lib/server.js:730:30)
    at f (/Users/nathanpeck/Documents/Code/StoryDesk/api/node_modules/restify/node_modules/once/once.js:17:25)
    at IncomingMessage.done (/Users/nathanpeck/Documents/Code/StoryDesk/api/node_modules/restify/lib/plugins/body_reader.js:78:13)
    at IncomingMessage.g (events.js:180:16)
    at IncomingMessage.emit (events.js:92:17)
    at _stream_readable.js:938:16
    at process._tickDomainCallback (node.js:463:13)
```

If post the content with a random content type such as `foo/bar` then the request succeeds and my route handler is called as expected with the csv content passed in as a buffer.

But if I attempt the same request while using the proper content type then Restify crashes.

micahr:
This issue was fixed in #756 which was merged in February.

PR #756, opened 2015-02-23
Fix issue #691

zacharyjsnow:
Body parser expects this function to return a handler function, rather than an array of handlers.  Currently, requests with `Content-Type: text/csv` will crash the server unless there is a handler for uncaught exceptions.


yunong:
Hi, could you please add a unit test for your changes? Please prefix the test name is "#691"

zacharyjsnow:
It turns out this escaped detection because its unit tests did not properly check that the was no error returned.

PR #763, opened 2015-02-26
Allows partial regex for named parameters

gchauvet:
I must to re-create a merge request for #704 because my github repository was deleted.... my apologies


gchauvet:
Welll, I don't understand why they have 22 modified files.... I cancel this merge

Issue #769, opened 2015-03-10
Accept header being ignored


kevinpeno:
I noticed something affected by #563 and #564 today. While those changes do decide the default content-type based on the server's q-level, it completely ignores the client's accept header in the ordering.

Currently the changes above cause the content-type to be the servers favorite choice. The server is also validating that it *can* respond with what's in the accept header, but the two are disconnected from what I can tell (validated in `restify.acceptParser` and acted on in `Response.format`). This forces the routes to do content negotiation in order to supply the client's preferred output.

What I believe the server should do is to send the client its highest choice even if it's not the server's favorite choice.

Example 1: Client wants an SVG, the server has support for SVG and json on GET /image/:id
Client's Accept: `image/xml+svg`
Server's formatters: `application/json, image/xml+svg`
Result: `application/json`
Expected: `image/xml+svg`

The second example is not directly related to the problem above, but the code in the response module forces `application/octet-stream` if a formatter isn't defined. This shouldn't happen because no formatter should return 406, but also `*/*` should resolve to server's favorite choice.

Example 2: Client accepts anything 
Client: `*/*`
Server: `application/json, image/xml+svg, application/octet-stream; q=0.2`
Result: `application/octet-stream`
Expected: `application/json`

I'm not sure the best way to manage this in the system, but if you'd provide some insight I'd be happy to supply the PR. Thanks!

yunong:
@kevinpeno For example 1, we should definitely be honouring the client accept headers. That should be fixed in the response.js file. If you could send a PR that includes a test that'd be great.

For example 2, I would imagine application/octet-stream is the default for the server -- I'm not sure whether @mcavage intended that or whether we should be using the server's favourite. @mcavage care to weigh in on ex 2?

kevinpeno:
@yunong thanks for the response. I think I mixed up my errors locally and accidentally thought there was an issue causing Example 1 when Example 2 was actually causing it.

For Example 2, it appears as though the resulting Content-Type will actually be the highest q-value. However, formatters that don't have a q-value specified are not considered (even though they should default to 1). See PR #771

micahr:
Closing, as the initial bug has been fixed in #771

PR #771, opened 2015-03-10
q-value choice on wildcards ignores default q-value of 1

kevinpeno:
See #769 

Server responds with highest priority q-value when wildcards are used, but ignores setting the default q-value to 1 for mime-types without q-value specified.

Example:
Client: `*/*`
Server: `application/json, text/plain; q=0.9, application/octet-stream; q=0.2`
Result: `text/plain`
Expected: `application/json`


kevinpeno:
Locally this test fails every time. Not sure why travis is giving a green light.

yunong:
@kevinpeno For some reason I only see the tests as part of this commit, and not the actual restify changes. Are we missing files from the commit? Also, could you please prefix the test with the issue #? e.g. '#771 qvalue...'

kevinpeno:
@yunong I didn't include a fix per our convo (not sure where to apply it). Any pointers on ideas/where to apply the fix is welcome. I'll update to add the issue number to the test.

kevinpeno:
Unfortunately I run windows at home and am having issues running `npm test` so I'll take a look at both a fix and fixing the test name tomorrow.

kevinpeno:
@yunong updated the test per request and added code to make the test pass. I made `q-value` default to 1 per RFC 2616 sec 14 (emphasis added):

> Each media-range MAY be followed by one or more accept-params, beginning with the "q" parameter for indicating a relative quality factor. The first "q" parameter (if any) separates the media-range parameter(s) from the accept-params. Quality factors allow the user or user agent to indicate the relative degree of preference for that media-range, using the qvalue scale from 0 to 1 (section 3.9). *The default value is q=1.*

This is very likely a breaking change as people may have used this bug as a work around for `q-value` ordering

micahr:
Thank you for the PR @kevinpeno . Can you fix up the merge conflicts so this can be integrated?

kevinpeno:
Sure thing. I'll work on that today. Sorry for the delay!

kevinpeno:
@micahr All merged.

PR #775, opened 2015-03-14
Decode utf8 strings as whole, not in parts, fixes #774

### LINKED ISSUE #774: utf8 corruption in body parser
The body reader gets chunks of binary data, and tries to convert each chunk to string (via utf8) encoding, and then concatenate them. This doesn't work when a utf8 code point is split between binary chunks.

Testcase:

    var restify = require('./lib/index.js');
    var server = restify.createServer({  name: 'myapp', });
    server.use(restify.bodyParser());
    server.post("/", function (req, res, next) {
      if (/\ufffd/.test(req.body.text)) {
        console.log("got corruption")
      }
      res.json({ len: req.body.text.length })
      return next();
    });
    
    server.listen(1234, function () {
    });
    
    // Client
    var client = restify.createJsonClient({
      url: 'http://localhost:1234',
    });
    
    // create a long string of unicode characters
    var t = ""
    for (var i = 0; i < 100000; ++i) t += "\u2661"
    
    client.post('/', { text: t }, function (err, req, res, obj) {
      console.log('POST    /   returned: %j', obj);
    });
    

The server gets 100004 characters in the string instead of 100000.

Pull request to follow shortly.

mmoskal:


yunong:
Thanks so much for the PR. Could you please write a test for this? Prefix the test name with 'GH-774'.

mmoskal:
Test case added.

lib/plugins/body_reader.js
```diff
@@ -20,22 +20,27 @@ var MD5_MSG = 'Content-MD5 \'%s\' didn\'t match \'%s\'';
...
-        return (function (chunk) {
-            req.body += chunk.toString('utf8');
-        });
+        isText = true;
     }
 
     req.body = new Buffer(0);
     return (function (chunk) {
-        req.body = Buffer.concat([req.body, chunk]);
+        if (chunk)
```
yunong: Please add braces around this if block -- since you've got braces around the else block

lib/plugins/body_reader.js
```diff
@@ -58,6 +63,8 @@ function bodyReader(options) {
         var bodyWriter = createBodyWriter(req);
 
         function done() {
+            bodyWriter(null); // finish
```
yunong: Why are you ending the bodyWriter here?

mmoskal:
Any news on integrating this? I see someone else is also hitting this.

Thanks!
Michal

PR #784, opened 2015-04-02
Enable use of a reviver function to control how a request body is parsed...

sturnquest:
Allow more control over how request body's are transformed when parsed by specifying a reviver function in the options.

### DIFF
--- a/test/json_body_parser.test.js
+++ b/test/json_body_parser.test.js
@@ -1,5 +1,5 @@
-var restify = require('../../../lib');
-var helper = require('../helper.js');
+var restify = require('../lib');
+var helper = require('./lib/helper.js');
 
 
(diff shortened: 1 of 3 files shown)


test/json_body_parser.test.js
```diff
@@ -0,0 +1,70 @@
...
+        params: {}
+    };
+
+    parser(req, null, function () {
+        t.equal(req.params.apple, 'red');
+        t.equal(req.params.orange, 'orange');
+        t.equal(req.params.banana, 'yellow');
+        t.end();
+    });
+});
```
yunong: Hi, thanks for the PR. could you please add a \n at the end of this file? Also, please prefix your tests with 'gh-784'.

PR #792, opened 2015-04-21
Fix #791 body_reader.js#createBodyWriter() may corrupt on multi-bytes character

### LINKED ISSUE #791: body_reader.js#createBodyWriter() may corrupt on multi-bytes character
Multi-bytes character like Chinese use two or three or four bytes to encode. When we post request with large body, underly TCP may transfer it over multiple packets.

When TCP split packet in the middle of a multi-bytes character, following code may broken and corrupt the output:
```javascript
//@@ -27,9 +27,9 @@ function createBodyWriter(req) {
30        req.body = '';
31        return (function (chunk) {
32            req.body += chunk.toString('utf8');
33        });
```

Fix:
Use `Buffer` and concat them, then on request done convert at once may fix this issue.

soplwang:
Use Buffer and concat them, then on request done convert at once to fix
this issue.

### DIFF
--- a/lib/plugins/body_reader.js
+++ b/lib/plugins/body_reader.js
@@ -27,9 +27,9 @@ function createBodyWriter(req) {
         contentType === 'multipart/form-data' ||
         contentType.substr(0, 5) === 'text/') {
 
-        req.body = '';
-        return (function (chunk) {
-            req.body += chunk.toString('utf8');
+        req.body = new Buffer(0);
+        return (function ub(chunk) {
+            req.body = Buffer.concat([req.body, chunk]);
         });
     }
 
@@ -75,6 +75,9 @@ function bodyReader(options) {
                 return;
             }
 
+            if (bodyWriter.name === 'ub') {
+                req.body = req.body.toString('utf8');
+            }
             next();
         }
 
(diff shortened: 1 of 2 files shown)


lib/plugins/body_reader.js
```diff
@@ -75,6 +75,9 @@ function bodyReader(options) {
                 return;
             }
 
+            if (bodyWriter.name === 'ub') {
```
yunong: Why are you checking for the name of the bodywriter here?

test/plugins.test.js
```diff
@@ -400,6 +400,33 @@ test('body json ok (no params)', function (t) {
     });
 });
 
+test('body json ok (large)', function (t) {
```
yunong: Could you prefix this test name with GH-792?

soplwang: OK

lib/plugins/body_reader.js
soplwang: I want to minimize the changes and, align the text chunk handling with the other hand binary body's. I choose use bodywriter's name to distinct these two mode buffers.

At here, with check of bodywriter's name, I can know the buffer type. If buffer is actual utf8 mode text, then convert it to 'String' to compatible with intent behavior.

soplwang:
dup with #792, but has minimal changes

yunong:
Closing as dupe of #775

PR #796, opened 2015-04-30
Hypermedia API fails when paths have multiple patterns with sub-regexs

fangel:
Hi,

I was trying to use the Hypermedia API to render all the internal links in my API, when suddenly some of the paths was rendered wrongly - the latter parameters was simply left out of the rendered path.

So if I had a path like

    /:param1([0-9]+)/:param2([0-9]+)

Then when I did a render with `param1` to 1, and `param2` to 2, the rendered path was only `/1`, and not `/1/2` as expected.

I tracked the problem down to how the different params are identified using a regular expression. The pattern was too greedy, so faced with two sets of parameters, it would simply match all the way from the start of the first to the end of the latter parenthesis. So the regex would have the following captures

    match #     group 0                              group 1      group 2
    1           /:param1([0-9]+)/:param2([0-9]+)     param1       ([0-9]+)/:param2([0-9]+)

What I expected was that it had the following captures

    match #     group 0              group 1      group 2
    1           /:param1([0-9]+)     param1       ([0-9]+)
    2           /:param2([0-9]+)     param2       ([0-9]+)

The easiest way to fix this issue was to make the sub-pattern looking for parenthesis non-greedy by changing it from

    /\/:([A-Za-z0-9_]+)(\([^\\]+\))?/g

to

    /\/:([A-Za-z0-9_]+)(\([^\\]+?\))?/g

---

I have modified the pattern, and updated the test-case for sub-regexs to check for the error. If you can find a better way to fix this bug, I will gladly update the PR. 

Kind regards
Morten

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -181,7 +181,7 @@ Router.prototype.render = function render(routeName, params, query) {
         return (null);
 
     var _path = route.spec.path;
-    var _url = _path.replace(/\/:([A-Za-z0-9_]+)(\([^\\]+\))?/g, pathItem);
+    var _url = _path.replace(/\/:([A-Za-z0-9_]+)(\([^\\]+?\))?/g, pathItem);
     var items = Object.keys(query || {}).map(queryItem);
     var queryString = items.length > 0 ? ('?' + items.join('&')) : '';
     return (_url + queryString);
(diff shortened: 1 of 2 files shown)


yunong:
@fangel This looks great. One minor thing, could you please add the test cases as a new test? Prefix the test name with 'GH-796'

fangel:
Hi @yunong. I've moved the tests into a separate test-case. I've squashed the commits, as I assume you would prefer to have this fix as a single commit.

fangel:
Wonderful, thanks you! Any idea when the next version bump is, so we can update the version of our dependency?

Issue #686, opened 2014-09-30
res.send can't send 0, false and null


yuriykashin:
Sometimes it makes sence to send data in simple raw form like _0_ or _false_, but Response.send ignores any falsy body.

DarkPark:
@yuriykashin - I agree

```js
app.get('/', function(req, res) {
    res.send(200, false);
});
```
this really gives empty body on the client which is strange
please consider changing this

DonutEspresso:
This seems to work fine now.

janopae:
This issue is a thing again, `false` and `''` get casted to null, while `0` results in an empty body. I have a rest endpoint that should just return a boolean value – I ended up using `response.sendRaw('false', { 'Content-Type': 'application/json' });` instead. But `0`, `''` and `false` shouldn't be casted in such an inexpected way.

janopae:
That 0 results in an empty body actually makes kind of sense if 0 is interpeted as a status code. But the other ones being converted to `null` don't.

PR #800, opened 2015-05-15
Allow 0, false, and null as json body #686

AlexDobeck:
0, false, and null are all valid body contents of a request, however currently only an empty body is returned in these cases.

For Example:
response.json(200, true);
- currently returns:
200 OK
true

- however false doesn't return a body:
response.json(200, false);
200 OK

Change allows false to also return:
response.json(200, false);
200 OK
false

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -227,7 +227,7 @@ Response.prototype.send = function send(code, body, headers) {
             log.trace({res: self}, 'response sent');
     }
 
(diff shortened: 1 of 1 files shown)


yunong:
Hi, thanks so much for the PR. Could you please add a test case for this? Prefix the name with gh-800.

Burnett01:
+huge support

PR #802, opened 2015-05-19
Router.render support multilevel queries

11bucks:
for example: `?sort[owner]=1&sort[date]=-1`

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -13,6 +13,8 @@ var cors = require('./plugins/cors');
 var errors = require('./errors');
 var utils = require('./utils');
 
+var param = require('node-jquery-param');
+
 
 ///--- Globals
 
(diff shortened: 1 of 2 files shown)


micahr:
@11bucks Can you add a test or two that validates your change?

lib/router.js
```diff
@@ -182,8 +180,7 @@ Router.prototype.render = function render(routeName, params, query) {
 
     var _path = route.spec.path;
     var _url = _path.replace(/\/:([A-Za-z0-9_]+)(\([^\\]+?\))?/g, pathItem);
-    var items = Object.keys(query || {}).map(queryItem);
-    var queryString = items.length > 0 ? ('?' + items.join('&')) : '';
+    var queryString = Object.keys(query || {}).length ? ('?' + qs.stringify(query)) : '';
```
micahr: This line is too long, and is failing the tests currently. Please reduce it to less than 80 characters.

It might be helpful to move `Object.keys.length` call to a separate line. Similar to how it was before, with items.

yunong:
@11bucks For this one off case that doesn't adhere to the query RFC https://tools.ietf.org/html/rfc3986#page-23 I am not inclined to merge this into restify. You are however, very welcome to write a restify plugin for this. I can then add your plugin to a list of restify plugins.

PR #803, opened 2015-05-20
add redirect support to res

DonutEspresso:
Added a res.redirect method, which is primarily a sugar method, but makes redirects a lot easier to manage within Restify. I used the node core url module to avoid having to reinvent the wheel for URL parsing, but am open to suggestions.

Edit: Another nice thing about having redirect is to avoid the need for users to call `next(false)` after doing a manual redirect via `res.send(302)`. Calling next with false is required to stop the handler execution chain today. Given that, we will throw if redirect is invoked without next.

```js
// redirect to another domain 
res.redirect('http://www.foo.com', next);
res.redirect(301, 'http://www.foo.com', next);
// redirect to another url on same domain
res.redirect('/foo', next);
// reload the page on https using the options object.
// reload is implied due to missing pathname and hostname.
res.redirect({
    secure: true
}, next);
// redirect while appending to existing query string params
res.redirect({
    pathname: '/foo',
    query: {
        foo: 'bar'
    }
}, next); 
// redirect while stomping on existing query string params
res.redirect({
    pathname: '/foo',
    overrideQuery: true,
    query: {
        foo: 'bar'
    }
}, next); 
// redirect to another host/path using options object
// => https://www.foo.com/bar?a=1
res.redirect({
    hostname: 'www.foo.com',
    pathname: '/bar',  
    secure: true,
    query: {
        a: 1
    }
}, next);
```

Don't feel particularly strong about all the extra sugar properties - but I do think reload to https is an especially common use case.


lib/response.js
```diff
@@ -297,3 +300,111 @@ Response.prototype.writeHead = function restifyWriteHead() {
...
+    } else {
+        // otherwise if we have a next, but wer are missing a finalized uri
+        // by this point, pass an error to next.
+        if (!finalUri) {
+            return (next(new InternalServerError('could not construct url')));
+        }
+    }
+
+    // now we're done constructing url, send the res
+    this.send(statusCode, null, {
```
yunong: don't use this. declare var self at the top of the function and use self.send()

lib/response.js
```diff
@@ -297,3 +300,111 @@ Response.prototype.writeHead = function restifyWriteHead() {
...
+        }
+    }
+
+    // if we're missing a next we should probably throw. if user wanted
+    // to redirect but we were unable to do so, we should not continue
+    // down the handler stack.
+    if (typeof (next) !== 'function') {
+        throw new InternalServerError('res.redirect() requires a next param');
+    } else {
+        // otherwise if we have a next, but wer are missing a finalized uri
```
yunong: spelling s/wer/we

BreaGaudioso:
@DonutEspresso Is the body of the response on this supposed to return null?

DonutEspresso:
There's currently no facility to deliver a payload using this redirect API. Feel free to file an issue if this is something you'd like to see! In the meantime, I'd recommend doing a plain old `res.send` with the appropriate header and status codes.

BreaGaudioso:
Thanks for the quick response!

PR #804, opened 2015-05-21
migrate to eslint

DonutEspresso:
Moving to eslint gives us a lot more flexibility for custom linting rules. Also, hooray for AST, it trips up less on silly things. Tried to keep most of the rules 1:1, very little clean up was needed to move to strict. However, there is one file which I'll need help with before this can be merged.

Most diffs are silly:
- {} required for if statements everywhere, even one liners
- 'use strict' everywhere

More interesting changes were:
- string_client.js:145 have to futz with hoisting vars and functions out, in strict mode all functions must be hoisted at top level (no nesting within if blocks)
- http_error.js and rest_error.js now have a named constructor called 'BaseError'. arguments.callee is disallowed in strict mode, and as far as I can tell there's no reason not to give the function a name.

To fix:
- lib/response.js:214 there's an err from the formatter that's not being handled, but as far as I can tell none of the formatters take callbacks? 
- Do you want to make TODO/FIXME comments show up as warnings? Would be nice to know about them, but not fail the build on it. It's currently set to do that. There's only one TODO in the codebase now, which is always a good sign.

All unit tests pass, but CI won't pass until we fix lib/response.js.


lib/router.js
```diff
@@ -482,10 +505,11 @@ Router.prototype.find = function find(req, res, callback) {
     }
 
     // Check for 405 instead of 404
+    var j;
     var urls = Object.keys(this.reverse);
-    for (i = 0; i < urls.length; i++) {
```
DonutEspresso: This was previously undefined AFAICT. Not sure how JSL missed it.

tools/mk/Makefile.targ
```diff
@@ -157,27 +127,18 @@ check-bash: $(BASH_FILES:%=%.bashchk) $(BASH_FILES:%=%.bashstyle)
...
-
-%.jslwebchk: % $(JSL_EXEC)
-	$(JSL) $(JSL_FLAGS) $(JSL_FLAGS_WEB) $<
-
 .PHONY: check-jsstyle
 check-jsstyle: $(JSSTYLE_FILES:%=%.jsstylechk)
 
+.PHONY: check-eslint
+check-eslint::
+	$(ESLINT) .
```
DonutEspresso: I opted to use .eslintrc and .eslintignore here instead of redefining another set of files to feed in. The dotfiles are a well known paradigm that other tools will respect (IDEs included).

tools/mk/Makefile.targ
```diff
@@ -198,7 +159,7 @@ xref: cscope.files
 .PHONY: cscope.files
 cscope.files:
 	find $(CSCOPE_DIRS) -name '*.c' -o -name '*.h' -o -name '*.cc' \
-	    -o -name '*.js' -o -name '*.s' -o -name '*.cpp' > $@
+		-o -name '*.js' -o -name '*.s' -o -name '*.cpp' > $@
```
DonutEspresso: Not sure what happened here, but normalized all the tabs in the files and these diffs showed up. Might have been smart tabs...

lib/clients/http_client.js
```diff
@@ -9,14 +10,13 @@ var os = require('os');
 var querystring = require('querystring');
 var url = require('url');
 var util = require('util');
-var zlib = require('zlib');
+var path = require('path');
```
yunong: Imports should be alphabetically sorted. So path needs to go before querystring

tools/mk/Makefile.targ
yunong: Are these \t or 8*\w? make syntax requires the use of tabs...

.eslintrc
```diff
@@ -0,0 +1,125 @@
...
+        "no-use-before-define": [ 2, "nofunc" ],
+
+        // node.js
+        "handle-callback-err": [ 2, "^.*(e|E)rr" ],
+        "no-mixed-requires": [ 2 ],
+        "no-new-require": [ 2 ],
+        "no-path-concat": [ 0 ],
+        "no-process-exit": [ 0 ],
+
+        // stylistic
```
yunong: Thought we decided to drop the style checker since we already use jsstyle

Makefile
```diff
@@ -27,8 +28,6 @@ NPM		:= npm
 #
 DOC_FILES	 = index.restdown
 JS_FILES	:= $(shell find lib test bin -name '*.js')
-JSL_CONF_NODE	 = tools/jsl.node.conf
```
yunong: We should maintain the same convention and explicitly define the eslint configs here.

yunong:
Nice job, but looks like eslint is failing on CI
```
lib/plugins/conditional_request.js
  116:8  warning  Unexpected todo comment  no-warning-comments
lib/response.js
  214:4  error  Expected error to be handled  handle-callback-err
✖ 2 problems (1 error, 1 warning)
```

.eslintrc
DonutEspresso: Ah, yeah, good catch.

tools/mk/Makefile.targ
DonutEspresso: Hm, good point, will revert.

.eslintrc
DonutEspresso: Turns out some of the stylistic rules are on by default. I have to keep the stylistic rules in just to turn them off.

DonutEspresso:
Yar, I mentioned it in the first post, there's an unhandled err in lib/response.js and I'm not sure what the correct fix is. Would appreciate your input.

yunong:
@DonutEspresso Feel free to remove that err and change the two invocations from _cb(null, ret) to _cb(ret);

PR #809, opened 2015-05-24
Removed regexp for socket.io check

nkt:
Actually this check is still not fully cover `socket.io` usage. I mean `/socket.io` prefix could be redefined.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -258,7 +258,7 @@ function Server(options) {
     this.server.on('request', function onRequest(req, res) {
         self.emit('request', req, res);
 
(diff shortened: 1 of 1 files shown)


nkt:
Actually this check is still not fully cover `socket.io` usage.

yunong:
@nkt Not quite sure what you're trying to do here -- mind elaborating a bit?

nkt:
@yunong I've replaced regexp-driven check to `indexOf` check, which is faster.

yunong:
The perf benefits are negligible, and using the regex is a bit clearer in terms of documentation.

PR #819, opened 2015-06-02
The request expiry work with unit tests.

michaelbpaulson:


michaelbpaulson:
@yunong this should do it

michaelbpaulson:
Important
----------------

travis-ci is not claiming bad builds when unit tests fail.  This is because make is not catching the exit 1 condition from node unit on a test case failure.

docs/index.restdown
```diff
@@ -1081,6 +1082,21 @@ Options:
 
 Note that `ip`, `xff` and `username` are XOR'd.
 
+### Request Expiry
+
+Request expiry is used to kick out old requests that have been timed out somewhere upstream.
```
yunong: I might change the prose here to something a bit more succinct, like.

"Request Expiry can be used to throttle requests that have already exceeded their timeouts. Requests can be sent with a configured header, e.g. 'x-request-expiry-time', which gives in absolute wall time, when this request will be timed out by the client. 

This plugin will throttle all incoming requests via a 504 where 'x-request-expiry-time'  < Date.now() -- since these incoming requests will no longer be processed by the client."

docs/index.restdown
```diff
@@ -1081,6 +1082,21 @@ Options:
...
+expired time should be absolute time.  If the expire time is in the future (greater than
+`Date.now()` or does not exist the request will continue to execute the chain.  If the
+provided expiry time is less than `Date.now()` next will be provided a 504, Gateway
+Timeout Error.
+
+    server.use(restify.requestExpiry({
+        header: 'x-request-expiry-time'
+    });
+
+The only options provided is `header` which is the key used to look in each request header.
```
yunong: The only options provided is `header` which is the request header used to look specify the client timeout.

lib/plugins/index.js
```diff
@@ -23,6 +23,7 @@ module.exports = {
     sanitizePath: require('./pre/pre_path'),
     serveStatic: require('./static'),
     throttle: require('./throttle'),
+    requestExpiry: require('./request_expiry'),
```
yunong: please sort this key alphabetically.

lib/plugins/request_expiry.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+var assert = require('assert-plus');
+var GatewayTimeoutError = require('./../errors').GatewayTimeoutError;
+
+/**
```
yunong: Add @param jsdoc.

lib/plugins/request_expiry.js
```diff
@@ -0,0 +1,29 @@
...
+function requestExpiry(options) {
+    assert.object(options, 'options');
+    assert.string(options.header, 'options.header');
+    var headerKey = options.header;
+
+    return function (req, res, next) {
+        var expiryTime = Number(req.headers[headerKey]);
+
+        // The request has expired
+        if (expiryTime !== undefined && Date.now() > expiryTime) {
```
yunong: expiryTime will either be 0 if the header is null, or NaN if the header is a string, or a number. I'd check whether expiryTime is NaN here instead of undefined. Since Number() will never return undefined.

lib/plugins/request_expiry.js
```diff
@@ -0,0 +1,29 @@
...
+
+var assert = require('assert-plus');
+var GatewayTimeoutError = require('./../errors').GatewayTimeoutError;
+
+/**
+ * A request expiry will use the headers to tell if the
+ * incoming request has expired or not.  The header is
+ * expected to be in absolute time.
+ */
+function requestExpiry(options) {
```
yunong: Also explicitly specify that we take in ms since epoch

lib/plugins/request_expiry.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+var assert = require('assert-plus');
+var GatewayTimeoutError = require('./../errors').GatewayTimeoutError;
+
+/**
+ * A request expiry will use the headers to tell if the
+ * incoming request has expired or not.  The header is
+ * expected to be in absolute time since the epoch.
+ * @param {Object} options
```
yunong: Can you add a description of what options is and also add @param {string} options.header with a description?

michaelbpaulson:
ok, how about now @yunong ?

PR #822, opened 2015-06-04
Request

michaelbpaulson:
The purpose of this PR is to be able to specify a list of header keys to transfer from the request header to the props object of the request logger.  This will make the headers top level vars on the bunyan log.


michaelbpaulson:
Good luck @yunong

micahr:
Do you mind providing a description @michaelbpaulson ?

PR #825, opened 2015-06-09
Testing for issue #686 (Allow 0, false, and null as json body) #800

AlexDobeck:
Late testing for issue #686 requested in pull request #800


AlexDobeck:
Each test fails pre- #686 with a 'AssertionError' 0,false,null does not equal ''.

PR #826, opened 2015-06-13
req.getQuery() always returns a string

### LINKED ISSUE #374: req.query() wierdness
Currently if there is a query string in the url `req.query()` returns a string however, if there is no query string it returns an empty object.  This is really strange and almost impossible to deal with as you get two results and neither is easy to detect.

You may want to tell the url parser to parse the query string, or make the default an empty string.

DonutEspresso:
Fixes #374, but also changes the default behavior of req.getQuery(). There was some blurring of behavior between the default behavior, and the behavior expected from using the queryParser plugin.

Previously, if you were not using the queryParser plugin, you'd get an empty object, which is consistent with behavior from the queryParser plugin, but __inconsistent__ with the behavior of req.getQuery().

```
/foo 
req.getQuery(); // => {}
req.query; // => {}

/foo?a=1
req.getQuery(); // => 'a=1'
req.query; // => {a: 1}
```

With this pull request, the behavior is now:

```
/foo 
req.getQuery(); // => ''
req.query; // => {}
```

If you're using the plugin, you'll always get objects returned using `req.query`, but now `req.getQuery()` will consistently return a string even if you don't use the queryParser plugin.

### DIFF
--- a/docs/index.restdown
+++ b/docs/index.restdown
@@ -1339,6 +1339,12 @@ instance that is a child component of the one restify has:
 
     var log = req.getLogger('MyFoo');
 
+### getQuery()
+
+Returns the raw query string. Returns empty string if no query string is found.
+To get the parsed query object, use the `queryParser` plugin. More info can
+be found about the plugin in the [bundled plugins](#bundled-plugins) section.
+
 ### time()
 
 The time when this request arrived (ms since epoch)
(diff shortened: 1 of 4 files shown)


DonutEspresso:
Hmm, not sure why CI failed only for 0.11...?

PR #694, opened 2014-10-13
multipart form POST, formidable hash option

ManRueda:
Active formidable option to create a hash of uploaded files.

PR #831, opened 2015-06-15
Copied fix from #694 and added test to verify

blakevanlan:
Took up where PR #694 left off to add support passing the hash option to Formidable for multipart file uploads.


micahr:
Looks good to me. Thanks for adding this @blakevanlan.

Issue #830, opened 2015-06-15
Installation of dtrace-provider fails using iojs 2.x


yads:
There is an issue with dtrace-provider and nan. Looks like it was [resolved in version 0.5.0](https://github.com/chrisa/node-dtrace-provider/pull/61). Can we get a version bump? Here's some output:

```
  c++ '-DNODE_GYP_MODULE_NAME=DTraceProviderBindings' '-D_DARWIN_USE_64_BIT_INODE=1' '-D_LARGEFILE_SOURCE' '-D_FILE_OFFSET_BITS=64' '-DBUILDING_NODE_EXTENSION' -I/Users/vadimkazakov/.node-gyp/2.3.0/src -I/Users/vadimkazakov/.node-gyp/2.3.0/deps/uv/include -I/Users/vadimkazakov/.node-gyp/2.3.0/deps/v8/include -I../libusdt -I../node_modules/nan  -Os -gdwarf-2 -mmacosx-version-min=10.5 -arch x86_64 -Wall -Wendif-labels -W -Wno-unused-parameter -std=gnu++0x -fno-rtti -fno-exceptions -fno-threadsafe-statics -fno-strict-aliasing -MMD -MF ./Release/.deps/Release/obj.target/DTraceProviderBindings/dtrace_provider.o.d.raw  -c -o Release/obj.target/DTraceProviderBindings/dtrace_provider.o ../dtrace_provider.cc
In file included from ../dtrace_provider.cc:1:
In file included from ../dtrace_provider.h:1:
In file included from ../node_modules/nan/nan.h:74:
In file included from ../node_modules/nan/nan_new.h:181:
../node_modules/nan/nan_implementation_12_inl.h:172:66: error: too many
      arguments to function call, expected at most 2, have 4
  ...v8::Signature::New(v8::Isolate::GetCurrent(), receiver, argc, argv);
     ~~~~~~~~~~~~~~~~~~                                      ^~~~~~~~~~
/Users/vadimkazakov/.node-gyp/2.3.0/deps/v8/include/v8.h:4188:3: note: 
      'New' declared here
  static Local<Signature> New(
  ^
1 error generated.
```

PR #832, opened 2015-06-16
dtrace-provider version bump

yads:
Fix for #830

Also see https://github.com/chrisa/node-dtrace-provider/pull/61

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "verror": "^1.4.0"
   },
   "optionalDependencies": {
-    "dtrace-provider": "^0.4.0"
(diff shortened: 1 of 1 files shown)


yads:
Thanks @yunong. When do you think we'll be able to see this in a release?

corbinu:
Any chance this will be released before 3.1.x?

yunong:
We're in the middle of preparing for the 4.x release. In the meantime, you
can refer to the fix in your pacakge.json by using git urls.
https://docs.npmjs.com/files/package.json#git-urls-as-dependencies

PR #835, opened 2015-06-16
Wavewash patch 1

micahr:
Adds a test for @Wavewash's change in #719

Also fix up the JSCS errors.

Closes #719


Wavewash:
Thanks for doing this! I've been swamped.

PR #836, opened 2015-06-17
jsdoc comments

DonutEspresso:


micahr:
Such docs. Very commented. Well done.

:+1:

lib/bunyan_helper.js
```diff
@@ -43,8 +51,9 @@ function appendStream(streams, s) {
  * A Bunyan stream to capture records in a ring buffer and only pass through
  * on a higher-level record. E.g. buffer up all records but only dump when
  * getting a WARN or above.
- *
- * @param {Object} options contains the parameters:
+ * @public
+ * @class
+ * @param {Object} opts contains the parameters:
  *      - {Object} stream The stream to which to write when dumping captured
```
yunong: Is this the blessed way to doc a complex object? I had thought it was something like
```js
/**
 * @param {Object} opts
 * @param {Object} opts.someotherobject
 * @param {String} opts.someString
 */
```

micahr: Per the [JSDoc docs](http://usejsdoc.org/tags-param.html#parameters-with-properties): 
> If a parameter is expected to have a specific property, you can document that property by providing an additional @param tag.

So it should probably be documented as @yunong proposed.

lib/index.js
```diff
@@ -12,6 +12,13 @@
 var shallowCopy = require('./utils').shallowCopy;
 
 
+/**
+ * creates a HttpClient or a flavor of it.
```
yunong: A flavor? What kind of flavour? Watermelon? Mint? All kidding aside, might be worth elaborating on what that is.

lib/index.js
```diff
@@ -90,6 +118,13 @@ function createHttpClient(options) {
 }
 
 
+/**
+ * creates a server.
+ * @public
+ * @function createServer
+ * @param    {Object} options an options object
```
yunong: as a TODO, once the new docs are in place, we should link to them from within the jsdocs.

lib/plugins/cors.js
```diff
@@ -53,25 +53,29 @@ function matchOrigin(req, origins) {
...
+ * 2. If the resource supports credentials add a single
+ *    'Access-Control-Allow-Credentials' header with the value as "true", and
+ *    ensure 'AC-Allow-Origin' is not '*', but is the request header value,
+ *    otherwise add a single Access-Control-Allow-Origin header, with either the
+ *    value of the Origin header or the string "*" as value
+ *
+ * 3. Add Access-Control-Expose-Headers as appropriate
+ *
+ * Pre-flight requests are handled by the router internally
+ * @public
```
yunong: Nit, new line before the @public

lib/bunyan_helper.js
DonutEspresso: Good catch, that was an existing comment block, so I didn't look too closely. I'll fix it up as proposed.

lib/index.js
DonutEspresso: Yeah, I experimented with generating the JSDocs, but they're super wonky, and there are lots of corner cases which are less than ideal. My instinct is to keep the JSDoc comments in place purely for developer use (which is always good), and continue to shore up user docs as a separate thing.

PR #839, opened 2015-06-19
Send 500 status if async formatter sends error

jameswomack:
# Async Formatter Handling Errors

**lib/response.js** contained a TODO and ESLint override related to the async formatter handler not dealing with errors being passed back. This TODO caused a warning in the lint task which led to me trying to tackle this in the lightest manner possible.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -302,9 +302,12 @@ Response.prototype.send = function send(code, body, headers) {
 
     this._body = body;
 
-    // TODO: restify doesn't ship with any async formatters, but
-    // presumably if an async formatter blows up, we should return a 500.
-    function _cb(err, _body) { // eslint-disable-line handle-callback-err
+    function _cb(err, _body) {
+        // If an async formatter blows up, we should return a 500.
+        if (Error.prototype.isPrototypeOf(err)) {
+            self.statusCode = 500;
+        }
+
         self._data = _body;
         Object.keys(headers).forEach(function (k) {
             self.setHeader(k, headers[k]);
--- a/test/formatter.test.js
+++ b/test/formatter.test.js
@@ -21,6 +21,7 @@ var CLIENT;
 var SERVER;
 
 var isFirstRequest = true;
+var isSecondRequest = true;
 
 
 ///--- Tests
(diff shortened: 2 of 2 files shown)


micahr:
Nice. Looks good to me.

yunong:
Not sure if we should merge this yet -- we should figure out what the implications of async formatters are first. Gonna revert for now.

micahr:
Sure. This was to just handle errors from them, if one was thrown in a CB.

DonutEspresso:
The error should probably be propagated back to the formatter via the callback, I imagine the formatter should have a chance to handle the error before sending a 500.

jameswomack:
My understanding was that I'm testing for when the async formatter itself passes an error. No?

DonutEspresso:
Ah, sorry, you are right, that's what I get for reading pull requests on a phone. I think the missing piece is that errors from async formatters aren't actually getting propagated to `_cb`: 

https://github.com/restify/node-restify/blob/master/lib/response.js#L330
https://github.com/restify/node-restify/blob/master/lib/response.js#L333

Will defer to @yunong re: implications.

micahr:
Errors from response formatters would be propagated in if the user called the last argument that was passed to them with an error, but nothing would handle it:

https://github.com/restify/node-restify/blob/master/lib/response.js#L327 ->
https://github.com/restify/node-restify/blob/master/lib/response.js#L150

Anyone could write an async response formatter, and pass an error to the `_cb`, this is just handling it if someone does. Throwing a 500 seems like the expected behavior, especially if users are writing their own response formatters.

DonutEspresso:
Oh yikes, I've been reading this all wrong, you two are on the mark. 

I do think it would better to send out whatever error the formatter returned, though. This would allow users to craft their own custom errors to give more context, vs a generic InternalServerError.

micahr:
Perhaps a check for the Error object being one from restify-errors, and if it isn't, send a generic 500, otherwise, send the error the formatter returned?

DonutEspresso:
Yeah, I'd vote for always sending the error if it's an instane of HttpError or RestError (even custom errors will inherit from RestError using restify-errors). Only override with InternalServerError when err is not actually an Error object.

I think the other implication here is that `res.send()` wraps the formatting, so not even sure how async formatter behavior would work as expected. As is, you could end up calling next() before the formatter has even returned.

yunong:
Yes, let's: 
1) Return the formatter error if derives from HttpError.
2) Return InternalError if it's not an instance of Http or Rest error.

yunong:
Filed #845.

PR #842, opened 2015-06-20
Adding req.matchedVersion()

micahr:
Fix up JSCS errors from #635 by @nathanpeck

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -302,6 +302,13 @@ Request.prototype.getVersion = function getVersion() {
 };
 Request.prototype.version = Request.prototype.getVersion;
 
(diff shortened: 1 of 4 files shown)


lib/request.js
```diff
@@ -302,6 +302,13 @@ Request.prototype.getVersion = function getVersion() {
 };
 Request.prototype.version = Request.prototype.getVersion;
 
+Request.prototype.matchedVersion = function matchedVersion() {
+    if (this._matchedVersion !== undefined) {
```
DonutEspresso: Code this just be return `(this.matchedVersion || this.version())`?

micahr: Sure, I didn't change any of the code that came from @nathanpeck, just added tests.

Issue #692, opened 2014-10-07
Support strict routing


Dante-Edge:
Like the strict options in express. "/foo" and "/foo/" can be treated different by route.

DonutEspresso:
Fixed, will be released with 5.x before end of the month.

maxsummers:
End of what month? )

DonutEspresso:
Hello - we've been testing the 5.x beta on the beta tag over last couple of months (`npm i restify@beta`). We intend to formalize the release soon. Feel free to check out the beta, the API is pretty much finalized.

PR #843, opened 2015-06-20
Support URLs that have a path component

micahr:
Combine both PRs #633 and #628 

Pull the logic from #633 and clean it up a bit
Pull the tests from #628

Thanks to both @nakosung and @calebbrown for their contributions.

The goal of this PR is to use the path from both the Client options and the requested url.

For example, when a user creates a client for a non-root url, any subsequent requests should use that as part of the request path.

```javascript
var client = restify.createClient({
    url: 'http://127.0.0.1/str'
});

client.get('/foo', function (err, req, res, data){});
```

The resulting request should go to `http://127.0.0.1/str/foo`.

### DIFF
--- a/lib/clients/http_client.js
+++ b/lib/clients/http_client.js
@@ -630,14 +630,15 @@ HttpClient.prototype._options = function (method, options) {
 
     // url[http://localhost:3000/something] + get[/foo] -> http://localhost:3000/something/foo
     if (opts.path) {
-        if (this.url.path[this.url.path.length-1] == '/' && opts.path[0] == '/') {
+        if ( this.url.path[this.url.path.length - 1] === '/' &&
+            opts.path[0] === '/' ) {
+
             opts.path = this.url.path + opts.path.substr(1);
-        }
-        else {
+        } else {
             opts.path = this.url.path + opts.path;
         }
     }
-    
+
     Object.keys(this.url).forEach(function (k) {
         if (!opts[k]) {
             opts[k] = self.url[k];
--- a/lib/clients/http_client.js
+++ b/lib/clients/http_client.js
@@ -628,6 +628,11 @@ HttpClient.prototype._options = function (method, options) {
         opts.socketPath = this.socketPath;
     }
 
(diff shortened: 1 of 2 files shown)


lib/clients/http_client.js
```diff
@@ -628,6 +629,15 @@ HttpClient.prototype._options = function (method, options) {
         opts.socketPath = this.socketPath;
     }
 
+    // Add the request path to the base path of the client
+    // Don't combine the request url and client path if they both
+    // have a slash
+    if (opts.path && this.url.path &&
+        opts.path !== '/' && this.url.path !== '/') {
+        // clean up the url path, stripping the last /
+        opts.path = utils.sanitizePath([this.url.path, opts.path].join('/'));
```
DonutEspresso: Based on #692, I think it's safe to say that we probably don't want to strip trailing slashes. Make the users be explicit?

DonutEspresso: Also, could probably use `url.resolve()` in favor of `[].join('/')`;

micahr: This wouldn't affect #692 as this is a http client making requests to any other http server / endpoint, not the server managing routes.

`url.resolve()` won't work if the request url has a / at the beginning, it will replace whatever path is already at the beginning of the path. 

```
url.resolve('/foo/bar', '/baz') === '/baz';
```

DonutEspresso: That's a bummer, maybe we should look into something like URIjs, there's a lot of URL mugging that should be centralized (plugins too).

Re: above sorry, I should have been clearer, I just think that we should be consistent, presumably if users need strict URLs on the server, them they need a client that can hit those strict URLs. If we strip trailing slashes automatically them it's impossible to do so. Maybe we can do a strict:true option?

micahr: I agree, we should definitely standardize the management of urls via the server and client libraries.

yunong:
This is definitely a feature we want. But I'm going to lean on not merging this specific PR. We should figure out the URL interface and standardize on that per what @micahr suggested. Will close and file #846 to that effect.

PR #849, opened 2015-06-25
clarify that certain req properties are functions.

### LINKED ISSUE #638: id is a method not property
req.id() returns id of request.
In documentation details that this is a property:

id	String	A unique request id (x-request-id)

Please update documentation, its really deprecated and we cant learn to use this library correctly.

DonutEspresso:
Fixes #638.


yunong:
Looks good. Can you update the gh-pages branch with your doc fixes?

PR #851, opened 2015-06-27
Formatter functions now must always return a cb of the type f(err, data)

yunong:
In the past, formatters could either directly return the body if synchronous, or return a cb(err, data) if it was async.

Standardizing this interface so that all formatters must *always* return a callback instead. We also now handle the error that comes back from the formatter.

### DIFF
--- a/docs/index.restdown
+++ b/docs/index.restdown
@@ -412,14 +412,14 @@ in a hash of content-type -> parser at server creation time:
 
     var server = restify.createServer({
       formatters: {
-        'application/foo': function formatFoo(req, res, body) {
+        'application/foo': function formatFoo(req, res, body, cb) {
           if (body instanceof Error)
             return body.stack;
 
           if (Buffer.isBuffer(body))
-            return body.toString('base64');
+            return cb(null, body.toString('base64'));
 
-          return util.inspect(body);
+          return cb(null, util.inspect(body));
         }
       }
     });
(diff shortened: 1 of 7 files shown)


magwo:
This is a breaking change that we encountered while upgrading (from 2.6ish but will happen for 3.0.3 -> 4.0.0 also). Maybe it should be marked as breaking in the docs?

micahr:
Thanks for the feedback @magwo, we had that in the release notes for 4, but didn't call it out explicitly. We are working to improve our release process and appreciate the feedback.

trentm:
+1 on calling that out in the relnotes, I hit this too.

trentm:
Ah, I see that it was marked in the '5.x' branch. I'm looking forward to that branch content eventually being on master where it is easier to see. For bonus points, getting the "BREAKING" marker on the 4.x branch would be helpful.

A note for others coming here for details on the API changes necessary. Where you previously might have had something like this:

```
function formatJSON(req, res, body) {
    // ...
    return (data);
}

var server = restify.createServer({
    formatters: {
        'application/json': formatJSON
    },
    // ...
});
```

You need to change that to always use a given callback:

```
function formatJSON(req, res, body, cb) {
    // ...
    cb(null, data);
}

var server = restify.createServer({
    formatters: {
        'application/json': formatJSON
    },
    // ...
});
```

PR #862, opened 2015-07-07
restify version bump so that a tag can be cut.

michaelbpaulson:
I need access to the `requestExpiry` plugin.  So i need to get a new tag published.  @yunong could you cut a tag / publish npm when you can?

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
   "name": "restify",
   "homepage": "http://restifyjs.com",
   "description": "REST framework",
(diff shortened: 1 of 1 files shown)


micahr:
I'm not sure why a revision change would be necessary. I don't think we are ready to do a release at this point. You can refer to a specific git commit directly through NPM, which would be a lot safer than us cutting a release.

michaelbpaulson:
Sure, i can kill this PR.

PR #866, opened 2015-07-14
Improve request handler name handling

LeMisterV:
This change allow request handler dev to force a name for a request handler and see this name in bunyan logs instead of a meaningless "?"

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -891,13 +891,15 @@ Server.prototype._run = function _run(req, res, route, chain, cb) {
                 return (next());
             }
 
+            handlerName = (chain[i].handlerName ||
+                           chain[i].name ||
+                           ('handler-' + req._anonFuncCount++));
+
             if (log.trace()) {
-                log.trace('running %s', chain[i].name || '?');
+                log.trace('running %s', handlerName);
             }
 
             req._currentRoute = (route !== null ? route.name : 'pre');
-            handlerName = (chain[i].name ||
-                           ('handler-' + req._anonFuncCount++));
             req._currentHandler = handlerName;
             req.startHandlerTimer(handlerName);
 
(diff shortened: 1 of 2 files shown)


LeMisterV:
Tests seems unstable. Restarting on my fork worked just well.... subject for a future pull request maybe ;)

micahr:
I restarted the build, there are a couple timing based tests, so that can happen.

micahr:
@yunong The PR looks good, just moved the handlerName assignment up so the trace log can make use of it too.

DonutEspresso:
Hmm, this will affect the way conductor handles nested handler timing. Would have to see what the impact is on my branch.

lib/server.js
```diff
@@ -891,13 +891,15 @@ Server.prototype._run = function _run(req, res, route, chain, cb) {
                 return (next());
             }
 
+            handlerName = (chain[i].handlerName ||
```
yunong: @LeMisterV Presumably ```handlerName``` is a property on the function that you'd set when you initialize the route? If so we'll need to update docs to reflect this before we merge.

LeMisterV: That's right, this handlerName property should be mentioned in the docs.
I'm not really sure where and how to add this in the doc, and as I'm not native english speaker I'm not sure I'll do it right.... but I'll try something ;)

LeMisterV: I just pushed some docs changes to add handler naming possibility, tell me if this is not the way you would see it.

LeMisterV: By the way, CI tests failed again.... without any code change I think it's again a timing problem.... don't know how to get it passed

DonutEspresso:
I think this looks good, all. Any reason not to get this merged?

PR #867, opened 2015-07-17
Allow objects for errors when calling next()

andyburke:
I was surprised when I tried the following code and it allowed the handler chain to proceed:

```javascript
function someMiddleware( request, response, next ) {
    // ...
    next( {
        error: 'permission denied',
        message: 'You do not have permission.',
        statusCode: 403
    } );
}
```

This change allows passing an object to next() that will be sent as an error.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -842,7 +842,7 @@ Server.prototype._run = function _run(req, res, route, chain, cb) {
                 done = true;
             } else if ( typeof arg === 'object' ) {
                 log.trace({err: arg, errName: 'Unknown'}, 'next(err=Unknown)');
-                res.send(arg, arg.statusCode || 500 );
+                res.send(arg.statusCode || 500, arg);
                 done = true;
             } else if (typeof (arg) === 'string') { // GH-193, allow redirect
                 if (req._rstfy_chained_route) {
--- a/lib/server.js
+++ b/lib/server.js
@@ -840,6 +840,10 @@ Server.prototype._run = function _run(req, res, route, chain, cb) {
                     res.send(arg);
                 }
                 done = true;
+            } else if ( typeof arg === 'object' ) {
+                log.trace({err: arg, errName: 'Unknown'}, 'next(err=Unknown)');
+                res.send(arg);
+                done = true;
             } else if (typeof (arg) === 'string') { // GH-193, allow redirect
                 if (req._rstfy_chained_route) {
                     var _e = new errors.InternalError();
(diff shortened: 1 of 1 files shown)


micahr:
Thank you for the Pull Request, Andy. 

In order to properly throw an error in this case, you will also need to emit the error, so others can listen to the event. Check the if clause for an example of what needs to be emitted and how. Also, please add some tests, referencing the PR number (867).

DonutEspresso:
Hm, I'm not sure this is a desirable pattern, as a plain object doesn't have a stack trace, capital E Error objects are useful in that regard.

Can you share some more info about your use case? You do this, today:

```
var errors = require('restify').errors;
var myErr = errors.codeToHttpError(403, 'you do not have permission');
next(myErr);
```

andyburke:
@DonutEspresso I can understand the point about the stack trace. Would your example work with something like this?:

```javascript
var errors = require( 'restify' ).errors;
var myErr = errors.codeToHttpError( 403, {
    error: 'internal error code',
    message: 'This is the default error message in this case.'
} );
next( myErr );
```

Such that the HTTP response would be a JSON object that looked something like:

```JSON
```

andyburke:
Looks like perhaps I would have to do something like:

```javascript
var errors = require( 'restify' ).errors;
var myErr = errors.codeToHttpError( 403, 'This is the default error message.', {
    error: 'internal error code'
} );
next( myErr );
```

??

andyburke:
Actually, on further reflection, I still feel like this PR has value. In my instance, I am trying to write some middleware. That middleware talks to other pieces of code and gets callbacks which may have errors. I don't know what those errors are, but I want to be able to percolate them up to restify to send back an error response.

It seems like if I were to go this way, I'd need to do the checking in the middleware for what type of object I've gotten back and then handle it accordingly with restify.

It seems better to push that handling all the way to the 'top' in the restify error handling. I see two ways forward:

1) I can improve this PR to do things like add a stack trace, even when given a string or an object.
2) I could allow for installing a custom error handling mechanism in restify.

Would one of those be preferable?

yunong:
@andyburke The right thing to do here is to wrap the error from other pieces of code in a restify.HttpError or restify.RestError.

andyburke:
@yunong can you give me a quick example?

yunong:
@andyburke here's an example: https://github.com/restify/node-restify/blob/master/test/server.test.js#L1082-L1085

PR #869, opened 2015-07-20
Add `all` to server API

jameswomack:
Added a `server.all` method, which is primarily a sugar method, but makes migrating from Express easier and makes leveraging Express knowledge for Restify easier.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -44,6 +44,17 @@ var PROXY_EVENTS = [
 ];
 
 
+var VERBS = [
+    'del',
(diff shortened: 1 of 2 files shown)


yunong:
Thanks for the PR @jameswomack. However I do think this is a pretty drastic change to the API, in addition it's not entirely clear to me how this affects routes already defined. I would lean towards making this a third party module for now. At some point, we'll want to revisit the API design, which then will be a good place to talk about potentially merging this in.

PR #871, opened 2015-07-31
Make sure signRequest is called before send

heiparta:
If the socket connected very quickly (e.g. to localhost), the signRequest
callback was never called.


heiparta:
Just noticed that there is an ongoing action to move the clients into their own repository, so made a PR there as well: https://github.com/restify/clients/pull/6

micahr:
I'll close this issue, in favor of https://github.com/restify/clients/pull/6

PR #876, opened 2015-08-04
Correct examples/dtrace/demo to use cb formatter

jameswomack:
There were some inconsistencies in the DTrace demo.


micahr:
Thanks for fixing the demo.

PR #879, opened 2015-08-14
case-insensitive content-type check (fixes #877)

### LINKED ISSUE #877: body parser doesn't treat content types as case-insensitive
Example server:
```javascript
var server  = restify.createServer();

server.post('/', restify.jsonBodyParser(), function(req, res, next) {
  console.log('Body:', req.body);
  res.send(200);
  next();
});

server.listen(8080);
```

When called as follows, `req.body` unexpectedly is a `Buffer`:
```
$ curl -XPOST localhost:8080 -d '{"foo":"bar"}' -H'Content-Type:Application/Json'
```
(notice the content type case)

It works as expected with this:
```
$ curl -XPOST localhost:8080 -d '{"foo":"bar"}' -H'Content-Type:application/json'
```

According to [RFC2045](http://tools.ietf.org/html/rfc2045#section-5.1): _"Matching of media type and subtype is ALWAYS case-insensitive"_.

gcochard:
This should fix #877 without any regressions. Please let me know if I can improve it in any way.

### DIFF
--- a/lib/plugins/json_body_parser.js
+++ b/lib/plugins/json_body_parser.js
@@ -25,7 +25,7 @@ function jsonBodyParser(options) {
     var override = options.overrideParams;
 
     function parseJson(req, res, next) {
-        if (req.getContentType() !== 'application/json' || !req.body) {
+        if (!/^application\/json$/i.test(req.getContentType()) || !req.body) {
             next();
             return;
         }
--- a/test/json_body_parser.test.js
+++ b/test/json_body_parser.test.js
@@ -72,3 +72,30 @@ test('no reviver', function (t) {
     });
 });
 
+test('case-insensitive content-type', function (t) {
+    var jsonBodyParser = restify.plugins.jsonBodyParser;
+
+    var parser = jsonBodyParser({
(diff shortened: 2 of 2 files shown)


lib/plugins/json_body_parser.js
```diff
@@ -25,7 +25,7 @@ function jsonBodyParser(options) {
     var override = options.overrideParams;
 
     function parseJson(req, res, next) {
-        if (req.getContentType() !== 'application/json' || !req.body) {
```
yunong: Just change this to `req.getContentType().toLowerCase();

yunong: Actually, this needs to be fixed in request.js, such that it applies to all content-types. I'll do that. Thanks for the PR though.

gcochard: I considered doing that first, but I didn't want to modify the output in case there was reliance on it. On second thought, I can't imagine anyone depending on that output being case-sensitive.

yunong:
Closing in favour of fixes from 7a549c4

Issue #882, opened 2015-08-24
Route name is stripped of hyphens


sean3z:
I understand the route names are typically used internally solely for the `router.render()` method but, this doesn't seem like expected behavior nor does it allow for a given route name to be useful in any other cases.

Is there any way to get around this?

```js
app.get({
  name: 'my-awesome-route',
  path: '/my-awesome-route'
}, ...)
```

Will provide the following `req.route` object

```js
route: { 
  name: 'myawesomeroute',
  path: '/my-awesome-route',
  method: 'GET',
  versions: [] 
}
```

**edit**: It looks like this is happening for several other characters as well

yunong:
@sean3z this seems like a bug, do you have the cycles to submit a PR?

sean3z:
@yunong sure thing ^^ :)

sean3z:
Fixed in https://github.com/restify/node-restify/commit/30ff6cc9013d4d0168ade28082289ef8f49239e7

PR #883, opened 2015-08-25
Removing replace logic for route name

sean3z:
GH Issue #882

This PR ensures that a given route name is returned the same as specified by removing the (seemingly unnecessary) replace logic. As it stands, route names provided will be stored with any `/[^\w]/g` characters removed. This effectively makes them useless as a reference outside of the `router.render()` usage.

```js
app.get({
  name: 'my-awesome-route',
  path: '/my-awesome-route'
}, ...)
```

Currently provides the following `req.route` object

```js
route: { 
  name: 'myawesomeroute',
  path: '/my-awesome-route',
  method: 'GET'
}
```

The change here proposes that the `name` property should instead be, `my-awesome-route`

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -225,8 +225,7 @@ Router.prototype.render = function render(routeName, params, query) {
         return (encodeURIComponent(key) + '=' + encodeURIComponent(query[key]));
     }
 
-    var routeKey = routeName.replace(/\W/g, '').toLowerCase();
-    var route = this.mounts[routeKey];
+    var route = this.mounts[routeName];
 
     if (!route) {
         return (null);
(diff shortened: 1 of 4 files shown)


micahr:
Thanks for adding this @sean3z. Since this involves changing route names, and how they are matched, we should probably rev at least a minor version.

kodypeterson:
Nice work @sean3z always room to shave off some processing time :smile:

sean3z:
@micahr done :)

micahr:
@sean3z Sorry, I meant that for our future reference. We won't want to rev to 4.1.0 until we are ready to do another release.

sean3z:
ah fair enough, I've reverted the bump

micahr:
@yunong - This should be good to go, any objections?

PR #887, opened 2015-09-05
Bump dtrace provider for iojs 3.x and future node support

corbinu:
Dtrace 0.6 has been upgraded to Nan 2 for future node support.


micahr:
Looks good to me.

corbinu:
@michar This needs to happen on restify-clients also looks like

micahr:
I can make the same change there, thanks!

PR #891, opened 2015-09-08
stop processing requests on which the close event has been fired

### LINKED ISSUE #733: Allow option to short circuit requests that were closed early
When the underlying connection on a request is terminated, a 'close' event is fired on the request. It would be nice to have an option to be able to set a flag or call an exposed API somewhere that will tell Restify to stop processing that request. I can listen to the 'close' event, but am unable to do anything about it. 

In the case of web apps that may do lots of rendering, terminated requests continue to move through the handler chain. These requests should be dropped and the response ended to prevent it from taking up CPU time/resources.

DonutEspresso:
Fixes #733.

Not sure if this is too naive, as I'm using the same place we check for `next(false)` to determine if the request should be stopped. I believe from a resource management perspective we should be okay, but would be good to get second pair of eyes on it.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -824,7 +824,7 @@ Server.prototype._run = function _run(req, res, route, chain, cb) {
     // (or we lose the connection).
     function _requestClose() {
         reqClosed = true;
-        req.earlyClose = true;
+        req.clientClosed = true;
     }
     req.once('close', _requestClose);
 
--- a/lib/server.js
+++ b/lib/server.js
@@ -824,6 +824,7 @@ Server.prototype._run = function _run(req, res, route, chain, cb) {
     // (or we lose the connection).
     function _requestClose() {
         reqClosed = true;
+        req.earlyClose = true;
     }
     req.once('close', _requestClose);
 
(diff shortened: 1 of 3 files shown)


test/server.test.js
```diff
@@ -1929,3 +1937,38 @@ test('GH-877 content-type should be case insensitive', function (t) {
     });
     client.end();
 });
+
+test('GH-733 if request is closed early, stop processing', function (t) {
```
yunong: For this test, you'll need to verify that all of the post processing on the server has completed. e.g. audit logs, timers, etc.

DonutEspresso:
@yunong Missing audit log tests were added - also changed the logic a bit for determining whether or not to continue. Good to go from my end.

lib/server.js
```diff
@@ -813,11 +813,21 @@ Server.prototype._run = function _run(req, res, route, chain, cb) {
...
 
     if (cb) {
         cb = once(cb);
     }
 
+    // attach a listener for 'close' event, this will let us set a flag so that
+    // we can stop processing the request if the client closes the connection
+    // (or we lose the connection).
+    function _requestClose() {
+        reqClosed = true;
```
yunong: Now that I think about it, we might want to add an additional field to the audit log to indicate that this was a prematurely closed request. Perhaps a field named ```clientClosed```.

Sorry totally my bad for not asking about this earlier.

DonutEspresso: done

Issue #888, opened 2015-09-07
Error param when fn emitRouteError emits 'after' event


nicolaspeixoto:
I'm trying to log all my server requests just using 'after' event emitted from server. But when i send an error in next function the error isn't sent to after event.

```javascript
server.get('/hello/:name', function(req, res, next) {
  // some internal unrecoverable error
  var err = new restify.errors.InternalServerError('oh noes!');
  return next(err);
});
```
emitRouteError:
``` javascript
   req.log.trace({name: name, err: err}, 'entering emitRouteError');
    if (server.listeners(name).length > 0) {
        server.emit(name, req, res, err, once(function () {
            server.emit('after', req, res, null);
        }));
    } else {
        res.send(err);
        server.emit('after', req, res, null);
    }
```
How can i log the error if i don't want to use a listener for each type of error?

micahr:
You can catch all error events with `uncaughtException`: 

```javascript
server.on('uncaughtException', function(req, res, route, err) {
    // log and cleanup your error here
});
```

This is good practice to ensure you are logging all possible events. If, inside this function, you want to handle individual errors in different ways, you should be able to.

As for logging everything in the `after` event, I would recommend using the audit logger with Bunyan to ensure you get all the information about your request: http://restify.com/#audit-logging.

nicolaspeixoto:
I am already using uncaughtException, but in this case, i'm sending the error on next function, and i need to build my own log function.

for example: 
``` javascript
if(err){
    next(new InternalServerError(err, 'oh noes!'));
}

server.on('after', function(req, res, route, err){
    /* err is null! */
    if(err){
         req.log.error(err);
     }
}
```
Why emitRouteError() doesn't send the error param?

micahr:
`emitRouteError` is called when a route couldn't be found using `router.get` or `router.find`. In that case, the request is effectively over because a suitable routing chain couldn't be determined. Is is assumed that if you got the error from the router, then you weren't able to find a route.

[The code](https://github.com/restify/node-restify/blob/master/lib/server.js#L828-L842) for handling an error passed to `next` will only send the error to the callback, which will in turn [send](https://github.com/restify/node-restify/blob/master/lib/server.js#L713) it to `after`, if there is a listener for that error. The normal behavior for sending an error to `next`, is that you are wanting to send this to the client. If there is no listener for any given error, then it is assumed that the application developer won't be handling the error and it needs to be sent to the client.

[...]

PR #892, opened 2015-09-08
GH-667: support generic error listener

### LINKED ISSUE #667: Error Logging
At the moment I'm using
```
if (err)
    log.error(err);    
    return next(err);
```

In Restify 2.1 we can use `next.ifError(err);`

How can I create a general handler that will log all errors before they are sent in the response?, i.e. whenever it's handed to 'next()'? 

Will it show up in...

`server.on('uncaughtException', function (req, res, route, err)` ?

In which case can I then just ...

```
log.error(err);
res.send(err);
```
Or do I need to write my own handler?

DonutEspresso:
Fixes #667 and #888. Emit to a generic error event when typed error event listeners are not registered.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -14,6 +14,7 @@ var mime = require('mime');
 var once = require('once');
 var spdy = require('spdy');
 var uuid = require('node-uuid');
+var vasync = require('vasync');
 
 var dtrace = require('./dtrace');
 var formatters = require('./formatters');
@@ -843,24 +844,29 @@ Server.prototype._run = function _run(req, res, route, chain, cb) {
                 // possible. fall back on generic 'error' listener if we can't
                 // find one for the error we got.
                 var hasErrListeners = false;
-                var errEvtName;
+                var errEvtNames = [];
 
                 // if we have listeners for the specific error
                 if (self.listeners(errName).length > 0) {
                     hasErrListeners = true;
-                    errEvtName = errName;
+                    errEvtNames.push(errName);
                 }
                 // or if we have a generic error listener
-                else if (self.listeners('error').length > 0) {
+                if (self.listeners('restifyError').length > 0) {
                     hasErrListeners = true;
-                    errEvtName = 'error';
+                    errEvtNames.push('restifyError');
                 }
 
                 if (hasErrListeners) {
-                    self.emit(errEvtName, req, res, arg, once(function () {
-                        res.send(arg);
-                        return (cb ? cb(arg) : true);
-                    }));
+                    vasync.forEachPipeline({
+                        func: function emitError(errEvtName, vasyncCb) {
+                            self.emit(errEvtName, req, res, arg, vasyncCb);
+                        },
+                        inputs: errEvtNames
+                    }, function (err, results) {
+                        res.send(err || arg);
+                        return (cb ? cb(err || arg) : true);
+                    });
                     emittedError = true;
                 } else {
                     res.send(arg);
--- a/lib/server.js
+++ b/lib/server.js
@@ -811,7 +811,6 @@ Server.prototype._run = function _run(req, res, route, chain, cb) {
     var log = this.log;
     var self = this;
     var handlerName = null;
-    var errName;
     var emittedError = false;
     var reqClosed = false;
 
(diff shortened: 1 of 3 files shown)


lib/server.js
```diff
@@ -826,12 +825,32 @@ Server.prototype._run = function _run(req, res, route, chain, cb) {
 
         if (arg) {
             if (arg instanceof Error) {
-                errName = arg.name.replace(/Error$/, '');
-                log.trace({err: arg, errName: errName}, 'next(err=%s)',
-                    (arg.name || 'Error'));
 
+                var errName = arg.name.replace(/Error$/, '');
```
micahr: This variable was hoisted before, any reason to bring this in instead of leaving it at the top?

micahr:
:+1: 

Nice work @DonutEspresso

yunong:
Looks good. @DonutEspresso can you please update the docs for this?

lib/server.js
```diff
@@ -826,12 +825,32 @@ Server.prototype._run = function _run(req, res, route, chain, cb) {
...
+                var errEvtName;
+
+                // if we have listeners for the specific error
                 if (self.listeners(errName).length > 0) {
-                    self.emit(errName, req, res, arg, once(function () {
+                    hasErrListeners = true;
+                    errEvtName = errName;
+                }
+                // or if we have a generic error listener
+                else if (self.listeners('error').length > 0) {
```
yunong: What's the semantic here if someone has both the general and specific error listener registered? I feel like we'll want to fire both listeners. In this case we're not actually doing that.

lib/server.js
```diff
@@ -826,12 +825,32 @@ Server.prototype._run = function _run(req, res, route, chain, cb) {
...
+                    hasErrListeners = true;
+                    errEvtName = 'error';
```
yunong: Don't use 'error' as the name, that's generally reserved for uncaught exceptions in node. Instead use something like 'server-error', or something to that effect.

lib/server.js
DonutEspresso: As envisioned originally we'd only fire one event, whatever was the most specific match found. 

The generic error event would only ever fire if you didn't register a specific handler for that error type. Thinking about it now, it seems more reasonable to _always_ fire the generic event. 

I think we should fire the most specific one first, then the generic listener. Sound reasonable?

lib/server.js
DonutEspresso: Sure, maybe just 'genericError'? 'expectedError'? Between uncaughtException and InternalServer I'm hesitant to add 'server*', could easily get confusing.

lib/server.js
yunong: Sounds good. :+1:

lib/server.js
yunong: maybe just ```restify-error```. generic or expected aren't that specific.

DonutEspresso: I went with 'nextError', but happy to change it to 'restifyError'. I thought it might reduce the number of questions we get around the difference between InternalError vs InternalServerError vs uncaughtException, but maybe that's a docs issue.

yunong: restifyError sounds good. +1 on updating docs to better reflect the
different classes of errors.

DonutEspresso:
Now uses vasync to fire all event listeners in serial. 

Event listeners can choose to return a completely different type of error, which vasync will just pass up the chain, going back to the client.

yunong:
Why not just fire them all at once?

DonutEspresso:
Like parallel? Not sure, I was thinking it might be nice to give people hooks to handle it at the most specific level first. i.e., do something, then say `err.handled = true`, then fire the generic one after, and they'd be able to determine if they need to do something with it.

DonutEspresso:
@yunong/@micah, we never closed the loop on the order of precedence for these error events. I'm still in favor of emitting in serial for the reasons stated above, any objections?

yunong:
Still not groking why we need to fire them serially. Presumably consumers of listeners will just listen for whatever error they care about, and any events that aren't listened for are just no-ops.

DonutEspresso:
Here's the use case for serial, if you have a generic listener in conjunction with specific listeners:

```js
// generic error handler
server.on('restifyError', function(req, res, err, cb) { 
  // so some generic metrics/logging
});

// handle 500s
server.on('InternalServer', function(req, res, err, cb) {
  // user could annotate errors here by saying err.handled = true,
  // because we must ALWAYS fire the generic handler above for all errors
});

// in some route, create a 500
  return next(new InternalServerError('oh noes!'));
  // this will hit the InternalServerError FIRST, allow user to handle it some fashion,
  // before firing restifyError event. the semantics of the generic handler means it should 
  // always be fired, but doesn't mean we shouldn't allow users to handle it first within
  // the error handler they care about. this is only possible if we fire events in serial. 
});
```

micahr:
I'd vote for serial to allow the handling in the most specific to least specific preference.

DonutEspresso:
Moving to #944 for target 5.x branch

PR #903, opened 2015-09-10
Update error handling documentation

### LINKED ISSUE #890: Update docs to reflect how to set custom content for errors
The example on the docs page is as follows:

```js
server.on('InternalServerError', function (req, res, err, cb) {
  err._customContent = 'something is wrong!';
  return cb();
});
```

But that doesn't actually work, as of 4.0 the response looks for a `body` property to render error objects, so the following is what actually works:

```js
server.on('InternalServerError', function(req, res, err, cb) {
    err.body = 'hello world';
    return cb();
});
// --> 'hello world' is returned to client instead of serialized error object
```

We should either update the docs to reflect this change, or update res to support another field which can be converted into body.

jquatier:
Fixes #890. I also noticed that the event that fires for errors is just `InternalServer` and not `InternalServerError`, so that's been updated as well. Confirmed that the tests also listen for that event.

### DIFF
--- a/docs/index.restdown
+++ b/docs/index.restdown
@@ -502,7 +502,7 @@ Alternatively, restify 2.1 supports a `next.ifError` API:
 
 Sometimes, for all requests, you may want to handle an error condition the same
 way. As an example, you may want to serve a 500 page on all
-`InternalServerErrors`. In this case you can add a listener for this error that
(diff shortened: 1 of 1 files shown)


micahr:
Thanks for updating the docs.

Issue #910, opened 2015-09-17
Unable to override Content-Type


honzajavorek:
I want to set no `formatters` for my API as it is going to accept and send various combinations of media types, not just one or two. I want to handle media types directly when constructing the response. However, I am unable to override the default Content-Type:

```javascript
res.header('Content-Type', 'application/hal+json');
res.send(200, JSON.stringify({
  _links: {
    self: {href: '/'},
  }
}));
next();
```

This results in

```
Content-Type: application/octet-stream
```

I am able to override it using the native methods, though:

```javascript
res.writeHead(200, {
  'Content-Type': 'application/hal+json',
});
res.write(JSON.stringify({
  _links: {
    self: {href: '/'},
  }
}));
```

Also, in documentation there is

>  Of course, you can always explicitly set the content-type:
> ```
> res.setHeader('content-type', 'application/foo');
> res.send({hello: 'world'});
> ```

I think this is strange and it cannot work.

1. `setHeader` is not documented anywhere in Restify docs. There is `header` function.
2. Without setting `formatters`, `{hello: 'world'}` won't get rendered as JSON or anything else. There should probably be `JSON.stringify({hello: 'world'})` in the example.

DonutEspresso:
FYI, `setHeader` is the native method on the response object (https://nodejs.org/docs/v0.4.0/api/http.html#response.setHeader)

Regarding the behavior, if a formatter isn't found for the specific content-type you've specified, it falls back on `application/octet-stream`. So you'd want to define all the different content-types you want to serve when you create the server, and then setting your header would point restify at the correct formatter to use.

DonutEspresso:
Closing this issue since it's not a bug - let us know if you have other questions.

PR #918, opened 2015-09-21
should fail to set content-type header when formatter is missing

DonutEspresso:
Adding a test to verify behavior per #910


micahr:
:+1:

yunong:
lgtm thanks!

PR #923, opened 2015-09-24
add emit log in auditlog.js, add ringbuffer to store certain amount of logs

leitsubomi:
Add emit log in auditlog.js
Add ringbuffer to store certain amount of logs logMetrics.js
Unit tests plugin.test.js
Add unit test in server.test.js to verify restify behavior when user creates duplicate route path.


lib/plugins/audit.js
```diff
@@ -25,6 +25,10 @@ var HttpError = require('restify-errors').HttpError;
 function auditLogger(options) {
     assert.object(options, 'options');
     assert.object(options.log, 'options.log');
```
yunong: Assert for the server object here.
```js
assert.object(options.server, 'options.server');

File: lib/plugins/audit.js [hunk 2]
```diff
@@ -25,6 +25,10 @@ var HttpError = require('restify-errors').HttpError;
 function auditLogger(options) {
```
yunong: Update the docs to reflect your API changes.

lib/plugins/audit.js
```diff
@@ -110,6 +115,14 @@ function auditLogger(options) {
 
         log.info(obj, 'handled: %d', res.statusCode);
 
+        if (logMetrics && logMetrics instanceof Object) {
```
yunong: You don't need these checks here if you just assert like the other objects up on line 27

lib/plugins/audit.js
```diff
@@ -110,6 +115,14 @@ function auditLogger(options) {
...
+            obj.timestamp = timestamp;
+            logMetrics.pushLog(obj);
+        }
+
+        if (server && server instanceof Object) {
```
yunong: same here, no need to check, just assert this object up top.

lib/plugins/logMetrics.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+var bunyan = require('bunyan');
+
+function logMetrics(options) {
```
yunong: Classes should start with a capital letter. Let's call this MetricBuffer instead of logMetrics.

yunong: Also add jsdocs about this class.

lib/plugins/audit.js
```diff
@@ -25,6 +25,10 @@ var HttpError = require('restify-errors').HttpError;
 function auditLogger(options) {
     assert.object(options, 'options');
     assert.object(options.log, 'options.log');
+    //use server object to emit log data
+    var server = options.server;
+    //use logMetrics object to store certain period of log records
+    var logMetrics = options.logMetrics;
```
yunong: The LogMetric class is overkill. Just have the user pass in a ringbuffer object themselves, and directly use that instead.

lib/plugins/audit.js
yunong: Since this class is no longer just writing logs, change its name to `audit`  instead.

test/plugins.test.js
```diff
@@ -6,9 +6,12 @@ var fs = require('fs');
 var http = require('http');
 var net = require('net');
 var path = require('path');
+var util = require('util');
 
 var bunyan = require('bunyan');
 var restifyClients = require('restify-clients');
+var assert = require('chai').assert;
```
yunong: Please sort imports by alphabetical order.

lib/plugins/audit.js
```diff
@@ -91,6 +95,7 @@ function auditLogger(options) {
 
     function audit(req, res, route, err) {
         var latency = res.get('Response-Time');
+        var timestamp = new Date().toString();
```
yunong: Don't use the fully qualified date object here, this will make it hard to parse. Instead, give it ns since epoch

lib/plugins/audit.js
```diff
@@ -20,11 +20,29 @@ var HttpError = require('restify-errors').HttpError;
...
+    assert.optionalObject(options.logBuffer, 'options.logBuffer');
+    assert.optionalBool(options.printLog, 'options.printLog');
+    //use server object to emit log data
+    var server = options.server;
+    //use logMetrics ringbuffer to store certain period of log records
+    var logMetrics = options.logBuffer;
+    //default always print log
+    var printLog = options.printLog;
+
+    if (typeof printLog === 'undefined') {
```
yunong: this can just be `if (!printlog)`

yunong:
@leitsubomi Looks great. Can you update CHANGES.md with your changes as well as docs/index.restdown?

DonutEspresso:
Looking at this PR, I think we actually want this merged in against restify/plugins. That means we don't have to tie it in with the 5.x release. 

The tests for duplicate routes can probably be refiled as a separate PR, still against this repo.

PR #938, opened 2015-10-25
Fixes response hanging when no content-type header specified

### LINKED ISSUE #937: 504 when no formatter exists for 'accept' header
if a client makes a request with `accept` header set to a format for which the server has not formatter (example: `text/html` is not built in and the dev has not defined a formatter for it), the client will never receive any response and will eventually time out.

DonutEspresso:
Fixes #937


DonutEspresso:
Whoops, wrong target branch. Will open one against 4.x and one against 5.x

PR #939, opened 2015-10-25
GH-937: fix bug when missing content-type header results in response …

### LINKED ISSUE #937: 504 when no formatter exists for 'accept' header

DonutEspresso:
Fixes #937 for 5.x


yunong:
:+1:

PR #942, opened 2015-10-29
altered if to account for when body is null

lukebrdn:


DonutEspresso:
Thanks for the pull request! Any chance you can add some unit tests?

lukebrdn:
Happily, but I don't have much experience writing proper unit tests. Could you point me in the write direction?

DonutEspresso:
You can add a test to this file by copying/pasting one of the tests in the response unit test file:

https://github.com/restify/node-restify/blob/5.x/test/response.test.js#L77

Just have the server send a null response (or whatever value was triggering your original issue), and ensure that that client gets the response you'd expect.

DonutEspresso:
Based on the passing tests provided by @michaelnisi, I'm going to assume this was a problem caused by something else in your application. Please do let us know if you continue to see this; a code snippet would go a long way in helping us pinpoint the issue.

PR #943, opened 2015-10-31
Typos

azollyx:
I think I found two typos in http://restify.com/
1) new Error('blah) => missing closing quote
2) when a browser sends a from => a *form* I guess


DonutEspresso:
Thanks!

PR #944, opened 2015-11-05
GH-667: support generic error listener

DonutEspresso:
@yunong @micahr rebased onto 5.x branch.

### DIFF
--- a/docs/index.restdown
+++ b/docs/index.restdown
@@ -503,10 +503,12 @@ Alternatively, restify 2.1 supports a `next.ifError` API:
 
 Sometimes, for all requests, you may want to handle an error condition the same
 way. As an example, you may want to serve a 500 page on all
-`InternalServerErrors`. In this case, you can add a listener for the `InternalServer` error event that
-is always fired when this Error is encountered by Restify as part of a
-`next(error)` statement. This gives you a way to handle all errors of the same
-class identically across the server.
+`InternalServerErrors`. In this case, you can add a listener for the
+`InternalServer` error event that is always fired when this Error is
+encountered by Restify as part of a `next(error)` statement. This gives you a
+way to handle all errors of the same class identically across the server. You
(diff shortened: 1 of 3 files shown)


DonutEspresso:
Fixes #667 and #888.

yunong:
LGTM

goyney:
Is there an ETA on the 5.x release? Badly need the generic error catcher.

mauron85:
Hi there. I'm testing 5.x branch and generic error middleware.

And I must say that I'm confused, (but also my assumptions might be just wrong).

In 4.x branch there was possibility to log every error in `server.on("after")` event handler, which catches both internal restify errors like `ResourceNotFound` or `BadMethod` and generic throw errors (throw new Error).

```
    server.on('after', function (req, res, route, err) {
        if (err) {
            logger.error(err, 'Error returned from route %s', route.name);
        }
    });
```

But in 5.x behavior seems to be changed and now `err` param is null for restify internal exceptions (`ResourceNotFound`, `BadMethod`).

There is new event `restifyError` but it doesn't seem working for me (it didn't catch any kind of error).

I think behavior of 4.x where any kind of exception can be captured and logged seems better fit for me. I don't want to have group of server.on listeners to log every possible restify exception.

retrohacker:
@mauron85 sorry for the radio silence on this. It was appended to a closed PR and fell off my radar. Mind opening a new issue using our sniney new templates? :smile: 

An error first callback not returning an error seems like a bug, but I haven't drilled down too deep into this. Have a few things (including a documentation site) on our plate at the moment so any help you could provide in repro cases and root-causing would def help! :heart:

mauron85:
@retrohacker no problem at all. I understand there things we do for living, that are keeping us busy. I also need to find time to write some test case scenario, so you can take look on it. I'll create new issue then.

PR #945, opened 2015-11-05
Add compatibility with Node 4 for Http Status Codes

DonutEspresso:
Migrating this fix from 4.x


DonutEspresso:
NM, we should merge the entire 4.x branch.

PR #946, opened 2015-11-05
Merge4to5

DonutEspresso:
@yunong @micahr  merging latest of 4.x to 5.x


yunong:
We should probably update changes.MD with the changes.

PR #951, opened 2015-11-11
Allow responses to skip formatters. was (Add support for pre-gzipped application/json)

mramato:
If a valid `Content-Encoding` is already set on an `application/json` response, do not convert `body` to a string and instead return it as is. The use case is sending the response JSON that has already been gzipped rather than going through the gzip plug-in and wanting to provide the correct `application/json` `Content-Type`.

I apologize if there is a reason this isn't already supported (perhaps I'm doing something wrong on my side).  I wanted to add a test case, bit couldn't find where the existing json formatter tests lived, if you can point me to them, I'll be happy to add one.

Here's an example of what I am talking about, this causes the gzipped data to be incorrectly turned into a string in `4.0.3` but is fixed in my branch.  In my specific use case, I'm not gzipping on the fly and instead have a in-memory buffer of already gzipped JSON data.

```
function preGzip(req, res, next) {
    zlib.gzip(JSON.stringify(result), function (err, data) {
        if (err) {
            return next(err);
        }
        res.setHeader('Content-Type', 'application/json');
        res.setHeader('Content-Encoding', 'gzip');
        res.setHeader('Content-Length', data.length);
        res.send(data);
        next();
    });
};
```

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -254,7 +254,6 @@ Response.prototype.link = function link(l, rel) {
     return (this.header('Link', _link));
 };
 
-
 /**
  * sends the response object. convenience method that handles:
  *     writeHead(), write(), end()
@@ -266,6 +265,24 @@ Response.prototype.link = function link(l, rel) {
  * @returns  {Object}                           self, the response object
  */
 Response.prototype.send = function send(code, body, headers) {
+    return this._send(code, body, headers, true);
+};
+
+/**
+ * sends the response object as-is without formatting. convenience method that
+ *     handles: writeHead(), write(), end()
+ * @public
+ * @function send
+ * @param    {Number} code                      http status code
+ * @param    {Object | Buffer | Error} body     the content to send
+ * @param    {Object}                  headers  any add'l headers to set
+ * @returns  {Object}                           self, the response object
+ */
+Response.prototype.sendRaw = function sendRaw(code, body, headers) {
+    return this._send(code, body, headers, false);
+};
+
+Response.prototype._send = function _send(code, body, headers, format) {
     var isHead = (this.req.method === 'HEAD');
     var log = this.log;
     var self = this;
@@ -334,15 +351,16 @@ Response.prototype.send = function send(code, body, headers) {
         }
     }
 
-    if (body !== undefined) {
+    if (body === undefined) {
+        _cb(null, null);
+    } else if (format) {
         this.format(body, _cb);
     } else {
-        _cb(null, null);
+        _cb(null, body);
     }
 
     return (this);
 };
-
 
 /**
  * sets a header on the response.


DonutEspresso:
Thanks for the pull request, appreciate the time you took to dig into the issue! 

I'm not sure that this is how we want to solve the problem, as given your use case this could be an issue for every formatter (and not just JSON responses). Your use case is pretty unique, and I think it might make more sense for you to use gzip only on endpoints that you know aren't serving pre-gzipped content. You can move plugin usage into your `server.get` handlers array instead of keeping them in `server.use`.

mramato:
> Your use case is pretty unique, and I think it might make more sense for you to use gzip only on endpoints that you know aren't serving pre-gzipped content.

Just to be clear, I'm already doing this.  The problem is that if `Content-Type` is set to `application/json` for pre-gzipped content, the json formatter crashes. I'm okay if you want to fix the issue in a different manner, but this is definitely a bug in all of the default restify formatters.  Basically, it's impossible to use pre-gzipped content with any non-binary default formatter.

My use case is not as unique as you think, since serving precompressed content is extremely common when dealing with large tiled geospatial data sets on the web.

If you can provide some guidance on what you think the correct fix would be, I'd be happy to work on that elsewhere in the codebase.

DonutEspresso:
If the formatter is crashing (and in turn crashing your server) that is definitely a bug. I apologize if I misunderstood the problem, that wasn't clear to me. If you have a short example that exhibits this problem, that'd be a great starting point.

mramato:
Here's a complete standalone example to explain the problem I'm seeing.  It's not an actual crash but rather a data corruption issue.

```javascript

var data;

//Normally data comes from a large gzipped JSON file but for the
//sake of this example, I'm generating a small valid gzip payload here.
var zlib = require('zlib');
zlib.gzip('{ "hello": "world"}', function (err, result) {
    data = result;
});

//Begin actual code that triggers restify bug

server.get('/', function (req, res, next) {
    res.setHeader('Content-Type', 'application/json');
    res.setHeader('Content-Encoding', 'gzip');
    res.setHeader('Content-Length', data.length);

    // In master, data gets base64 encoded by the JSON formatter
    // and therefore decoding fails on the client.  This branch
    // updates the formatter to honor 'Content-Encoding'
    res.send(data);
    next();
});

server.listen(3000, function(){
});
```

yunong:
This is fair. We need to expose a feature or flag to let your response skip formatters all together. This seems like a reasonable feature.

mramato:
That sounds like a better solution. A random idea would be a new function off response, maybe named something like `sendRaw`?  I'd be happy to work on this if you can give me the high level details of the approach you would like want to take.

DonutEspresso:
I like the approach of having a parallel `sendRaw()` method. You could take the approach of extracting what's in `send()` today and have some common shared method, but only call the formatters in the final step for `send()`. Seems like a cleaner approach than tossing some flag on to the request object. 

@yunong @micahr, what do you guys think?

micahr:
`sendRaw` does sound a lot better for maintaining backwards compatibility.

mramato:
Thanks for the feedback, I'll look into adding `sendRaw`.

mramato:
I'm sorry for taking so long to get back to this.  I've reverted my old changes and added `Repsonse.sendRaw` as discussed. I didn't see an obvious place to add a test (nothing in `response.test.js` actually calls `send`.)  Point me in the write direction and I can add one if you want.

Thanks!

DonutEspresso:
Thank you! Yes, the tests need some organization for sure. :) I think `response.test.js` is a fine place to add tests. 

I just realized this original PR is filed against master, which is bit outdated. Unfortunately we haven't yet had a chance to reconcile that against 4.x or 5.x. Would you mind re-opening your branch against 5.x? Thank you!

mramato:
> Would you mind re-opening your branch against 5.x?

No problem.  It also looks like there are some failing tests I need to address (need to wait until I'm on my linux machine for that since it looks like tests don't run under windows (or at least requires more set up).

As a side note, is restify 5.x close to release?  I only ask since that's where this PR is going and I was hoping to take advantage of the new feature (but can monkey patch in the mean time).

Thanks again for reviewing.

DonutEspresso:
I want to say soon-ish, but the reality is probably whenever we can get the [cors plugin](https://github.com/restify/plugins/pull/10) landed, since 5.x removes all plugins from core. It's the last PR I'd like to complete before cutting 5.x. Unfortunately, I just haven't had the bandwidth to work on it recently.

mramato:
Opened #1025 as requested.  Thanks.

PR #960, opened 2015-12-02
Fix incorrect client websocket example

richardkiene:
The documentation was using a `req` variable that is not declared. Additionally the second anonymous function definition hides the `err`, `res`, `socket`, and `head` variables.


micahr:
Thank you for the PR, the docs definitely need some love and this is a good start.

PR #961, opened 2015-12-03
CORS partial wildcards and protocol

StefanoDalpiaz:
Hi, this PR adds some flexibility to the CORS configuration for allowed origins.

Currently, the cors middleware only allows two ways of defining an allowed origin:
* a "global" wildcard `*`
* a scheme + domain name string - e.g. `http://foo.bar`

With this additional logic, the following definitions are supported:
* a "partial" wildcard, useful for whitelisting subdomains - e.g. `http://*.foo.bar`
* a domain name without the protocol, in which case both the http and https versions will be allowed - e.g. `foo.bar` or `*.foo.bar`
* a regular expression, to allow for total control over the format of the allowed domains - e.g. `/foo[0-9]?\.bar/i`

### DIFF
--- a/lib/plugins/cors.js
+++ b/lib/plugins/cors.js
@@ -93,10 +93,10 @@ function cors(opts) {
     assert.optionalArrayOfString(opts.headers, 'options.headers');
 
     if (opts.origins) {
-        assert.ok(Array.isArray(opts.origins), 'options.credentials');
+        assert.ok(Array.isArray(opts.origins), 'options.origins');
         opts.origins.forEach(function (origin, index) {
             assert.ok(typeof origin === 'string' || origin instanceof RegExp,
-              'options.credentials[' + index + ']');
+              'options.origins[' + index + ']');
         });
     }
 
--- a/lib/plugins/cors.js
+++ b/lib/plugins/cors.js
@@ -35,16 +35,24 @@ var AC_EXPOSE_HEADERS = 'Access-Control-Expose-Headers';
 
 ///--- Internal Functions
 
+function toRegExp(str) {
+    if (str instanceof RegExp) {
+        return str;
+    }
+    var regexStr = str
+      // escape all the special characters, except *
+      .replace(/[\-\[\]\/\{\}\(\)\+\?\.\\\^\$\|]/g, '\\$&')
+      // replace all instances of * with a proper regex wildcard
+      .replace(/\*/g, '.*');
+
+    return new RegExp('^(https?://)?' + regexStr + '$', 'i');
+}
+
 function matchOrigin(req, origins) {
     var origin = req.headers.origin;
 
     function belongs(o) {
-        if (origin === o || o === '*') {
-            origin = o;
-            return (true);
-        }
-
-        return (false);
+        return o.test(origin);
     }
 
     return ((origin && origins.some(belongs)) ? origin : false);
@@ -88,7 +96,7 @@ function cors(opts) {
     cors.origins = opts.origins || ['*'];
 
     var headers = (opts.headers || []).slice(0);
-    var origins = opts.origins || ['*'];
+    cors.originPatterns = cors.origins.map(toRegExp);
 
     EXPOSE_HEADERS.forEach(function (h) {
         if (headers.indexOf(h) === -1) {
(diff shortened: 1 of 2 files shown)


Maxwell2022:
+1

DonutEspresso:
Thanks for the PR! Could you please add some unit tests for the new use cases you are supporting?

StefanoDalpiaz:
As requested, I have added some tests. I have also made a fix to the type assertions during plugin initialisation, so that the `origins` array can contain regular expressions and not only strings.

lib/plugins/cors.js
```diff
@@ -80,15 +89,22 @@ function matchOrigin(req, origins) {
 function cors(opts) {
     assert.optionalObject(opts, 'options');
     opts = opts || {};
-    assert.optionalArrayOfString(opts.origins, 'options.origins');
     assert.optionalBool(opts.credentials, 'options.credentials');
     assert.optionalArrayOfString(opts.headers, 'options.headers');
 
+    if (opts.origins) {
+        assert.ok(Array.isArray(opts.origins), 'options.credentials');
```
DonutEspresso: I think the string here should be 'options.origins'. Also, any reason not to use `assert.array`?

lib/plugins/cors.js
```diff
@@ -80,15 +89,22 @@ function matchOrigin(req, origins) {
...
+        opts.origins.forEach(function (origin, index) {
+            assert.ok(typeof origin === 'string' || origin instanceof RegExp,
+              'options.credentials[' + index + ']');
```
DonutEspresso: Same thing here, I think you mean `'options.origins[' + index + ']'`.

lib/plugins/cors.js
```diff
@@ -35,16 +35,25 @@ var AC_EXPOSE_HEADERS = 'Access-Control-Expose-Headers';
...
     var origin = req.headers.origin;
 
     function belongs(o) {
-        if (origin === o || o === '*') {
-            origin = o;
-            return (true);
-        }
-
-        return (false);
+        return o.test(origin);
```
DonutEspresso: I'm not convinced we want to convert all normal strings origins into a RegExp. Instead, I think we should maintain existing behavior, but inside `matchOrigin()` special case the RegExp scenario vs string scenario. That would reduce the scope of changes for only RegExp values passed in.

DonutEspresso:
Love seeing so many new tests! Definitely appreciate all the help we can get. I left some comments - please take a look at your convenience.

lib/plugins/cors.js
StefanoDalpiaz: That was a copy-paste error - thanks! :)
As for `assert.array`, I would use it but the problem is that the version of `assert-plus` in use (0.1.5) does not have that method available. I didn't want to mess with the dependencies in this PR, so I had to fall back to a more generic `assert.ok`. If you're ok with that, I can update `assert-plus` to the latest version (0.2.0), so that we can use the new method. I have checked it locally, and updating breaks one test, though it looks like it would be easy to fix.

lib/plugins/cors.js
StefanoDalpiaz: I was following that approach in the beginning. The problem is that for values like `*.example.com` , or also simply `www.example.com` (any protocol), I need to convert the string to a regex. So at least some values in the array need to be turned into a regular expression. If I do an `if (o instanceof RegExp) { ... }` (and I would need to do it more than once, considering the preprocessing), then the code starts to look messy, so I decided to turn every value into a regex instead. If you prefer, I can go back to my initial implementation and differentiate between previous implementation (when in format `http://fixed.url` or `*`) and new one (`*.any.subdomain`, `any.protocol`, or `/regex/`).

DonutEspresso: Fair point - I think I'd still prefer the differentiation due to subtle behavior changes that may not be captured in existing tests (plus the additional regex overhead). But I don't feel particularly strongly on this issue. @yunong, thoughts?

DonutEspresso:
@StefanoDalpiaz, I took the liberty of migrating your changes over to this [major rewrite of CORS](https://github.com/restify/plugins/pull/10) over at the plugins repo. and I've included all your regexp changes as well as the tests. Let me know if that's ok - I've added you to the contributors list. Please take a look and let me know if all looks well. I'm trying to consolidate various CORS thread that have been abandoned for some time, and trying to get them all fixed in one go. 

If you'd like the commit history to retain your commits, let me know and I'd be happy to let you migrate them over.

DonutEspresso:
Closing in favor of restify/plugins#10

PR #963, opened 2015-12-04
feat(response): redirect, set & docs enhancements

jameswomack:
# James' Response changes & fixes for Restify 5

:star: :star2: :facepunch:

* Make `set` behave in a dev-friendly way whether it's called with a key-val pair or with a collection of key-val pairs
* Make `redirect` emit a `redirect` event with the `redirectLocation`
* Human-friendly assertion strings in `set`, more inline with recent assertions such as that in `redirect`
* Improve Restify install time and eventual readability by removing unused dependencies **keep-alive-agent, backoff, verror** & **tunnel-agent**
* Document that `writeHead` emits `header`
* Test that `writeHead` emits `header`
* Grammar correction for a comment

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "Guillaume Chauvet",
     "Isaac Schlueter",
     "Jacob Quatier",
+    "James Womack",
     "Jonathan Dahan",
     "Josh Clulow",
     "Matt Smillie",
@@ -58,27 +59,23 @@
     "node": ">=0.10"
   },
   "dependencies": {
-    "assert-plus": "^0.1.5",
-    "backoff": "^2.4.0",
+    "assert-plus": "0.1.5",
     "bunyan": "^1.4.0",
     "csv": "^0.4.0",
     "escape-regexp-component": "^1.0.2",
     "formidable": "^1.0.14",
     "http-signature": "^0.11.0",
-    "keep-alive-agent": "^0.0.1",
     "lru-cache": "^2.5.0",
     "mime": "^1.2.11",
     "negotiator": "^0.5.1",
     "node-uuid": "^1.4.1",
     "once": "^1.3.0",
+    "qs": "^5.0.0",
     "restify-clients": "1.1.0",
     "restify-errors": "3.0.0",
-    "qs": "^5.0.0",
     "semver": "^5.0.1",
     "spdy": "^2.0.5",
-    "tunnel-agent": "^0.4.0",
-    "vasync": "1.6.3",
-    "verror": "^1.4.0"
+    "vasync": "1.6.3"
   },
   "optionalDependencies": {
     "dtrace-provider": "^0.6.0"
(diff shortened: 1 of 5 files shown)


jameswomack:
It looks like an unrelated test in server.test.js is failed in Node 4 & 5

```
✖ use - throws TypeError on non function as argument
TypeError: Cannot assign to read only property 'name' of function (arg, msg) {
            if (!arg) {
                _toss(msg, 'true', arg);
            }
        }
    at Function.AssertionError (assert.js:45:13)
    at Object.<anonymous> (/home/travis/build/restify/node-restify/test/server.test.js:205:22)
    at Domain.<anonymous> (/home/travis/build/restify/node-restify/test/lib/helper.js:73:24)
    at Domain.run (domain.js:221:14)
    at Object._ (/home/travis/build/restify/node-restify/test/lib/helper.js:65:15)
    at /home/travis/build/restify/node-restify/node_modules/nodeunit/lib/core.js:232:20
    at /home/travis/build/restify/node-restify/node_modules/nodeunit/deps/async.js:168:13
    at /home/travis/build/restify/node-restify/node_modules/nodeunit/deps/async.js:131:25
    at /home/travis/build/restify/node-restify/node_modules/nodeunit/deps/async.js:165:17
    at /home/travis/build/restify/node-restify/node_modules/nodeunit/deps/async.js:463:34
```

yunong:
Wow, thanks for this. Looks good to me. @micahr @DonutEspresso feel free to take a look.

DonutEspresso:
This looks great, thanks! 

The only thing I'm wondering is if the `redirect` event might make more sense on the server object than on the response object? I'm not sure what your original use case was, but I could imagine wanting to hook into a redirect event for metrics, and in that case maybe the server object seems easier to work with since it's longer lived? Though we could emit on both objects, too.

DonutEspresso:
Not officially, I don't think. Maybe @yunong has some context?

Alternatively, the server could listen to the event on res and basically re-emit it.

package.json
```diff
@@ -58,27 +59,23 @@
     "node": ">=0.10"
   },
   "dependencies": {
-    "assert-plus": "^0.1.5",
-    "backoff": "^2.4.0",
+    "assert-plus": "0.1.5",
```
micahr: What was the reason behind saving the exact dependency? As a library, we tend to want to be flexible for others to pull in any version that we support.

jameswomack: My thinking was "they've released 0.2 so I don't see them releasing more versions or 0.1—therefore saying *exactly* 0.15 communicates the best most accurate information". No problem with throwing the ~ in there though.

DonutEspresso: FWIW, I don't think carets work on 0.x releases anyway, cause of some escape clause thingy with 0.x (blame semver).

jameswomack:
Is `Server.prototype._run` a good place to proxy the response event to the server @DonutEspresso ?

DonutEspresso:
Yup, looks like the right place to do it.

jameswomack:
@DonutEspresso There's a seemingly unrelated "✖ audit logger anonymous timer test" failing. Other than that, I've added the redirect event to server as well.

DonutEspresso:
I just re-ran the test - the timer test fails periodically. LGTM, @micahr any last thoughts before we merge?

jameswomack:
:+1: :zap:

micahr:
LGTM, merge away.

DonutEspresso:
Awesome, thanks @jameswomack!

Issue #941, opened 2015-10-29
Response.send fails if body === null


lukebrdn:
Response.send attempts to format body on line 338 which throws a "Cannot read property toString of null"

DonutEspresso:
Hi, can you provide a code snippet?

DonutEspresso:
Closing based on passing tests. Please provide code samples if you continue to see this issue.

PR #965, opened 2015-12-09
Test sending of null body

michaelnisi:
This tests if sending `null` from the response—mentioned in #941 and #942—is handled properly. I do not see the issue filed by @lukebrdn. If the error is still occurring, please provide more context, so I can try to reproduce it.


micahr:
Thanks for the test case!

micahr:
Oh :disappointed: I was just happy to get a test only PR.

PR #970, opened 2015-12-17
Revert "Test sending of of null body"

DonutEspresso:
This reverts commit 5726e11b5cebe09c3c96d999bd9c37b764a959ea.


DonutEspresso:
FYI @micahr

micahr:
I don't think it hurts to keep the test in place.

DonutEspresso:
Fair enough - didn't want to add noise to tests but don't feel strongly either way.

PR #971, opened 2015-12-17
Fix "too few args to sprintf" for errors in static plugin

benurb:
Hello,

just a minor fix: If you use the serveStatic plugin and have urls like:
`http://localhost/public/dir%2Ffile.jpg` or even double escaped `http://localhost/public/dir%252Ffile.jpg` restify will return an internal error with the message above.

I think adding a `%s` as first parameter is the easiest way of fixing this.

Regards,
Ben


micahr:
Thanks for the PR, looks good to me.

PR #972, opened 2015-12-17
add nsp support

DonutEspresso:
- due to npm3, cannot generate shrinkwrap for non devDeps. so we will
  get vulnerabilities even in dev deps, which may be misleading.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,8 +19,10 @@
 #
 ESLINT		:= ./node_modules/.bin/eslint
 JSCS		:= ./node_modules/.bin/jscs
+NSP         := ./node_modules/.bin/nsp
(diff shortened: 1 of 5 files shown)


micahr:
The PR looks good, but I don't quite understand the comment. What is the reason behind not being able to include dev dependencies in shrinkwrap? I think that was fixed recently.

DonutEspresso:
It's that npm3 flattens the tree - so by default shrinkwrap includes devDependencies too. In this case, devDependencies (like nodeunit) have transitive dependencies on an old build of uglify-js, which is causing nsp to report a vulnerability. While technically true, it isn't a runtime path vulnerability, and so not particularly accurate as far as a server is concerned.

micahr:
That makes sense, thanks for the clarification.

PR #969, opened 2015-12-17
update dependencies, fix usage of assert.AssertionError

DonutEspresso:
Fixes #966, as well as the issue @jameswomack ran into when upgrading assert-plus.

Not sure how it ever worked in the first place, as assert.AssertionError is a constructor and should be called with `new`. I suspect something in core changed, since the tests run with no issue on 0.12.x.

Issue #966, opened 2015-12-14
Update to latest assert-plus


cavis:
Similar to restify/errors#15 ... I'm unable to run a restify server in strict mode.  Seems like upgrading to the latest `assert-plus` would fix this.

```
node --use_strict index.js
~/code/restify-test/node_modules/restify/node_modules/assert-plus/assert.js:41
                stackFunc = stackFunc || _assert.caller;
                                                ^

TypeError: 'caller' and 'arguments' are restricted function properties and cannot be accessed in this context.
    at _assert (~/code/restify-test/node_modules/restify/node_modules/assert-plus/assert.js:41:49)
    at Object.object (~/code/restify-test/node_modules/restify/node_modules/assert-plus/assert.js:141:9)
    at new RequestCaptureStream (~/code/restify-test/node_modules/restify/lib/bunyan_helper.js:70:12)
    at Object.createLogger (~/code/restify-test/node_modules/restify/lib/bunyan_helper.js:276:25)
    at Object.createServer (~/code/restify-test/node_modules/restify/lib/index.js:140:35)
    at Object.<anonymous> (~/code/restify-test/index.js:3:22)
    at Module._compile (module.js:425:26)
    at Object.Module._extensions..js (module.js:432:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:313:12)
```

cavis:
Thanks!  I think you also need to update your `restify-errors` dependency to `3.1.0` - otherwise it pulls in the old `assert-plus` module, and running a restify server fails with a similar error.

```
node --use_strict index.js
~/code/restify-test/node_modules/restify/node_modules/restify-errors/node_modules/assert-plus/assert.js:41
                stackFunc = stackFunc || _assert.caller;
                                                ^

TypeError: 'caller' and 'arguments' are restricted function properties and cannot be accessed in this context.
```

PR #973, opened 2015-12-18
Rev restify-errors so it uses the same version of assert-plus

micahr:
Update restify-errors to the latest version to assert-plus lines up.

Companion to #969 
Completes the fix for #966


DonutEspresso:
LGTM. Why'd we lock to a specific version again?

micahr:
Dunno :smile:

PR #976, opened 2015-12-21
Fix highlighting in sidebar by using a more specific title.

salehhamadeh:
The wrong list item was highlighting in the sidebar when a user scrolls to the second Server API heading or clicks on the link from the sidebar. This was fixed by changing the title to a more appropriate one, Node Server API.


micahr:
Thanks for the docs fix!

PR #978, opened 2015-12-21
move plugins into their own repo. 

DonutEspresso:
CORS gets to stay since router uses it.


DonutEspresso:
Waiting on restify/plugins#8

DonutEspresso:
Looks like we're good to go. PTAL

package.json
```diff
@@ -86,6 +87,7 @@
     "filed": "^0.1.0",
     "jscs": "^2.7.0",
     "nodeunit": "^0.9.0",
+    "restify-clients": "1.1.0",
```
micahr: Curious why clients is both a dev and normal dependency, shouldn't normal dependency win out?

DonutEspresso: Def unintentional - will fix.

DonutEspresso: Looks like clients and plugins are used only for tests, so I've moved them to dev deps.

micahr:
LGTM

PR #983, opened 2015-12-28
fixed #982 -- allow null to be passed to res.send()

### LINKED ISSUE #982: Cannot pass to null to res.send (includes possible solution)
When I pass `null` to `res.send`, I get a `Cannot read property 'constructor' of null` error.

Browsing though the source code, I found the issue in file `/lib/response.js` starting on line `275`:  
```javascript
    if (code === undefined) {
        this.statusCode = 200;
    } else if (code.constructor.name === 'Number') {
        this.statusCode = code;

        if (body instanceof Error) {
            body.statusCode = this.statusCode;
        }
    } else {
        headers = body;
        body = code;
        code = null;
    }
```

Seems like the conditionals do not account for a null value for the code parameter of the `send` method and assume that if the parameter is not `undefined`, it must have a `constructor` property, as it uses that to check if the parameter is a number.

Perhaps a better way of checking if the code parameter is a number is by using `typeof code === 'number'` as it seems more explicit and can handle `null` values and perhaps other edge cases. This technique is already being used on line `191` of `/lib/upgrade.js` to essentially check for the same thing.

Is there any reason why `code.constructor.name` makes more sense in this situation?

felixmc:
Implements technique already being used at `/lib/upgrade.js:275` to check for the same thing. This allows for `null` and other possible edge values to be passed into `res.send()`.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -274,7 +274,7 @@ Response.prototype.send = function send(code, body, headers) {
 
     if (code === undefined) {
         this.statusCode = 200;
-    } else if (code.constructor.name === 'Number') {
+    } else if (typeof code === 'number') {
         this.statusCode = code;
 
         if (body instanceof Error) {
--- a/test/response.test.js
+++ b/test/response.test.js
@@ -420,3 +420,18 @@ test('should not fail to send null as body', function (t) {
         t.end();
     });
 });
+
+
+test('should not fail to send null as body without status code', function (t) {
+
+    SERVER.get('/13', function handle(req, res, next) {
+        res.send(null);
+        return next();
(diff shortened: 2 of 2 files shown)


DonutEspresso:
Thanks for the PR! Can you add a unit test for this scenario in test/response.test.js?

felixmc:
Added unit test to the end of `test/response.test.js`, does it look good?

micahr:
Looks good @felixmc. Thanks for adding the test.

yunong:
lgtm

DonutEspresso:
Thanks for the help @felixmc!

PR #984, opened 2015-12-29
jsonp() compatibility with res.json()

sean3z:
This allows `res.json()` to continue functioning when used with the `restify.jsonp()` plugin/middleware.

###### Proposal
With this change, if the `application/javascipt` Content-Type is already added (as through the jsonp plugin), we don't add the `application/json` type when `res.json` is invoked.

###### Advantages
Basically, if you're exposing an API that can be hit through JSONP _and/or_ JSON, you can still use `res.json()`. The following example currently breaks prior to this change:

```js
// server
server.use(restify.queryParser());
server.use(restify.jsonp());

server.get('/endpoint', function (req, res, next) {
  res.json({x: 'meow'})
});

// client
request.get('/endpoint?callback=angular.callback._0', function  (err, res, body) {
  // server app fails due to typeof error
  // lib/response.js (line 117): TypeError [].split()
});
```

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -230,7 +230,9 @@ Response.prototype.header = function header(name, value) {
  * @returns  {Object}
  */
 Response.prototype.json = function json(code, object, headers) {
-    if (!/application\/json/.test(this.header('content-type'))) {
+    var type = /application\/j[son|avascript]/;
+
+    if (!type.test(this.header('content-type'))) {
         this.header('Content-Type', 'application/json');
     }
 
--- a/lib/response.js
+++ b/lib/response.js
@@ -230,10 +230,8 @@ Response.prototype.header = function header(name, value) {
  * @returns  {Object}
  */
 Response.prototype.json = function json(code, object, headers) {
-    var type = /application\/j[son|avascript]/;
-
-    if (!type.test(this.header('content-type'))) {
-        this.header('Content-Type', 'application/json');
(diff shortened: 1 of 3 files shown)


DonutEspresso:
The jsonp plugin already adds a content-type header, and since restify has a built in formatter for jsonp, just calling `res.send()` will automatically format the output correctly. `res.json()` isn't what you want in any case, since json and javascript are two different content types.

sean3z:
Yup, this is a proposal is primarily for developer convenience. For applications that have `res.json()` littered throughout their code base and later introduce the jsonp formatter. As it stands, if you attempt to use both, restify simply crashes - so this semi-corrects that also by allowing both to co-exist.

micahr:
I'm in favor of not conflating json and jsonp, as they are very different responses. @DonutEspresso's suggestion of using `res.send` is the best approach for ensuring that the response is exactly what you are intending, instead of having `res.json` perform differently based on query parameters / content-type.

sean3z:
@micahr @DonutEspresso Fair enough. I've made a slight change to the original direction by ensuring that `res.json()` continues to work (as expected) when the `jsonp()` plugin is used. As it stands, this functionality does not work.

DonutEspresso:
I believe this issue will be fixed once #987 lands - can you take a look?

sean3z:
Yup, looks like that will take care of it. Thanks!

PR #987, opened 2016-01-04
Content-Type should not contain multi-values. Closes #986

### LINKED ISSUE #986: Response formatter blows up when headers turn into arrays
https://github.com/restify/node-restify/blob/20f8fa8f0084bfc2e5ff3f981aa24edde32f6758/lib/response.js#L118

I ran into a strange scenario today that I can't understand.

    // Setting the default content-type, assuming we will override it in another middleware
    res.setHeader('content-type', 'text/plain')
    // At this point: this.getHeader('Content-Type') == "text/plain"

    // ... some time later, in another middleware
    res.json(200, { ok: true })
    // At this point: this.getHeader('Content-Type') == ["text/plain", "application/json"]

This is incredibly strange. I don't understand why I would end up with an array of content types. At any rate, the formatter doesn't know how to deal with this as its expecting strings. Yet another fun day with dynamic languages!

jocull:
Fixes https://github.com/restify/node-restify/issues/986

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -19,6 +19,17 @@ var InternalServerError = errors.InternalServerError;
 
 var Response = http.ServerResponse;
 
+/**
+ * Headers that cannot be multi-values.
+ * @see #779, don't use comma separated values for set-cookie
+ * @see #986, don't use comma separated values for content-type
+ * @see http://tools.ietf.org/html/rfc6265#section-3
+ */
+var HeaderArrayBlacklist = {
+    'set-cookie': true,
+    'content-type': true
+};
+
 
 ///--- API
 
@@ -177,17 +188,6 @@ Response.prototype.getHeaders = function getHeaders() {
     return (this._headers || {});
 };
 Response.prototype.headers = Response.prototype.getHeaders;
-
-/**
- * Headers that cannot be multi-values.
- * @see #779, don't use comma separated values for set-cookie
- * @see #986, don't use comma separated values for content-type
- * @see http://tools.ietf.org/html/rfc6265#section-3
- */
-var HeaderArrayBlacklist = {
-    'set-cookie': true,
-    'content-type': true
-};
 
 /**
  * sets headers on the response.
--- a/lib/response.js
+++ b/lib/response.js
@@ -25,7 +25,7 @@ var Response = http.ServerResponse;
  * @see #986, don't use comma separated values for content-type
  * @see http://tools.ietf.org/html/rfc6265#section-3
  */
-var HeaderArrayBlacklist = {
+var HEADER_ARRAY_BLACKLIST = {
     'set-cookie': true,
     'content-type': true
 };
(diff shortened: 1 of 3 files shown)


test/server.test.js
```diff
@@ -1813,6 +1813,23 @@ test('gh-779 set-cookie fields should never have commas', function (t) {
...
+    CLIENT.get('/content-type', function (err, _, res) {
+        t.ifError(err);
+        t.equal(Array.isArray(res.headers['content-type']), false,
+                'content-type header should not be an array');
+        t.equal(res.headers['content-type'], 'bar');
+        t.end();
+    });
+});
+
+
```
micahr: Can you add another testcase using `res.setHeader`, which should have the same functionality, but just to cover both cases?

jocull:
@micahr Added tests. Got whined at by ESLint so the indentation is a little strange.

lib/response.js
```diff
@@ -178,6 +178,16 @@ Response.prototype.getHeaders = function getHeaders() {
 };
 Response.prototype.headers = Response.prototype.getHeaders;
 
+/**
+ * Headers that cannot be multi-values.
+ * @see #779, don't use comma separated values for set-cookie
+ * @see #986, don't use comma separated values for content-type
+ * @see http://tools.ietf.org/html/rfc6265#section-3
+ */
+var HeaderArrayBlacklist = {
```
DonutEspresso: Since this is a constant, can you cap case it and move it to the top of the file? Under the `/// Globals` section.

DonutEspresso:
Thanks for the PR! LGTM.

sean3z:
I believe the `Set-Cookie` header can legitimately be set multiple times.

lib/response.js
jocull: Added!

DonutEspresso: Uppercase please, since we're not yet on ES6  :)

jocull: Sorry, uppercase how? `HEADERARRAYBLACKLIST`, `HEADER_ARRAY_BLACKLIST`, ...

DonutEspresso: `HEADER_ARRAY_BLACKLIST` looks great.

jocull: Done!

test/server.test.js
```diff
@@ -1813,6 +1813,44 @@ test('gh-779 set-cookie fields should never have commas', function (t) {
...
+            t.ifError(err);
+            t.equal(Array.isArray(res.headers['content-type']), false,
+                    'content-type header should not be an array');
+            t.equal(res.headers['content-type'], 'bar');
+            t.end();
+        });
+    });
+
+
+test('gh-986 content-type fields should never have commas'
```
DonutEspresso: Just noticed this test is identical to the one above - did you mean to add one for set-cookie?

micahr: He was adding a test that I requested, for the use of `res.header` vs `res.setHeader`

DonutEspresso: Whoops, I can reads.

DonutEspresso:
Thanks @jocull!

Issue #952, opened 2015-11-13
Case for text/plain formatter to return unconditional toString()


chb0github:
After quite a bit of research I discovered that 
'accept: text/plain' invokes the default handler (expected) but that handler tries to get too smart.

```
function formatText(req, res, body, cb) {
    if (body instanceof Error) {
        res.statusCode = body.statusCode || 500;
        body = body.message;
    } else if (typeof (body) === 'object') {
        body = JSON.stringify(body);
    } else {
        body = body.toString();
    }

    res.setHeader('Content-Length', Buffer.byteLength(body));
    return cb(null, body);
}
```

Now, if we look at the formatter for JSON:

```
function formatJSON(req, res, body, cb) {
    if (body instanceof Error) {
        // snoop for RestError or HttpError, but don't rely on
        // instanceof
        res.statusCode = body.statusCode || 500;

        if (body.body) {
            body = body.body;
        } else {
            body = {
                message: body.message
            };
        }
    } else if (Buffer.isBuffer(body)) {
        body = body.toString('base64');
    }

    var data = JSON.stringify(body);
    res.setHeader('Content-Length', Buffer.byteLength(data));

    return cb(null, data);
}
```

What we see is that, fundamentally, _text/plain_ and _application/json_ use the same code to produce JSON string data. Each is also inconsistent about handling errors.

I offer that each handler should call their respective functions:

```
function formatJSON(req, res, body, cb) {
   var data = body ? JSON.stringify(body) : {};
    res.setHeader('Content-Length', Buffer.byteLength(data));

    return cb(null, data);
}

function formatText(req, res, body, cb) {

    body = body ? body.toString() : '';
    res.setHeader('Content-Length', Buffer.byteLength(body));
    return cb(null, body);
}
```
Not only does this lead to a cleaner implementation but gives perfectly consistent (and polymorphic!) results to all objects passed in. This also means that object implementers can manipulate results by overriding toString();

I can make the change and submit a PR if it will actually be accepted

DonutEspresso:
Thanks for logging the issue! What you've proposed here seems pretty reasonable to me. Thoughts, @yunong @micahr? 

Given that it'll be a breaking change, would be great to get this in for 5.x.

micahr:
Seems reasonable to me as well. Look forward to the PR.

yunong:
@chb0github That looks good, please send along a PR with tests.

chb0github:
So, I implemented the changes but it exposed some problems.

[...]

PR #988, opened 2016-01-05
5.x new handlers

chb0github:
Ok, I created this pull request. There are 2 still failing. I included the stack trace in comments in the code to see that there is a cycle.

For those unaware:
https://github.com/restify/node-restify/issues/952


test/server.test.js
```diff
@@ -1109,6 +1127,8 @@ test('next.ifError', function (t) {
 
     SERVER.get('/foo/:id', function tester(req, res, next) {
         process.nextTick(function () {
+            // this object fails to become JSON because it's picking up the
+            // arguments of the call stack which includes itself
             var e = new RestError({
```
chb0github: This error picks up the arguments from the call which includes itself and thus creates a cycle. Other errors nested errors but this one doesn't.

test/server.test.js
```diff
@@ -1100,7 +1101,24 @@ test('gh-278 missing router error events (415)', function (t) {
     });
 });
 
+/*
 
+ var data = body ? JSON.stringify(body) : '';
```
chb0github: I should have removed this and made it a review comment

lib/formatters/index.js
```diff
@@ -1,4 +1,4 @@
-// Copyright 2012 Mark Cavage, Inc.  All rights reserved.
```
micahr: What was the reason behind removing the copyright? Did your IDE remove these automatically?

lib/formatters/json.js
```diff
@@ -8,30 +6,14 @@
...
-            body = {
-                message: body.message
-            };
-        }
-    } else if (Buffer.isBuffer(body)) {
-        body = body.toString('base64');
-    }
-
-    var data = JSON.stringify(body);
+    var data = body ? JSON.stringify(body) : '';
```
micahr: The assumption here is that `data` would be a valid JSON string, which would mean it would need to be: `'""'` in order to be a valid.

micahr:
Please run: `make prepush` and fix up the formatting / line length issues when you get a chance.

.gitignore
```diff
@@ -3,6 +3,7 @@ cover_html
 node_modules
 docs/*.html
 docs/pkg
+.idea
```
micahr: Please add this to your personal gitignore file rather than the project's.

test/server.test.js
micahr: Please do move this to the PR description.

lib/formatters/index.js
chb0github: I assumed that a copyright notice on an open source project was a leftover from a copy and paste. I can add it back. I mean, technically, I wouldn't have the right to make a copy for a fork without his permission.

lib/formatters/json.js
chb0github: I thought about that. A null or undefined body represents a body payload of nothing. What you suggest is 2 bytes where none before were. If the accept type is application/json but the function returns nothing at all, then really, the accept type doesn't matter. The way I look at it, this code should produce the same results in both directions: f(g(x)) = g(f(x))
```
var thing = JSON.parse(JSON.stringify(null));
console.log(thing); // produces null, good, so this is our f(g(x)) = g(f(x))

var thing = JSON.parse(JSON.stringify(''));
console.log(thing); // produces "\"\""   which isn't right
```
Upon testing my own ideas I don't like my code either. How about in this condition we set the data to null?

.gitignore
chb0github: Ok.

chb0github:
npm test runs that target before hand (it caught some formatting issues) -- not sure what you're seeing for formatting. Happy to change them though

lib/formatters/json.js
micahr: As long as what is returned is valid JSON then it shouldn't matter what the default is. `''` isn't valid JSON, which was why I added the comment.

micahr:
Take a look at one of the travis runs to see the formatting issues I was referring to: https://travis-ci.org/restify/node-restify/jobs/100262028

chb0github:
Oh, those are errors about the comments I put in the code not meeting the format standard. Yeah, I yanked those already so it won't be an issue.

lib/formatters/json.js
chb0github: Ok, I will verify it with testing and change it to null. That should work best.

chb0github:
Did you see the comments from the issue? The way the errors are constructed makes them produce a cycle when calling JSON.stringify. Posting this PR to show others and get some feedback on how to proceed.

Issue #993, opened 2016-01-18
serveStatic should have caching?


avimar:
I would have hoped that serveStatic would check the file modify time and reply with 304 not modified, but it doesn't seem to do this.
How hard might this be?

avimar:
The issue seems to be that res.header('Last-Modified') is set AFTER conditional_request is already run.

I monkey patched libs/plugins/static.js and made it an option, but I didn't write any tests for a PR...

    function serveFileFromStats(file, err, stats, isGzip, req, res, next) {
        if (err) {
            next(new ResourceNotFoundError(err,
                req.path()));
            return;
        } else if (!stats.isFile()) {
            next(new ResourceNotFoundError('%s does not exist', req.path()));
            return;
        }

        if (res.handledGzip && isGzip) {
            res.handledGzip();
        }

        var maxAge = opts.maxAge === undefined ? 3600 : opts.maxAge;
        res.cache({maxAge: maxAge});
        res.set('Content-Length', stats.size);
        res.set('Content-Type', mime.lookup(file));
        res.set('Last-Modified', stats.mtime);
        if (opts.charSet) {
            var type = res.getHeader('Content-Type') +
                '; charset=' + opts.charSet;
            res.setHeader('Content-Type', type);
        }

        if (opts.etag) {
            res.set('ETag', opts.etag(stats, opts));
        }
       
        var ctime = req.header('if-modified-since');
        if (opts.checkIfModified && Date.parse(stats.mtime) <= Date.parse(ctime) && (req.method=='GET' || req.method=='HEAD')) {
            res.send(304);
            next(false);
        } else {
            var fstream = fs.createReadStream(file + (isGzip ? '.gz' : ''));
            fstream.once('open', function (fd) {
                res.writeHead(200);
                fstream.pipe(res);
                });
            fstream.once('end', function () {
                next(false);
            });
        }
    }

retrohacker:
https://github.com/restify/plugins/pull/23 for reference

retrohacker:
Closing this as stale for now as the PR is no longer open and the plugins have migrated. This is still an exciting feature and we would be happy to reopen this and help get a PR merged into the main repo if anyone is interested in carrying it over the finish line :heart:

avimar:
@retrohacker Sorry, but I'm not willing to put in the time to *figure out* how to test this (the static file tests look rather cryptic). The original changeset is here... I hope someone can help merge this feature.

https://github.com/restify/plugins/pull/23/files

[...]

PR #994, opened 2016-01-18
option for static serve to check if-modified-since

avimar:
Mentioned in #993

### DIFF
--- a/lib/plugins/static.js
+++ b/lib/plugins/static.js
@@ -73,10 +73,16 @@ function serveStatic(opts) {
             res.set('ETag', opts.etag(stats, opts));
         }
 
-        var ctime = req.header('if-modified-since');
+        try {
+            if(opts.checkIfModified){
+                var requestTime = Date.parse(req.header('if-modified-since')) || 0;
+                var modifiedTime = Date.parse(stats.mtime);
+            }
+        } catch(e){
+            opts.checkIfModified=false;
+        }
 
-        if (opts.checkIfModified && ctime
-            && Date.parse(stats.mtime) <= Date.parse(ctime)
+        if(opts.checkIfModified && requestTime && modifiedTime<=requestTime
             && (req.method === 'GET' || req.method === 'HEAD')) {
             res.send(304);
             next(false);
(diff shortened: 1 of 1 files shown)


avimar:
Is there a reason why this build failed? It looks like a travis config issue.

micahr:
Looks like an eslint config issue. I've updated master with the appropriate change.

avimar:
I fixed some lint issues...

DonutEspresso:
It might need to - saw something similar filed against ecstatic a little bit ago:
https://github.com/jfhbrook/node-ecstatic/pull/179

DonutEspresso:
Thanks for the PR! Can you elaborate with the issue with windows, I'm not familiar.

avimar:
Date is still an issue in `node v4.3.0`...

but if you do (windows) `curl "http://localhost:8000/bower.json" -H "Connection: keep-alive" -H "If-Modified-Since: 275760-10-13`
You get: `{"code":"InternalError","message":"illegal access"}"` but it doesn't crash the node process since restify handles exceptions.

So - if someone sends a super malformed request, it fails. (It doesn't show up in my logs..) Does that matter?

DonutEspresso:
In 5.x the uncaughtException handlers are going to be off by default, so yes, this would crash the server.

avimar:
I had borrowed some of the code from here which will then also also crash:
https://github.com/restify/node-restify/blob/4.x/lib/plugins/conditional_request.js#L122
`if (Date.parse(mtime) <= Date.parse(ctime)) {`

but later, https://github.com/restify/node-restify/blob/4.x/lib/plugins/conditional_request.js#L162 is wrapped in a try/catch, I guess that's what we have to do?

This was pushed upstream in v8 to not throw an exception, but current versions of node still throw.

DonutEspresso:
Yeah, I think it's reasonable to wrap in try/catch like in the conditional request plugin. Thanks for digging into it!

langflashcards:
is it known when it can be expected ot be merged?

DonutEspresso:
I think once @avimar gets a chance to add the try/catch in there I'd be happy to merge this in. 

That said, I think we'd like if you could file this PR against the restify/plugins repo instead. For our next 5.x release, all major plugins will be removed from core.

DonutEspresso:
Closing in favor of restifyplugins#23

avimar:
This was supposed to be a backport until the next restify version...

PR #998, opened 2016-02-03
Emit 'after' event even if next is not called.

gerrard00:
I would like the audit logger to log all requests, even if a particular handler doesn't call next for some reason. I see that the general consensus seems to be just make sure that you call next, but I wonder if you'd accept a patch that ensures that the after event is emitted even if the handler doesn't call next.

To accomplish that, I added code to handle the ServerResponse finish event.

I have opened another PR with the same fix for 4.x: https://github.com/restify/node-restify/pull/997

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -698,6 +698,7 @@ Server.prototype.toString = function toString() {
  */
 Server.prototype._handle = function _handle(req, res) {
     var self = this;
+    var afterCalled = false;
 
     function routeAndRun() {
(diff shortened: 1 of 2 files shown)


Jokero:
Ping @DonutEspresso, @yunong, @micahr

gerrard00:
This implementation is actually bad, because it leaks event emitters. I will open a new PR tonight with the implementation I am using in production.

Issue #1003, opened 2016-02-06
Auditing Request Bodies


lgomez:
I've been dealing with this for a little while and finally figured it out. When using the audit logger, if you'd like to log the request and response bodies, make sure you pass the `body: true` option.

Example:
```javascript
server.on('after', restify.auditLogger({ log: log, body: true }));
```

DonutEspresso:
Thanks for the PR!

PR #1004, opened 2016-02-06
Adding small options table to audit logger

lgomez:
This is to  to clarify that one can pass to include request and response bodies in log output. Ref #1003


micahr:
Thanks @lgomez!

PR #1010, opened 2016-02-11
Update documentation

mgwalker:
Updates documentation to include the `charSet` property that the static plugin accepts as an option.


yunong:
Thanks so much for your contribution :+1:

PR #1012, opened 2016-02-13
Update docs for request-api

### LINKED ISSUE #1011: getLogger - Does it exist?
I `grep`d this repo, but could only find it in `docs/index.restdown`.

Would like to use it, but am getting this error:
> req.getLogger is not a function

Posted here also: http://stackoverflow.com/questions/35333233/restify-req-getlogger-is-not-a-function

lukealbao:
- Remove deprecated Request.getLogger() reference and add info to
  Request.log
- Fix link in Plugins.requestLogger

fixes #1011


micahr:
Thank you for the update to the docs!

PR #1016, opened 2016-02-22
Initial attempt to add cache clearing for a removed route.

lgomez:
Hi there,

I had a little time today to work on this. It's my first time contributing code here so bare with me.

As described here: restify/conductor#32, when removing a route through `server.rm` the cache is not cleared for the removed route and an error is thrown under some circumstances when the router attempts to process a removed route. In my case this happens specifically when the audit logger is on though it is possible it can occur under other circumstances.

I've added the fix below based on what I did within my own project but I have a few notes/questions:

1. The tests don't pass for me on a fresh clone. I get some linting errors (see below).
2. I am using slightly different code on my project. The main difference is that here I wanted to use your pattern using `[].filter(fun(){...}` as opposed to a `forEach`. Since I wasn't able to test, I wanted to ask you if this is acceptable before I go down the test rabbit hole.
3. I do not see a granular unit test for mounting or unmounting a route on the router. I know it's indirectly tested through the server.rm test but, in this specific case I'd need to add the audit logger to be sure the issue is fixed in the patch. Would it be appropriate if I added a completely new test using a new server instead of the `SERVER` instance in test/server.test.js`?

Output from `npm test` on a fresh clone (see note 1 above):
```bash
Projects/node-restify - [5.x] » npm test

> restify@4.0.3 test /Users/lgomez/Projects/node-restify
> make prepush

not found: gawk
not found: nawk
./node_modules/.bin/eslint '.'

/Users/lgomez/Projects/node-restify/lib/router.js
  348:5   error  "server" is not defined  no-undef
  349:26  error  "change" is not defined  no-undef
  352:7   error  "log" is not defined     no-undef

/Users/lgomez/Projects/node-restify/test/formatter.test.js
  109:9  warning  Unexpected "todo" comment  no-warning-comments

✖ 4 problems (3 errors, 1 warning)

make: *** [check-eslint] Error 1
npm ERR! Test failed.  See above for more details.
```

Thank you.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -346,6 +346,13 @@ Router.prototype.unmount = function unmount(name) {
         delete this.reverse[route.path.source];
     }
 
+    var cached = this.cache.values().filter(function (c) {
+      return (c.name === name);
+    })
+    if (cached) {
+      this.cache.del(cached.key);
+    }
+
     delete this.mounts[name];
 
     return (name);


micahr:
Thanks for the PR, Luis. I think it's a good idea to go down the route you are considering for clearing out the cache of specific items. The one change I'd make is to not clear out all routes that have the same name (in the value), because there can be several routes with the same name, but different params / versions, etc. Also, by looping through all of the cache values, you introduce another `O(n)` lookup for every unmount operation. 

You should be able to track the removed routes from the filter function a few [lines up](https://github.com/restify/node-restify/blob/5.x/lib/router.js#L338) which you can use to find cache keys from the [`find`](https://github.com/restify/node-restify/blob/5.x/lib/router.js#L403) method.

As far as the tests go, a more focused set of tests on the router mounting / unmounting would be great. If you don't mind adding tests for mount / unmount to `router.test.js` I would be very grateful. As far as your tests not running: I was able to get them to run on a clean version of the 5.x branch. It looks like yours are failing due to code in the router not passing eslint and the tests aren't even run. Can you double check that you don't have any modified code?

Thanks again for your work on this!

lgomez:
Cool. Thanks for the reply. I've been paying more attention to the code now. Lets use the following example:

I add a route:
```javascript
server.get('/test/:id', ...);
```
Resulting in the following being mounted:
```javascript
{ name: 'gettestid',
  method: 'GET',
  path: { /^\/+test\/+([^\/]*)$/ restifyParams: [ 'id' ] },
  spec:
   { path: '/test/:id',
     method: 'GET',
     versions: [],
     name: 'gettestid' },
  types: [],
  versions: [] }
```
I then call it twice like so:
```javascript
client.get('/test/1', function(err, req, res, data){/*...*/});
client.get('/test/2', function(err, req, res, data){/*...*/});
```
Which results in the cache looking like this:
```bash
CACHE KEYS
-----------
[ 'GET/test/2*application/octet-stream',
  'GET/test/1*application/octet-stream' ]

CACHE VALUES
-----------
[ { methods: [ 'GET' ],
    name: 'gettestid',
    params: { id: '2' },
    spec:
     { path: '/test/:id',
       method: 'GET',
       versions: [],
       name: 'gettestid' } },
  { methods: [ 'GET' ],
    name: 'gettestid',
    params: { id: '1' },
    spec:
     { path: '/test/:id',
       method: 'GET',
       versions: [],
       name: 'gettestid' } } ]
```
Now...
* `Router.prototype.find` works in the context of a request. It sets the cache using **values** which give us keys like the ones you see above.
* `server.rm` only takes in a name. You either set the name yourself or keep a reference of the name returned by mounting it (ie: `server[method]`) which is automatically generated for you.
* We want to remove all cache entries for a route with a given name. I understand one route name can yield more than one cache entry but we want all of those entries removed regardless of method, version, CT, etc because one can only remove by the set name.

I don't see how we can avoid looping through cache entries to delete the relevant ones. Does this make sense or am I missing something?

Thank you!

PS: since cache is set and read in the context of a request, the other option (which I haven't thought through in detail) could be using the route pattern to build another pattern that includes the components of the cache key (ie: `<method><route's pattern><version><ct>`) to look for all matching cache keys but this still involves a loop and it's probably more expensive (?).

Or... I'm not understanding it and just wasted everybody's time haha.

jwb:
So is removing the route going to update the cache properly, or will the caller have to reach in and do that itself?

DonutEspresso:
Yes, this landed in #1089, looks like the issue didn't get linked. It's available in the restify 5.x beta, but is was not merged down into the 4.x branch.

PR #1020, opened 2016-02-25
GH 1019

### LINKED ISSUE #1019: Remove next(err)
``` javascript
server.get('/oops', 
           function clientFacing (req, res, next) {
             if (successful) res.send(202);
             return next();
           },
           function taskHandler (req, res, next) {
             var err = new InternalServerError('Some other service failed');
             return next(err);
           });
```

If you want to use the `next(err)` pattern, you currently have to ensure that no previous handlers have called `res.send`. Otherwise, you will get an uncaught exception for attempting to set headers after they've been sent. This hides the actual exception you are meaning to catch.

lukealbao:
Sorry for the multiple commits – got confused with the merged commits from other users.

Fixes #1019


lukealbao:
Gonna resubmit a trimmer PR

PR #1025, opened 2016-03-08
Add Response.sendRaw for sending without formatting.

mramato:
This is a continuation of #951 but PR'd into the `5.x` branch as requested.  See that discussion for details.

Factor out `Response.send` into a private function, `Response.__send`, that is  called by both `Response.send` and `Response.sendRaw`

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -268,7 +268,6 @@ Response.prototype.link = function link(l, rel) {
     return (this.header('Link', _link));
 };
 
-
 /**
  * sends the response object. convenience method that handles:
  *     writeHead(), write(), end()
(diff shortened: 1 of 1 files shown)


DonutEspresso:
Thank you for the PR! If you could add some tests, LGTM.

mramato:
Will do.  Thanks.

DonutEspresso:
Hi @mramato, I've been trolling our open PRs and seeing what we can roll into 5x, which is happening this month. Do you think you'll be able to get tests in the next week or two? If not, I'd be happy to copy your branch into origin and take a stab at it.

mramato:
Sorry, I keep meaning to get back to this but have been swamped at work.  I will definitely try and get to it by the end of this week.  If that doesn't happen; you are more than welcome to take a stab at it.  Sorry for the delay.

DonutEspresso:
Thank you, we appreciate the help!

DonutEspresso:
Closed in favor of #1109, which adds unit tests.

PR #1026, opened 2016-03-08
next(err) does not call res.send on a finished res

### LINKED ISSUE #1019: Remove next(err)

lukealbao:
If an upstream handler closes the response successfully (or not), and
a downstream handler passes an unrelated error to `next`, an uncaught
exception will be thrown for trying to set headers that have already been
sent.

This commit alters the behavior of `next` such that its error-handling
path only calls `res.send` if the response is not finished yet.

closes #1019

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -877,11 +877,13 @@ Server.prototype._run = function _run(req, res, route, chain, cb) {
                         },
                         inputs: errEvtNames
                     }, function (err, results) {
-                        res.send(err || arg);
+                        if (!res.finished) {
+                            res.send(err || arg);
+                        }
                         return (cb ? cb(err || arg) : true);
                     });
                     emittedError = true;
-                } else {
+                } else if (!res.finished) {
                     res.send(arg);
                 }
                 done = true;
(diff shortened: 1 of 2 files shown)


micahr:
A response isn't finished once the headers have been sent, so it is still possible to send an error, just not set the Response Code. This change seems prone to surprising a developer as it's not necessarily that an "unrelated" error is passed to next, it just might be sent after the headers have been processed.

DonutEspresso:
@lukealbao, I think we came to a good conclusion over on #1019, where we should look at decoupling next() from res.send. This would prevent your original scenario from occurring, and give you flexibility to use events independent of the response. Going to close this for now in favor of that approach, let me know if your concerns aren't captured in the other thread. Thanks for the contrib!

PR #1028, opened 2016-03-10
fix #1024: add server option: 'ignoreUncaughtExceptions'

### LINKED ISSUE #1024: Add option to disable restify's uncaughtException handler
I would like to be able to disable (turn off) restify's handling for uncaught exceptions.

This is so the node server when used with --abort-on-uncaught-exception will core dump (as currently restify will handle the exception, report and error to that request and then continue on processing).

twhiteman:
Allows the user to disable the uncaughtException error handler by setting *ignoreUncaughtExceptions* when creating the restify server. The default behaviour is unchanged (i.e. uncaught exceptions are still enabled when this option is not supplied).

Note: I added tests for this, but it felt a little awkward as it had to use a separate process to run the restify server - this is because disabling the uncaught exception handler will cause node to abort the process when an uncaught error occurs.

### DIFF
--- a/docs/index.restdown
+++ b/docs/index.restdown
@@ -143,6 +143,7 @@ below (and `listen()` takes the same arguments as node's
 ||certificate||String||If you want to create an HTTPS server, pass in the path to PEM-encoded certificate and key||
 ||key||String||If you want to create an HTTPS server, pass in the path to PEM-encoded certificate and key||
 ||formatters||Object||Custom response formatters for `res.send()`||
+||ignoreUncaughtExceptions||Boolean||When true (default is false), turns off restify's uncaughtException handler||
 ||log||Object||You can optionally pass in a [bunyan](https://github.com/trentm/node-bunyan) instance; not required||
 ||name||String||By default, this will be set in the `Server` response header, default is `restify` ||
 ||spdy||Object||Any options accepted by [node-spdy](https://github.com/indutny/node-spdy)||
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # restify Changelog
 ## Current
+ - #1024 Added server option: 'ignoreUncaughtExceptions' to allow disabling
+         of the uncaughtException error handling
  - #903 Update docs to reflect new error handling, Jacob Quatier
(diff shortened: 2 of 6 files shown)


docs/index.restdown
```diff
@@ -143,6 +143,7 @@ below (and `listen()` takes the same arguments as node's
 ||certificate||String||If you want to create an HTTPS server, pass in the path to PEM-encoded certificate and key||
 ||key||String||If you want to create an HTTPS server, pass in the path to PEM-encoded certificate and key||
 ||formatters||Object||Custom response formatters for `res.send()`||
+||ignoreUncaughtExceptions||Boolean||When true (default is false), turns off restify's uncaughtException handler||
```
yunong: Can we reverse this so the default is to turn it off? I'd rather restify ship out of the box to _not_ swallow errors.

yunong:
@twhiteman thanks so much for doing this. Everything looks great -- just added a comment about reversing the defaults. I think we want to ship this so that by default we _don't_ catch exceptions.

twhiteman:
@yunong Whilst I myself would prefer this to default to on (for Joyent's usage), I think it would be unexpected for the bulk of Restify's users, as they'd be wondering why their server is now crashing (instead of responding with an error response) for any unexpected Restify errors.

yunong:
I think we would gate this behind a major version change -- and update the change notes.

twhiteman:
I could go for that (i.e. with a major version change and docs) - certainly makes sense with node domain already being deprecated.

Anyone else have thoughts for this? If not, I'll update the PR.

micahr:
I'd second the desire to have the default be not to catch exceptions.

DonutEspresso:
+1 this looks great. Thanks @twhiteman!

yunong:
@micahr I think this needs to merge into 5.x?

micahr:
Yes, 5.x would be the best place to put it.

yunong:
@twhiteman would you mind changing this to pull into the 5.x branch?

twhiteman:
Okay - I've created a new PR (#1034) to target the 5.x branch and enable this feature by default. Closing this one.

PR #1032, opened 2016-03-16
fix potential xss vector

### LINKED ISSUE #1018: Cross-site-scripting error
I tested my restify server with this:
https://localhost:3000/no5_such3_file7.pl?%22%3E%3Cscript%3Ealert(73541);%3C/script%3E

The returned data contains "<script>alert(73541);</script>" which some browsers will execute.

Simple solution is to change the ResourceNotFoundError at the end of the Router.prototype.find to something like this:

callback(new ResourceNotFoundError('%s does not exist', url.parse(req.url).pathname));

DonutEspresso:
Fixes #1018 for 5.x

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -604,7 +604,12 @@ Router.prototype.find = function find(req, res, callback) {
         }
     }
 
-    callback(new ResourceNotFoundError('%s does not exist', req.url));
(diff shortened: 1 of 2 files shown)


micahr:
Looks good, like the other one.

PR #1038, opened 2016-03-22
Capitalize the location header field for redirect

Kynde:
The http header fields are case sensitive according to the
RFC 2616 and 7230. This changes the 302 header field "location"
into "Location".


DonutEspresso:
Thank you for the contrib!

PR #1041, opened 2016-03-24
Documented rejectUnknown option of bodyParser

Jokero:


micahr:
Thanks so much for the docs fix!

PR #1047, opened 2016-03-28
Necessary changes to make the SPDY example work

fidian:
It looks like this example was not updated when auditLogger was moved to restify-plugins.  These changes are enough to make the example work again, assuming the user has already installed `restify-plugins` (`npm install restify-plugins`).


micahr:
Thanks @fidian our examples are likely woefully out of date, so any help they can get is much appreciated.

PR #1056, opened 2016-04-01
Request.absoluteUri updated to use correct protocol.

davidmarek:
Request.absoluteUri used this.secure to determine if the protocol should be https or http, but this.secure doesn't exist anymore.
Switched to calling this.isSecure() to check if the connection is secure.


DonutEspresso:
LGTM. Second set of eyes and let's get it in.

micahr:
LGTM as well

yunong:
Looks good, time for a merge

PR #1065, opened 2016-04-21
Documented the ability to specify RegExp for named parameter

Jokero:
https://github.com/restify/node-restify/blob/5.x/lib/router.js#L90

I am amazed at how many things are not documented in restify...

### DIFF
--- a/docs/index.restdown
+++ b/docs/index.restdown
@@ -188,6 +188,10 @@ URL-decoded before being passed to you.
        return next();
      });
 
+You can specify `RegExp` in parentheses for named parameter:
+
+     server.get('/hello/:name([a-zA-Z]+)', send);
+
 You can also pass in a [RegExp](https://developer.mozilla.org/en/JavaScript/Reference/Global_Objects/RegExp)
 object and access the capture group with `req.params` (which will not
 be interpreted in any way):
--- a/docs/index.restdown
+++ b/docs/index.restdown
@@ -188,9 +188,9 @@ URL-decoded before being passed to you.
        return next();
      });
 
-You can specify `RegExp` in parentheses for named parameter:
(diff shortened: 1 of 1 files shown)


Jokero:
But I can't use backslashes:
```js
server.get('/hello/:name(\\d+)', send);
```
Because `url` module convert them to forward slashes

retrohacker:
Hey @Jokero,

Yes, the restify documentation definately needs some love, as does the website design and overall branding of the project. It is absolutely on our radar. Some of the heavier technical challenges for 5.x will probably take precedence first followed closely by a documentation overhaul. The rationale is that restify is currently being used in production in a lot of business critical places, and the teams supporting and using restify have been using the code base as the definitive source of truth in lieu of proper documentation. This is working fine :tm: for now from the perspective of those heavily invested in the framework, but we understand this isn't ideal for many people and is hindering adoption. We welcome contributions like this one :heart: they are definitely a step in the right direction.

I'm going to try to validate and proof read this PR today, will let you know soon!

retrohacker:
Tested with:

```javascript
var server = require('restify').createServer({ name: 'foobar' })

server.get('/hello/:name([a-zA-Z0-9]+)', function(req, res, next) {
  console.log(req.params)
  res.send(200)
  return next()
})

server.listen(8080)
```

retrohacker:
Hmm, terribly sorry @Jokero, after reviewing I went to resolve the conflicts and it appears this file no longer exists. Part of the work I described above is already underway and it seems the project has moved out from under this PR. I'm going to close this for now, we can revisit later. :heart:

retrohacker:
I'll add this to the `FEATURE_REQUESTS.md` file for now

PR #1069, opened 2016-04-29
update eslint to 2.x

DonutEspresso:
Not sure why greenekeeper's PRs have a million commits in them. Hopefully this will make it work again.


micahr:
They don't, they list what changed in the dependency. It's a simple changelog of what happened in the dep, not what is changing in this app.

DonutEspresso:
Ohhh, I can reads. In any case, looks like we needed to fix .eslintrc, so after this the other greenkeeper PR's should work.

PR #1072, opened 2016-04-30
Fix documentation and examples related to versioned routes

irnnr:
The documentation is correct in so far that restify picks the highest version for a route
if no Accept-Version header is given.
However, the example shows wrong results in that it behaves the opposite way.


DonutEspresso:
Thank you!

PR #1074, opened 2016-05-03
Formatters

chb0github:
text and json formatters now do only that.

1. text/plain formatter will now invoke a toString() on the body if it exists. Otherwise the body will be empty
2. application/json formatter will attempt to find a toJson() method on the body and invoke it. If that isn't found then it will invoke JSON.stringify on the body.


DonutEspresso:
Closing in favor of #1084

PR #1075, opened 2016-05-05
GH-999: normalize NotFound error handler to flush response

### LINKED ISSUE #999: Custom handler => 404 will not send response
Running the following code with the latest restify (installed using NPM), if an user tries to load any page other than "/hello" or "/400", the web browser will be always waiting for response.

 
```javascript

server.get('/hello/', (req, res, next) => {
  res.send("hello");
});

server.get('/400/', (req, res, next) => {
  return next(new restify.errors.BadRequestError());
});

server.on('NotFound', (req, res, err, cb) => {
  console.log("in error handling of 404");
  console.log("did not send response during this handler");
  console.log("cb = " + cb);
  return cb();
});

server.on('BadRequest', (req, res, err, cb) => {
  console.log("in error handling of 400");
  console.log("did not send response during this handler");
  console.log("cb = " + cb);
  return cb();
});

server.listen(8181, function() {
});
```

It seems that although generally speaking, adding a handler for 4xx errors will not prevent restify from sending responses (for example accessing "/400" still works fine), the handling for 404 will be broken by adding our own level of error handling.

This lack of symmetry might not be a huge problem but now we have to write our own code to generate error messages for all HTTP failures, because otherwise they might not share the same format.

DonutEspresso:
Fixes #999

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -196,6 +196,7 @@ function emitRouteError(server, req, res, err) {
 
     if (server.listeners(name).length > 0) {
         server.emit(name, req, res, err, once(function () {
+            res.send(err);
             server.emit('after', req, res, null);
         }));
     } else {
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -1729,7 +1729,6 @@ test('error handler defers "after" event', function (t) {
             t.ok(res2);
             t.end();
         });
-        res.send(404, 'foo');
         return (cb());
     });
     SERVER.once('after', function () {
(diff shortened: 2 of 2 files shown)


DonutEspresso:
@yunong can you think of any reasons why 404 listener doesn't flush response for you automatically when all the other error listeners do? 

Or should we normalize other way and don't send any response by default and expect all users to send a response manually when they've defined a custom listener?

DonutEspresso:
Per offline discussion, let's normalize NotFound even to act like other error events, and revisit later down the road if we want to change how error handlers work.

PR #1081, opened 2016-05-10
Documentation: default content-type is "application/json" (not "application/octet-stream")

Jokero:


DonutEspresso:
Thanks for the PR!

I think something has changed with the negotiating process (maybe an updated negotiator?). application/json appears to be the new default.

PR #1085, opened 2016-05-11
Emit 'after' event even if next is not called.

gerrard00:
I would like the audit logger to log all requests, even if a particular handler doesn't call next for some reason. I see that the general consensus seems to be just make sure that you call next, but I wonder if you'd accept a patch that ensures that the after event is emitted even if the handler doesn't call next.

To accomplish that, I added code to handle the ServerResponse finish event.

This is a replacement for https://github.com/restify/node-restify/pull/998 which leaked event emitters.

@Jokero I fixed the error that I mentioned.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -699,6 +699,7 @@ Server.prototype.toString = function toString() {
  */
 Server.prototype._handle = function _handle(req, res) {
     var self = this;
+    var afterCalled = false;
 
     function routeAndRun() {
         self._route(req, res, function (route, context) {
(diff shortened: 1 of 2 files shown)


DonutEspresso:
Thanks for the PR! 

However, after speaking with @yunong we feel this isn't the right patch for core. While I understand the motivation, I think one of the assumptions here is that flushing a response is equivalent to finishing a request. That isn't strictly true, since you may often continue to do other things after the response has been flushed. 

I'd be happy to continue discussion in a new thread though. We can look at ways to expose the right hooks via server emitted events or some other mechanism that would allow you to achieve this through a userland handler or otherwise.

gerrard00:
Thanks for taking the time to review it. We were able to solve the problem to our satisfaction using middleware based on similar logic, so we don't have a pressing need. 

Thanks again!

Issue #115, opened 2012-04-06
restify.client handling of paths with spaces 


timoxley:
If you accidentally use the client to request a path with spaces in it, everything hangs.

Doesn't reach the route, but it also doesn't return a status code to the client either. Just dead.

What do you think best approach here would be? 

Some kind of url encoding like " " to "%20" etc? throw error? 500? 403? 404?

https://github.com/timoxley/node-restify/blob/path-with-spaces/test/client.test.js#L76

deepakshrma:
Use nodejs __querysrting__ module
```js
console.log("?"+querystring.stringify({a:'ds d', b:"sdsd"}));
```
Good handles all possible checks.

PR #1087, opened 2016-05-12
Update package.json to spdy ^3.3.0

kolbma:
Update spdy in the stable branch to get HTTP/2 support for restify.


kolbma:
Any idea why the test client.test.js
✖ GH-115 GET path with spaces
fails with stable node?

DonutEspresso:
Thank you for the pull request! 

We're actually looking to release 5.x before end of this month - if you can wait, that will come with the latest version of spdy and a bunch of bug fixes and improvements. If you're inclined to stay on 4.x, I'll need @yunong to chime in and take a look at why the timer tests needed to be updated to pass.

kolbma:
Ok thanks for the information.
I can use a 5.x release.
But the timer logic with "greater" is wrong because the timer resolution is too big for fast processors so that the instructions might have finished just in time with the timer.  So there is the need of greater-equals. A few code lines above there is already a comment about this timing issue and the resolution was to use an timer with 1100 instead of 1000. But greater-equal should be the correct fix.

DonutEspresso:
restify@4.1.0 was just released with latest spdy to address this issue.

Official 5.x coming in a week or so.

PR #1089, opened 2016-05-14
Removing route from LRU cache on unmount.

lgomez:
This is an update on a previous PR with all the latest updates from 5.x. This time tests pass.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -347,6 +347,14 @@ Router.prototype.unmount = function unmount(name) {
 
     delete this.mounts[name];
 
+    var cached = this.cache.values().filter(function (c) {
+        return (c.name === name);
+    });
+
+    if (cached) {
+        this.cache.del(cached.key);
+    }
+
     return (name);
 };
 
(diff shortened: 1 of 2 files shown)


DonutEspresso:
Thanks for the PR! We'd really appreciate it if you can add a test or two to validate the behavior.

lgomez:
I was trying to do so but, when `.dump()`'ing the cache from within the server.test.js file in the `rm` test, all I get is an empty array. Any ideas on how to test that? I thought the `'/foo/:id'` route would've been cached (after performing the test request but it's not.

I'll keep trying but, if you have any pointers, it'd be great.

lgomez:
Never mind. Figured it out.

lgomez:
So... My code was actually broken. Added the test and fixed it. Let me know if you need anything else. Thank you!

DonutEspresso:
Thank you, this looks great! LGTM, a second  👀  and let's get this in.

micahr:
Looks good to me.

DonutEspresso:
@lgomez thanks for the help! Much appreciated.

lgomez:
My pleasure. Thank you for all the work on Restify.

Issue #1096, opened 2016-05-20
4.x to 5.x migration


DonutEspresso:
This should go out with 5.x.

PR #1097, opened 2016-05-20
update changes.md, add migration guide

DonutEspresso:
Addresses #1096


DonutEspresso:
Going to merge this in first so we don't have to keep updating the old docs in the repo. Will continue to iterate on these.

Issue #959, opened 2015-11-30
matchedVersion() returns range instead of version on every call but the first


doodlemeat:
When calling req.matchedVersion() the second time, it returns the range. It should always return the version.
	server.get({
		"path": "/test",
		"versions": ["0.1.0", "0.1.1"]
	}, function(req, res, next) {

		// requested version (Unaltered Accept-Version header)
		req.version(); 

		// the matched version (This should only return a version, not a range)
		// problem is that this function returns the range on the second call to this route because of caching
		req.matchedVersion();
		
		next();
	})

DonutEspresso:
Here's the code for `matchedVersion()`:

```js
Request.prototype.matchedVersion = function matchedVersion() {
     if (this._matchedVersion !== undefined) {
        return (this._matchedVersion);
     } else {
        return (this.version());
     }
 };
```

If `matchedVersion()` is returning the same thing, it's because the router couldn't find a version match. What's the version specified by the request?

doodlemeat:
The thing is, matchedVersion ONLY works the first time I do a request to that endpoint.

doodlemeat:
BUMP!

	var restify = require("restify");

	var server = restify.createServer();

	server.get({
		"path": "/test",
		"version": "0.5.0"
	}, function(req, res, next) {
		res.send(200, {
			"req.version()": req.version(),

			// First response gives correct version
			// All responses after gives the same as req.version()
			"req.matchedVersion()": req.matchedVersion()
		});
	});

	server.listen(8080);

I think this should be considered a bug.

micahr:
This could be a bug in your example code, but you don't call `next` in your handler.

doodlemeat:
Calling next() is not the problem. calling next() is optional. And by calling next after the response already has isnt going to change what the actual response is. This is all the code that is needed.

DonutEspresso:
Looks like the problem was with caching in the router. It was returning early and never getting a chance to set the matchedVersion on the request object.

tvjg:
Sorry to bump a closed issue, but is there any chance of a backport to 4.x branch? I am willing to assist if desired.

DonutEspresso:
Hi, at this point we're not working on 4.x anymore given the imminent (this month) release of 5.x, but we're happy to take PRs if that's something you're interested in.

tvjg:
Thanks @DonutEspresso , I've opened PR #1098 .

PR #1098, opened 2016-05-20
GH-959: fixes issue where cached routes were not setting maxVersion on the req

tvjg:
Here is the fix for #959 cherry-picked onto the 4.x branch. Please let me know if it needs and additional work -- I noticed there were a couple of `console.log` statements in this test.


DonutEspresso:
Thanks @tvjg! Node6 is failing but that's okay - we'll tackle that in 5.x. Appreciate it!

PR #1100, opened 2016-05-21
Provide explicit error message for route not found

irnnr:
When testify does not find a matching route it just returns an InternalServerError. However, that could be anything and it is not clear it was caused by a non-matching route.

Provide a message to the error to make it easier to find these mistakes.


DonutEspresso:
Hello, thanks for the PR. Could you expand a little on the scenario that's causing the issue?

A 404 is typically sent back when the router cannot find a matching route, along with an error message. It'd be helpful to have a few tests that demonstrate the behavior that this PR is addressing. Thanks!

irnnr:
Hey, sorry for the delay in getting back.

The functionality is not changed by this PR, just making things a bit easier to find; I was trying to forward a request into/to another handler chain using `next('otherHandler')`. That failed for me with just a plain Internal Server Error without any additional details of what went wrong. 

So adding an error message will simply make it easier for others to know what's going on and the message can easily be searched in the source.

PR #1103, opened 2016-05-21
GH-1086: reuse incoming request id headers to set request id

### LINKED ISSUE #1086: Allow the request id header to be configured
Instead of assuming `request-id` or `x-request-id` as the headers for an incoming request id, allow the header to be configured via a property.

DonutEspresso:
Fixes #1086


micahr:
👍 Thanks for adding that so quickly!

PR #1104, opened 2016-05-21
GH-845: support both sync and async formatters

### LINKED ISSUE #845: 'after' event does not always fire before res.send() is complete. was (Response.send should handle formatter errors)
From #839, we should 
1) return the error from the formatter if it is an instance of http or rest error.
2) return internalerror if the error is not one of those 2 instances.

DonutEspresso:
Attempts to fix #845, but there are some gaps. Conversation will continue in that thread.


micahr:
Looks good to me, exactly as we drew it up 😄

Issue #1067, opened 2016-04-27
EventEmitter deprecation message - spdy


Jovaage:
The version of spdy used in restify. are making calls to `process.EventEmitter`.  
This has since been deprecated, I believe in nodejs v6, and it's now logged in the console.  
Could the dependency be updated to a later version of spdy?

danielfigueiredo:
Yeah, I also started getting this error when updated to nodejs v6.0.0
```
(node) v8::ObjectTemplate::Set() with non-primitive values is deprecated
(node) and will stop working in the next major release.

==== JS stack trace =========================================

Security context: 0x20a8f96c9fa9 <JS Object>#0#
    1: .node [module.js:568] [pc=0x2c69edaa7a4] (this=0x33d394e94751 <an Object with map 0x30a94d28491>#1#,module=0x449e1063679 <a Module with map 0x30a94d29301>#2#,filename=0x449e1063581 <String[117]: /Path/To?my?project/.../DTraceProviderBindings.node>)
  .....
==== C stack trace ===============================

 1: v8::Template::Set(v8::Local<v8::Name>, v8::Local<v8::Data>, v8::PropertyAttribute)
 2: node::DTraceProbe::Initialize(v8::Local<v8::Object>)
 3: node::DTraceProvider::Initialize(v8::Local<v8::Object>)
 4: node::DLOpen(v8::FunctionCallbackInfo<v8::Value> const&)
 5: v8::internal::FunctionCallbackArguments::Call(void (*)(v8::FunctionCallbackInfo<v8::Value> const&))
 6: v8::internal::MaybeHandle<v8::internal::Object> v8::internal::(anonymous namespace)::HandleApiCallHelper<false>(v8::internal::Isolate*, v8::internal::(anonymous namespace)::BuiltinArguments<(v8::internal::BuiltinExtraArguments)1>)
 7: v8::internal::Builtin_HandleApiCall(int, v8::internal::Object**, v8::internal::Isolate*)
 8: 0x2c69ec0961b
 9: 0x2c69edaa7a4
(node:3841) DeprecationWarning: process.EventEmitter is deprecated. Use require('events') instead.
```

DonutEspresso:
Hello, thanks for the report. To help quickly debug, which combination of restify/spdy is causing the deprecation warning?

danielfigueiredo:
I'm not using spdy and I also get the error. My server is a regular rest API that works with json payload and upload images.
You can easily reproduce the error by installing Nodejs version:
[...]

PR #1105, opened 2016-05-22
Update SPDY for node 6.x

bobvanluijt:
Very simple edit but this should solve it https://github.com/restify/node-restify/issues/1067


DonutEspresso:
Thanks! The issue being reported was on the existing 4.x release. 5.x will ship with latest out the door.

PR #1106, opened 2016-05-23
throw error on node failure to access a server's address

wilzbach:
I just happened to peek over the source code and this line seemed pretty weird.

Git blame goes to @mcavage. Why did you choice `169.254.0.1` and not any other IP or localhost?

https://github.com/restify/node-restify/commit/19ddb2c288ad158a0fbe025c7e6fc8525ea13b34

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -360,7 +360,7 @@ function Server(options) {
             str += ':';
             str += addr.port;
         } else {
-            str += '169.254.0.1:0000';
+            throw new Error("Restify couldn't find a valid address to bind to.");
         }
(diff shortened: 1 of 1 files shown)


retrohacker:
Thanks for this PR @wilzbach,

I agree that it is strange we are returning what appears to be a random ip address. I'm :-1: on throwing on a property access though.

It might be cleaner to `return undefined` instead.

retrohacker:
Closing this for now, will discuss on #1316

If we decide throwing makes sense we can re-open this PR.

PR #1108, opened 2016-05-24
changed error message in node6. fixes broken unit tests

DonutEspresso:


DonutEspresso:
Client tests have been moved out to restify-clients as of 5.x. This fix is valid only for legacy 4.x.

PR #1114, opened 2016-05-29
JSDoc fix: net code and headers optional as they are. fixes #1113

### LINKED ISSUE #1113: JSDoc inconsistency at `res.json()`
I just found out that the jsdoc is not correct for `Response.prototype.json`. My IDE notes that the function expects exactly 3 parameters what is wrong.

https://github.com/restify/node-restify/blob/5.x/lib/response.js#L243

mrcrgl:
I just found out that the jsdoc is not correct for `Response.prototype.json`. My IDE notes that the function expects exactly 3 parameters what is wrong.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -235,9 +235,9 @@ Response.prototype.header = function header(name, value) {
  *     res.send({hello: 'world'});
  * @public
  * @function json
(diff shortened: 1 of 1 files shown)


DonutEspresso:
Thank you!

PR #1122, opened 2016-06-02
rev dependencies

DonutEspresso:


DonutEspresso:
This also removes `http-signature` which now lives in restify-plugins and restify-clients.

PR #1124, opened 2016-06-03
GH-1123: add deprecation warnings for domain dependent features

### LINKED ISSUE #1123: Add deprecation warnings for features requiring domains
Using the `handleUncaughtExceptions` flag will now trigger deprecation warnings. 

This also disables `next.ifError`, and only attaches the function if the flag is true since it requires domains.

DonutEspresso:
Fixes #1123

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -12,6 +12,7 @@ var errors = require('restify-errors');
 var assert = require('assert-plus');
 var mime = require('mime');
 var once = require('once');
+var semver = require('semver');
 var spdy = require('spdy');
 var uuid = require('node-uuid');
 var vasync = require('vasync');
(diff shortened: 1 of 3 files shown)


lib/server.js
```diff
@@ -32,9 +31,8 @@ require('./response');
 ///--- Globals
 
 var sprintf = util.format;
-
+var maxSatisfying = semver.maxSatisfying;
```
micahr: maximum satisfaction

micahr:
LGTM

It may become necessary at some point to have a more centralized feature detection within Restify, especially if we are trying to support back to at least Node 0.12. Because the number of instances of checking `this.handleUncaughtExceptions` is pretty low, it isn't necessary now, but as the velocity of Node releases increases and interfaces continue to change, we might want to revisit how we handle this detection.

DonutEspresso:
Yup. Sounds reasonable to me.

PR #1126, opened 2016-06-03
GH-1128: don't serialize and send domain errors to client

### LINKED ISSUE #1125: New JSON formatter formats errors poorly
JSON formatter was recently refactored to be more bare bones. restify-errors constructors have custom toJSON implementations so they are unaffected, but this causes standard Error objects to be sent down the wire as empty objects:

```js
server.use(function(req, res, next) {
  res.send(new Error('boom'));
});
```

```sh
$ curl localhost:3000/
{}
```

### LINKED ISSUE #1128: Domain errors should be cleaned up before serialization
`res.send` should send back an empty response (as with standard Errors) in the event of a domain exception.

DonutEspresso:
Fixes #1128

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -402,8 +402,11 @@ function __send(maybeCode, maybeBody, maybeHeaders, maybeCallback, format) {
         return _flush(body);
     }
 
-    // if no body, then no need to format.
-    if (body === null || body === undefined) {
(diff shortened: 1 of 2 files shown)


DonutEspresso:
Reviewed by @yunong

PR #1130, opened 2016-06-03
GH-1129: always invoke res.send callback if provided

### LINKED ISSUE #1129: Sync formatters should invoke res.send callback
It should be possible to use a `res.send(obj, next)` signature with sync formatters. Call the callback if provided, always.

DonutEspresso:
Fixes #1129


DonutEspresso:
Reviewed by @yunong

PR #1132, opened 2016-06-04
Create a settings funcition

danieldms:
built to be accessed from any part of the system or the system itself,
to store your settings.


micahr:
This seems like something that would be better positioned as a plugin. If you are interested in making it into a plugin, please add it to: [restify/plugins](https://github.com/restify/plugins)

retrohacker:
Closing as stale, if this is still of interest, please open a PR against https://github.com/restify/plugins as suggested above and I would be happy to do a review :heart:

PR #1135, opened 2016-06-13
Feature: Ability to find a route by a path

jquatier:
I have a use case where I'm adding routes to the server dynamically after the server is up. I would like to be able to check if a route exists for a path before I add it, because if another route already has a similar path regex, the new route will not fire (the first route that matches the request is always the one returned from the router).

Example:

```javascript
// existing route
server.get('/users/:id', () => {});   // regex: /^\/+users\/+([^\/]*)[\/]*$/
// new route
server.get('/users/:myParamName', () => {});   // regex: /^\/+users\/+([^\/]*)[\/]*$/
```

In this case, I wouldn't want to add the new route because the existing route has the same regex even though the parameter names are different. I would like to throw an error that says a route at that path already exists on the server, and then output the details of that route.

I didn't see another way to do this since the `compileUrl` method in router.js method isn't exported and that's what translates a path with parameter names to a RegEx. So rather than copy that method, I created this new method to translate a path to a RegEx and compare it against the route collection. I made it a `findByPath` method so that it can return the route that had the same path.

The other way I thought about doing it was to do my own parsing of the path and then set that as the route name (the `mount` method in the router will return `false` for routes that have the same name). But, I thought it might be good to still be able to name my routes differently from the path.

Anyway, TLDR; let me know if you have any feedback! Thanks! 👍

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -577,6 +577,16 @@ Router.prototype.find = function find(req, res, callback) {
     ));
 };
 
+/**
+ * Find a route by path. Scans the route list for a route with the same RegEx.
+ * i.e. /foo/:param1/:param2 would match an existing route with different
+ * parameter names /foo/:id/:name since the compiled RegExs match.
+ * @public
+ * @function findByPath
+ * @param    {String | RegExp}    path      a path to find a route for.
+ * @param    {Object}             options   an options object
+ * @returns  {Object}             returns the route if a match is found
+ */
 Router.prototype.findByPath = function findByPath(path, options) {
     assert.string(path, 'path');
     assert.object(options, 'options');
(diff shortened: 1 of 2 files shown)


micahr:
Trying to kick travis into gear, going to close and reopen to see if that helps.

micahr:
PR looks good to me, thanks for adding the helper.

micahr:
@yunong or @DonutEspresso any concerns?

yunong:
If we can add some JSdocs then it looks good to merge.

DonutEspresso:
Agreed - a blurb in the docs with an example usage would be great. LGTM!

micahr:
Looks good with the JSDoc addition. I'd settle for filing an issue for adding docs to the new doc site once it becomes available.

PR #1153, opened 2016-07-09
Make user agent plugin remove content-length header on HEAD requests only for curl user agents

misterdjules:
This is a backport from https://github.com/restify/plugins/pull/40.

### DIFF
--- a/lib/plugins/pre/user_agent.js
+++ b/lib/plugins/pre/user_agent.js
@@ -41,11 +41,11 @@ function userAgentConnection(opts) {
 
         if (ua && re.test(ua)) {
             res.setHeader('Connection', 'close');
-        }
 
-        if (req.method === 'HEAD') {
-            res.once('header',
-                res.removeHeader.bind(res, 'content-length'));
+            if (req.method === 'HEAD') {
+                res.once('header',
+                    res.removeHeader.bind(res, 'content-length'));
+            }
         }
 
         next();
--- a/test/user_agent.test.js
+++ b/test/user_agent.test.js
@@ -1,0 +1,123 @@
+'use strict';
+
+var child_process = require('child_process');
+var http = require('http');
+
+var restify = require('../lib');
+
+var helper = require('./lib/helper');
+var plugins = require('../lib');
+
+var test = helper.test;
+var before = helper.before;
+var after = helper.after;
+
+var SERVER;
+var SERVER_PORT;
+var SERVER_ADDRESS = '127.0.0.1';
+var SERVER_ENDPOINT;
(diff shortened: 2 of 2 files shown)


yunong:
This looks good, could you please update changes.md?

misterdjules:
I'd be glad to do so, but I'm not sure how to update it.

First, the latest 4.x release is 4.1.1. but [CHANGES.md does not contain any info for this release](https://github.com/restify/node-restify/blob/4.x/CHANGES.md).

Then, should I add these changes in a section named `Current changes` or `Unreleased changes` in `CHANGES.md`? I've seen both used in the past, but this is my first contribution to restify itself, so I'm not sure what to do.

I did not find the process of updating `CHANGES.md` documented anywhere, but I may have missed it.

DonutEspresso:
Apologies for the delay, @misterdjules! I've updated changes.md with the latest release info. 

That said, it looks like is a breaking change, is that correct? I'm not sure how to best version this given 5.x is already being staged.

misterdjules:
@DonutEspresso 

> That said, it looks like is a breaking change, is that correct? 

Absolutely, my apologies for not realizing that in the first place, and thanks for double checking :)

> I'm not sure how to best version this given 5.x is already being staged.

It seems that it would be best to not land this in any 4.x version anyway. When is the first 5.x release scheduled?

DonutEspresso:
We are thinking of publishing a 5.x-beta this week. Whatever is on the 5.x branch right now is exactly what will go into that release. The only thing keeping us from doing a full 5.x release now is CORS support, since we've removed the 4.x implementation with something brand new, and we didn't want to launch with feature regression. :)  

It is also be possible to use restify-plugins with the 4.x release of restify (I've been doing that myself for some time, to pick up some of the new plugin features).

misterdjules:
> We are thinking of publishing a 5.x-beta this week. Whatever is on the 5.x branch right now is exactly what will go into that release. The only thing keeping us from doing a full 5.x release now is CORS support, since we've removed the 4.x implementation with something brand new, and we didn't want to launch with feature regression. :)

Sounds good, thank you for the info!

> It is also be possible to use restify-plugins with the 4.x release of restify (I've been doing that myself for some time, to pick up some of the new plugin features).

Absolutely. In my case I've been a bit reluctant to add a dependency on `restify-plugins` to my project just to avoid duplicating the `userAgent` plugin because the result of `npm install restify-plugins` takes around 12MB of disk space. It is a lot more space used for little benefit, but again, that's specific to my use case.

I'll close this PR then, thank you @yunong and @DonutEspresso for all the help!

PR #1158, opened 2016-07-25
Expose bodyReader plugin

rundef:
Hi,

I'm using **restify** to create an API gateway (it basically does authentication, acl and rate limiting) and it then dispatches incoming requests to other microservices. It's just forwarding the requests to a microservice - which is why I need to have access to the non-parsed request body.

### DIFF
--- a/lib/plugins/index.js
+++ b/lib/plugins/index.js
@@ -9,6 +9,7 @@ module.exports = {
     auditLogger: require('./audit'),
     authorizationParser: require('./authorization'),
     bodyParser: require('./body_parser'),
+    bodyReader: require('./body_reader'),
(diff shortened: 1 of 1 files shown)


micahr:
Plugins have been moved out of Restify for v5.x, which is being released soon. Can you make your PR against: [restify/plugins](https://github.com/restify/plugins) instead?

rundef:
Thanks I just did. Would be great to have this in v4.x too, in the meantime :)

micahr:
I've published v1.1.0 of Restify plugins. I believe you should be able to pull that in to unblock you for now, it will also prep you for the upgrade to Restify 5.

Issue #1171, opened 2016-08-23
Removing a single versioned route breaks routing for all versions


jquatier:
It looks like currently when you remove a single version of a route at runtime, the `reverse` mapping that the router class maintains internally gets removed, even though there may be other versions of that API with the same source path. This appears to be a problem with the reverse mapping itself, it assumes there will only be a single route per source path and method and doesn't take route versioning into account.

https://github.com/restify/node-restify/blob/5.x/lib/router.js#L352

Example project:

```javascript
const restify = require('restify');

var server = restify.createServer();
const routeOne = server.get({ path: '/hello/:name', version: '1.0.0'}, (req, res, next) => {
  res.send('hello ' + req.params.name);
  next();
});

const routeTwo = server.get({ path: '/hello/:name', version: '2.0.0'}, (req, res, next) => {
  res.send('hello 2.0 ' + req.params.name);
  next();
});

const routeThree = server.get({ path: '/hello/:name', version: '3.0.0'}, (req, res, next) => {
  res.send('hello 3.0 ' + req.params.name);
  next();
});

server.rm(routeThree); // remove the 3rd version, maintain the others.

server.listen(9111, function() {
  console.log('%s listening at %s', server.name, server.url);
});

```

If you run this application and then try to hit /hello/something, you'll get a stack trace because the reverse mapping is now undefined after the route was removed:

```
/node_modules/restify/lib/router.js:524
        res.methods = reverse.slice();
                             ^
TypeError: Cannot read property 'slice' of undefined
    at Router.find (/node_modules/restify/lib/router.js:524:30)
```

Anyway, this basically prevents users from removing versioned routes at runtime because the routing breaks after the first version is removed. You can't remove more than 1 either because the `unmount` method also results in a undefined reference for the same reverse mapping.

micahr:
That indeed does look like a bug. Instead of wholesale removing any route with the same name / route the version should be checked and only the appropriate version should be removed. I think this points out further deficiencies in how routes are stored in that unique routes aren't created for each version just each verb and path.

PR #1195, opened 2016-09-27
Update Router unmount to check if other versions exist

gcssabbagh:
Updated Router unmount() to check if there are other versions for a certain route to fix https://github.com/restify/node-restify/issues/1171

First open-source contribution here ever.. so hope I'm doing it right.

I used the existing Router findByPath method instead of an O(n) filter()

Cheers,
Georges

### DIFF
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -229,14 +229,14 @@ test('rm route and clear cached route', function (t) {
     });
 });
 
-test('rm one version of the routes, other versions should still work',
+test('GH-1171: rm one version of the routes, other versions should still work',
     function (t) {
-        SERVER.get({ path: '/hello/:name', version: '1.0.0'},
+        var routeOne = SERVER.get({ path: '/hello/:name', version: '1.0.0'},
             function (req, res, next) {
                 res.send('hello ' + req.params.name);
                 next();
             });
-        SERVER.get({ path: '/hello/:name', version: '2.0.0'},
+        var routeTwo = SERVER.get({ path: '/hello/:name', version: '2.0.0'},
             function (req, res, next) {
                 res.send('hello ' + req.params.name);
                 next();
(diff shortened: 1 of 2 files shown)


test/server.test.js
```diff
@@ -229,6 +229,33 @@ test('rm route and clear cached route', function (t) {
     });
 });
 
+test('rm one version of the routes, other versions should still work',
```
DonutEspresso: Can you prepend this test name with the issue it is fixing? 'GH-1171: rm one version...'

test/server.test.js
```diff
@@ -229,6 +229,33 @@ test('rm route and clear cached route', function (t) {
...
+
+        var routeThree = SERVER.get({ path: '/hello/:name', version: '3.0.0'},
+            function (req, res, next) {
+                res.send('hello ' + req.params.name);
+                next();
+            });
+
+        t.ok(SERVER.rm(routeThree));
+
+        CLIENT.get('/hello/whatever', function (err, _, res) {
```
DonutEspresso: Can we do a strict check here, and make a request specifically for both 1.0.0 and 2.0.0 to ensure they're both available? Should be able to add a header version to do that.

DonutEspresso:
Thanks for the PR, and congrats on your first contribution! :tada: 

I left some minor comments. I think it would also be helpful to add another test that ensures a 404 after removing all the routes, since that's functionality that's now gated behind the new check you added.

gcssabbagh:
Thanks @DonutEspresso for the feedback! I'm happy to contribute, loving Restify so far :)

I have added tests to look at specific versions (200 on the remaining versions, 400 on the removed version) and a test for 404 once all route versions are removed as you specified

DonutEspresso:
Cheers - this LGTM. 

@yunong can I get a another pair of eyes? Just in case I missed something.

yunong:
LGTM

PR #1198, opened 2016-10-04
note that this is the breaking change in 4.x

trentm:
This relates to https://github.com/restify/node-restify/pull/851
https://github.com/restify/node-restify/pull/924 is the equivalent change to the changelog in the 5.x branch.


DonutEspresso:
Thanks! FWIW, this is changing again in 5.x:

https://github.com/restify/node-restify/blob/5.x/4TO5GUIDE.md#sync-vs-async-formatters

We'll be publishing a new docs site with all this information up front when it goes live.

PR #1199, opened 2016-10-07
Updated examples to es6

amilajack:


yunong:
@amilajack Thanks for doing this. In general we don't advocate for arrow functions as these become anonymous and hard to debug post-mortem. Would you mind updating this PR to only include the const/let changes?

amilajack:
Fixed

yunong:
Great, thanks for your contribution!

PR #1209, opened 2016-10-25
support passing most qs options to restify.queryParser

### LINKED ISSUE #925: allow passing of qs options
Problem:

qs limits the number of array elements in a URL query to 20 - above that, it will be converted to object properties, see: https://github.com/hapijs/qs.
Highly undesirable for some cases.

> qs will also limit specifying indices in an array to a maximum index of 20. Any array members with an index of greater than 20 will instead be converted to an object with the index as the key.
> This limit can be overridden by passing an arrayLimit option:

if the options which can be passed here:

server.use(restify.queryParser({...}));

would allow for a qs section or some argument pass-through mechanism, qs could be configured by the restify user. Even a specific: arrayLimit option would work.

If this request is found useful, absolutely willing to make a pull request - not sure of the standard process here. It's such a minor thing that the whole pull request/review cycle may be a bit overkill?

(note: I am aware that it's not hard to replace the restify queryParser - so it's not that there is no solution. But would consider configuration options for the built-in parser the superior solution)

trentm:
Fixes #925

### DIFF
--- a/lib/plugins/query.js
+++ b/lib/plugins/query.js
@@ -5,6 +5,23 @@
 var qs = require('qs');
 var assert = require('assert-plus');
 
+var EXPOSED_QS_OPTIONS = {
+    allowDots: assert.optionalBool,
+    arrayLimit: assert.optionalNumber,
+    depth: assert.optionalNumber,
+    parameterLimit: assert.optionalNumber,
+    parseArrays: assert.optionalBool,
+    plainObjects: assert.optionalBool,
+    strictNullHandling: assert.optionalBool
+
+    /*
+     * Exclusions (`qs.parse` options that restify does NOT expose):
+     * - `allowPrototypes`: It is strongly suggested against in qs docs.
+     * - `decoder`
+     * - `delimiter`: For query string parsing we shouldn't support anything
+     *   but the default '&'.
+     */
+};
 
 /**
  * Returns a plugin that will parse the query string, and merge the results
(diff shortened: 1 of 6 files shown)


trentm:
@yunong @DonutEspresso poke. Thanks.

My intention is to:
- get this in and publish restify@4.2.0
- submit a PR to restify-plugins to allow passing the same options to `queryParser` there.

Yunong, if you recall, you and I discussed this work ages ago in private chat.

trentm:
Oh, also, I'd like to do the 'make cutarelease' Makefile task for node-restify.git as I did for restify/clients.git (https://github.com/restify/clients/commit/d691925d10bda2ad885b6de754be26af7975da5b)

lib/plugins/query.js
```diff
@@ -22,14 +39,33 @@ function queryParser(options) {
...
+     * compatibility for the restify 4.x stream while using the latest qs
+     * version, we need to maintain those defaults. Note that restify-plugins
+     * changes back to the pre-restify-4.x behaviour. See test/query.test.js
+     * for more details.
+     */
+    var qsOptions = {
+        plainObjects: true,
+        allowDots: true
+    };
+    Object.keys(EXPOSED_QS_OPTIONS).forEach(function (k) {
```
yunong: Not sure about the performance impact of Object.keys, but presumably you could use https://www.npmjs.com/package/lodash.foreach to optimize this.

lib/plugins/query.js
```diff
@@ -22,14 +39,33 @@ function queryParser(options) {
...
+        EXPOSED_QS_OPTIONS[k](options[k], k);
```
yunong: Mind dropping a comment to describe what's happening here?

trentm: done

yunong:
LGTM, feel free to merge.

trentm:
Thanks!

trentm:
restify@4.2.0 published with this

PR #1222, opened 2016-11-03
backport #1024: add server option to disable the uncaughtException handler

twhiteman:
Back porting the server option `handleUncaughtExceptions` which allows users to disable the restify uncaughtException handler.

Note that this option is **enabled** by default in 4.x (which means no functionality change in 4.x), whereas this option will become disabled by default in 5.x - as discussed in #1024.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## not yet released
 
+ - #1024 Add server option to disable the uncaughtException handler.
+
 ## 4.2.0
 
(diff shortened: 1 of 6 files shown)


yunong:
@twhiteman looks like style checks are failing, mind fixing those first?

twhiteman:
Sorry about that, I had missed that step - all good now.

yunong:
LGTM. Thanks!

PR #1249, opened 2016-12-13
always emit after event. capture request state in req._connectionState

DonutEspresso:


lib/server.js
```diff
@@ -1060,6 +1062,10 @@ Server.prototype._run = function _run(req, res, route, chain, cb) {
         } else {
             log.trace({err: err}, 'uncaughtException');
             self.emit('uncaughtException', req, res, route, err);
+            var uncaughtEx = new errors.UncaughtExceptionError(
```
DonutEspresso: Not sure I need to wrap this here - think I might remove this.

lib/server.js
```diff
@@ -1133,11 +1134,24 @@ function _finishReqResCycle(req, res, route, err) {
     // regardless of res status.
 
     // after event has signature of function(req, res, route, err) {...}
-    if (!res.finished && !req.clientClosed) {
+    if (!res.finished && !req._connectionState) {
```
yunong: Should probably explicitly check for `_connectionState === abort || closed` since when and if we add new states this might break.

DonutEspresso:
Updated with helper method

Issue #1254, opened 2016-12-22
Allow custom function to set incoming request id


DonutEspresso:
Currently, `createServer()` takes an array of request headers via `reqIdHeaders`, which allows users to reuse an incoming request id. In some scenarios this logic needs to be more complex (cascading, or require parsing of other headers, etc.). 

Allow `reqIdHeaders` to be a function which will be invoked with `req` and `res` objects, and returns a string value which will become the request id value set to `req._id`.

DonutEspresso:
On further discussion, we'll allow `req.id()` to become a setter, but for single use only. That way core gets out of the business of doing too much custom resolution of request id headers. Userland can use `pre()` handlers to set the id using whatever logic they want.

PR #1255, opened 2016-12-22
allow custom request id resolver

DonutEspresso:
I opted to introduce a new option since this new function resolver really has nothing to do with headers, so it's weird to reuse the`reqIdHeaders` option. It's currently just called `reqId` but open to naming suggestions.


DonutEspresso:
Reminder to self to update docs when this is done. Fixes #1254

DonutEspresso:
Closing, going to change direction on this slightly.

PR #1256, opened 2016-12-22
allow request id to be set. remove reqIdHeaders option

### LINKED ISSUE #1254: Allow custom function to set incoming request id

DonutEspresso:
Fixes #1254

### DIFF
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -13,7 +13,6 @@ var filed = require('filed');
 var plugins = require('restify-plugins');
 var restifyClients = require('restify-clients');
 var uuid = require('uuid');
-var validator = require('validator');
 
 var RestError = errors.RestError;
 var restify = require('../lib');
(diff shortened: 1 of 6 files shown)


test/server.test.js
```diff
@@ -44,8 +43,7 @@ before(function (cb) {
             dtrace: helper.dtrace,
             handleUncaughtExceptions: true,
             log: helper.getLog('server'),
-            version: ['2.0.0', '0.5.4', '1.4.3'],
-            reqIdHeaders: ['x-req-id-a', 'x-req-id-b']
+            version: ['2.0.0', '0.5.4', '1.4.3']
```
DonutEspresso: Going to move all these `reqIdHeaders` tests into plugins repo.

PR #1262, opened 2017-01-12
sanitizePath filters double-slashes in the queryString

fangel:
I stumbled upon a somewhat interesting bug in the sanitizePath pre-filter. It will continue it's sanitisation into the queryString-part of the URL.

The filter is meant to turn `foo///bar//` into `foo/bar`, but since the path it gets in also includes the QS, it will also turn `foo//bar//?baz=///` into `foo/bar?baz=/`. I doubt this is intentionally, as the contents of query-parameters often aren't subject to the same laxness as we assign urls.

So I have created a very simple fix, that simply tracks when the path transitions into the query-string part, and then skips any further filtering.

This is targeted against the 4.x branch, because I can see that this functionality has been moved to _restify-plugins_ in 5.x. But in my opinion this bug is important enough to have it applied to 4.x.
If this fix is accepted, I will gladly forward port the same changes to _restify-plugins_ so they can make it into 5.x as well.

Kind regards
Morten

### DIFF
--- a/lib/plugins/pre/pre_path.js
+++ b/lib/plugins/pre/pre_path.js
@@ -16,6 +16,7 @@ function strip(path) {
     var cur;
     var next;
     var str = '';
+    var inQS = false;
 
     for (var i = 0; i < path.length; i++) {
         cur = path.charAt(i);
@@ -24,8 +25,10 @@ function strip(path) {
             next = path.charAt(i + 1);
         }
 
-        if (cur === '/' && (next === '/' || (next === '?' && i > 0))) {
+        if (cur === '/' && (next === '/' || (next === '?' && i > 0)) && !inQS) {
             continue;
+        } else if (cur === '?') {
+            inQS = true;
         }
 
         str += cur;
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -1268,7 +1268,6 @@ test('GH-323: <url>/<path>/?<queryString> broken', function (t) {
     });
 });
 
-
 test('<url>/?<queryString> broken', function (t) {
     SERVER.pre(restify.pre.sanitizePath());
     SERVER.use(restify.queryParser());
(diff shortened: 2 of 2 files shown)


DonutEspresso:
Thanks for the bug report!

My reading of [RFC1738](https://www.ietf.org/rfc/rfc1738.txt) is that `/` is a reserved character, so they must be encoded in order to use them safely in URLs for any purpose than as part of the URL scheme. As an example, I believe `/` is also encoded if you're using encodeURIComponent.

fangel:
After reading through the specs, I agree with you that `/` is indeed a reserved character, and ought to be escaped by a standard-compliant client.

For those interested, it's in Section 3.3 of the RFC, which states the following:

> An HTTP URL takes the form: http://<host>:<port>/<path>?<searchpart>

and

> Within the <path> and <searchpart> components, "/", ";", "?" are reserved.

--

However, I still believe my proposed changes have some merit. I base this on the following arguments:

- Most end-users would probably be confused when a plugin that is labeled "sanitize(s) path" chooses to also sanitise the search-part of the URL. Case in point, I was very confused when it happened - and it took me a good while to track this down to the `pathSanitize` functionality, because I didn't think it would have anything to do with parsing query-parameters.

- It follows the guiding principle of "be lenient in what you accept and stringent in what you do". Just because a client forgets to properly encode the search-part/query-string of a query, I don't think a path-sanitising functionality should change how the query is parsed. 

If you are of a different opinion, then I totally respect preserving the existing behaviour as it does conform to the spec, in that it only operates on characters that are reserved for special use.

yunong:
@DonutEspresso any additional thoughts?

retrohacker:
Hey @fangel,

It looks like conversation around this PR has dropped off. In between the last comment and now, the plugins have moved to [restify/node-restify](https://github.com/restify/plugins). In addition, while surprising, it does appear that this behaviour conforms to spec.

Perhaps this would be better suited as an addition to the v5 documentation?

I'm going to close this for now, if you feel strongly about this, please open an issue our new plugins repo and we can work on getting it into documentation :smile:

PR #1263, opened 2017-01-13
return debug info about mounted routes and server

DonutEspresso:
WIP


DonutEspresso:
@yunong @rajatkumar this is ready for review.

PR #1264, opened 2017-01-13
Emit `close` event

jucrouzet:
Emit a `close` event when server is closed.
Can be useful for middlewares that needs this information to clear up things or stop timers.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -429,11 +429,14 @@ Server.prototype.listen = function listen() {
  * @returns  {undefined}
  */
 Server.prototype.close = function close(callback) {
+    var sendCloseEvent = this.emit.bind(this, 'close');
+
     if (callback) {
         assert.func(callback, 'callback');
     }
 
     this.server.once('close', function onClose() {
+        sendCloseEvent();
         return (callback ? callback() : false);
     });
 
--- a/lib/server.js
+++ b/lib/server.js
@@ -429,14 +429,14 @@ Server.prototype.listen = function listen() {
  * @returns  {undefined}
  */
 Server.prototype.close = function close(callback) {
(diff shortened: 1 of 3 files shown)


jucrouzet:
@yunong : done

lib/server.js
```diff
@@ -429,11 +429,14 @@ Server.prototype.listen = function listen() {
  * @returns  {undefined}
  */
 Server.prototype.close = function close(callback) {
+    var sendCloseEvent = this.emit.bind(this, 'close');
```
yunong: Can you please change this to just

```javascript
var self = this;

this.server.once('close', function onClose() {
    self.emit('close');
});
```

yunong:
@jucrouzet overall looks good, please just make that one last change. The `bind` function there is not needed if you just assign `this` to `self` up front.

jucrouzet:
Yep, sorry ... es6's fat arrow made me forget all the good reflexes ...
Done !

KostyaTretyak:
In the Server class we have this code:

```js
var PROXY_EVENTS = [
    'clientError',
    'close',
    'connection',
    'error',
    'listening',
    'secureConnection'
];

// ...
    PROXY_EVENTS.forEach(function (e) {
        self.server.on(e, self.emit.bind(self, e));
    });
```

So event 'close' already emits for restify's Server.

By the way, follow exclamation mark looks like bug:

```js
    if (!options.handleUpgrades && PROXY_EVENTS.indexOf('upgrade') === -1) {
        PROXY_EVENTS.push('upgrade');
    }
```
Or not?

DonutEspresso:
Yes, it does appear `close` is now being emitted twice in 5.x. We should revert this commit.

PR #1272, opened 2017-01-23
Added keywords to package.json

timmoy:
Hi, I was trying to find a good first bug and noticed that your 'package.json' doesn't use keywords. 

This fixes it to add keywords which will be picked up by npmjs and increase this project's visibility.

If you don't like the selection of keywords I've used, please let me know and I will change them as needed.


DonutEspresso:
Thanks! We definitely haven't septn much time on searching and visibility within npmjs' search.

PR #1281, opened 2017-02-24
Add debugInfo method to the server

yunong:


lib/router.js
```diff
@@ -633,3 +634,38 @@ Router.prototype.toString = function toString() {
...
+ * Return information about the routes registered in the router.
+ * @public
+ * @returns {object} The routes in the router.
+ */
+Router.prototype.getRoutes = function getRoutes() {
+    var self = this;
+    var routeInfo = [];
+    var routeName;
+
+    for (routeName in self.mounts) {
```
DonutEspresso: I forgot, are we using lodash? If so, `_.forOwn`

lib/router.js
```diff
@@ -633,3 +634,38 @@ Router.prototype.toString = function toString() {
...
     return (str);
 };
+
+
+/**
+ * Return information about the routes registered in the router.
+ * @public
+ * @returns {object} The routes in the router.
+ */
+Router.prototype.getRoutes = function getRoutes() {
```
DonutEspresso: This seems misnamed - it's not actually returning the routes, just a bucket of info about the routes. I remember we were thinking about implementing this as `toJSON()`?

docs/pages/components/server.md
```diff
@@ -89,9 +89,13 @@ via `use()` will run only after the router has found a matching route. If no
 match is found, these handlers will never run. Takes a function, or an array
 of functions.
 
-## unfinishedRequests()
+## inflightRequests()
```
DonutEspresso: I thought we settled on `unfinished` because the semantics of the core streams API uses `stream.on('finish', ...)`?

lib/server.js
```diff
@@ -1169,5 +1170,54 @@ function _finishReqResCycle(req, res, route, err) {
...
+
+
+
+/**
+ * Return debug information about the server.
+ * @public
+ * @method debugInfo
+ * @returns {Object}
+ */
+Server.prototype.debugInfo = function debugInfo() {
```
DonutEspresso: Was thinking about whether or not this also made more sense as `toJSON()`?

lib/router.js
DonutEspresso: Oh I see, you're using this in server's debug. Still think we could use a better name, though.

yunong: We could call this `getState()`

docs/pages/components/server.md
yunong: there was another round of discussion where we felt `inflight` was more concise. cc @ktrott @rajatkumar

lib/server.js
yunong: `toJSON()` isn't correct. This gives us the internal state of the object -- whereas `JSON` is merely the way in which we serialize that state -- which we could elect to change. I'm open to other names.

DonutEspresso: Oops sorry, should have removed this comment after I understood what was going on. This is what `toJSON()` would call. I think `debugInfo()` is fine.

lib/router.js
DonutEspresso: What about it just calling it `debugInfo()` like we do on server?

PR #1290, opened 2017-03-17
Fix description of createServer TLS options

bemasc:


DonutEspresso:
Thanks for the PR! It looks like markdown is interpreting the `|` as a column somehow - you might have to do `String or Buffer`, or if you know how GFM does escapes for the `|` character (I have no idea): 

https://github.com/bemasc/node-restify/blob/5.x/docs/pages/components/server.md

retrohacker:
Included in #1314

PR #1293, opened 2017-03-19
4.x fix tests

DonutEspresso:


DonutEspresso:
Oops, wrong target branch.

PR #1294, opened 2017-03-19
fix failing style rules and unit tests

DonutEspresso:


retrohacker:
This LGTM. Though when I checkout the [parent of this commit](https://github.com/restify/node-restify/commit/808f30a25166d96dfc098e85fab2aac351ce45d9) and run the `make test && make check`, I don't see any linting errors or failing tests. Am I missing something?

yunong:
You'll have to run `make prepush` which runs all of the prepush tasks
including lint and style.

retrohacker:
`make prepush` comes back green as well (looks like it may just be delegating work down to `make test` and `make check`).

I'm :+1: on going ahead and merging this since I agree with all of the changes. It doesn't really change the behaviour of anything outside of tests, other than a method returns `null` instead of `undefined`.

Its a little alarming though that eslint isn't picking up on the consistent return checks for me on the parent, might dive into that later.

Issue #1283, opened 2017-03-01
node-uuid is deprecated


piorek:
While doing npm install im getting deprecation message

> npm WARN deprecated node-uuid@1.4.7: use uuid module instead

fmauNeko:
Duplicate of #1233, already fixed in the 5.x branch.

piorek:
So you are not planning to do it in 4.x?

retrohacker:
This has been backported to both the 4.x and 5.x branches. Marking this as resolved for now :heart:

piorek:
@retrohacker Thanks!

PR #1295, opened 2017-03-19
Upgrade: update uuid

DonutEspresso:
This was fixed in 5.x, but since we haven't been able to get it out of beta yet, doing in on 4.x too.


retrohacker:
Fixes #1283

Issue #1296, opened 2017-03-23
Node-Restify 4.3 - Error when no uri is given when creating a http_client and no_proxy environment variable is present


Crazedpeanut:
Node-Restify 4.3 - Error when no uri is given when creating a http_client and no_proxy environment variable is present

TypeError: Cannot read property 'replace' of null
      at isProxyForURL (node_modules/restify/lib/clients/http_client.js:287:36)
      at JsonClient.HttpClient (node_modules/restify/lib/clients/http_client.js:376:24)
      at JsonClient.StringClient (node_modules/restify/lib/clients/string_client.js:29:16)
      at new JsonClient (node_modules/restify/lib/clients/json_client.js:23:18)
      at createClient (node_modules/restify/lib/index.js:46:22)
      at Object.createJsonClient (node_modules/restify/lib/index.js:79:13)
      at PromiseInvoker.invoke (src/promise-invoker.js:9:3786)
      at Context.it (test/promise-invoker.unit.test.js:256:30)

retrohacker:
I'm closing this as stale for now due to inactivity on #1297 

We can re-open and discuss if we get a repro case :smile:

PR #1297, opened 2017-03-24
Bugfix - https://github.com/restify/node-restify/issues/1296

Crazedpeanut:


retrohacker:
Hey @Crazedpeanut,

Thanks for this! It might be helpful if you give an example of

1. the _expected_ behaviour before this PR
2. the _actual_ behaviour before this PR

This could be a simple repro case we can run that demonstrates the bug.

Crazedpeanut:
Sure thing! I am a bit new to contributing to open source projects

retrohacker:
Hey @Crazedpeanut,

I'm closing this as stale for now. If you have time to create the requested examples we can re-open and do a proper review of this PR :smile:

retrohacker:
We are always excited to see first time contributors, especially those who are new to open source! If you need any assistance in getting a repro case together, let us know :heart:

Crazedpeanut:
Sorry for the super long response, here is an example of the error I am getting.

Let me know if you need more information!

[Reproduce-Restify-No-Proxy-Error.txt](https://github.com/restify/node-restify/files/1087036/Reproduce-Restify-No-Proxy-Error.txt)

retrohacker:
Cool beans! I did verify that this is a bug, and that this does in fact fix it! :tada:

The new code lives in this file here: https://github.com/restify/clients/blob/master/lib/HttpClient.js#L355

Would you mind opening a PR against that repo @Crazedpeanut? With this the only outstanding thing will be to provide a test that

* Passes when your fix is introduced
* Fails without your fix

Your repro case is pretty close to this, it just needs to use a restify server instead, which the tests in the clients repo automatically setup for you. There are some good examples of how to do this in the existing tests for the repo: https://github.com/restify/clients/blob/master/test/index.js#L267-L273

Closing this for now, will merge when a PR is opened against the clients repo with a test :smile:

Thank you for all of your hard work on this @Crazedpeanut :heart:

PR #1302, opened 2017-04-06
jsonp formatter unicode escapes line and paragraph separators

g-k:
http://timelessrepo.com/json-isnt-a-javascript-subset

see also:

https://golang.org/src/encoding/json/encode.go#L184
https://github.com/expressjs/express/blob/9722202df964bfbfc0f579e4baeb5a4e1b43b344/lib/response.js#L314-L317


yunong:
Thanks @g-k for your contribution.

PR #1313, opened 2017-04-26
Fix documentation for conditionalRequest

### LINKED ISSUE #473: conditionalRequest Plugin doesn't work
Hi,

I have been trying to get restify to send etag headers and check for if-\* headers to work nicely with my application. However when i try using the conditionalRequest plugin it does nothing.

Additionally the sample code on the homepage is also incorrect, since the first middleware never calls next, and thus the route is actually never served. Secondly the etag is hard coded, doesn't not look for etag from the route.

Can someone please provide me with a workable example, or perhaps help me understand how this is suppose to work.

retrohacker:
Fixes #473

### DIFF
--- a/index.html
+++ b/index.html
@@ -1322,10 +1322,12 @@ with a <code>String</code> key, and an <code>Object</code> value.</p>
 
 <p>Some example usage:</p>
 
-<pre><code>server.use(function setETag(req, res, next) {
-  res.header('ETag', 'myETag');
-  res.header('Last-Modified', new Date());
-});
+<pre><code>server.use(function (req, res, next) {
+  res.etag = "" + req.method + req.url
+  res.header('Etag', res.etag)
+  res.header('Last-Modified', Date.parse('12-31-2016'))
+  return next()
+})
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1334,6 +1336,8 @@ <h3 id="conditional-request-handler">Conditional Request Handler</h3>
 });
 </code></pre>
 
+The above example sets <code>res.etag</code> which is compared against the request's headers by <code>restify.conditionalRequest()</code> to determine if a <code>304</code> should be returned. It also sets the <code>Etag</code> header in the response, so that the client can include it in future requests. Finally, it sets the <code>Last Modified</code> header so <code>restify.conditionalRequest()</code> can determine if the client's copy of the request needs to be updated.
```
yunong: Nit, we generally stick to a column width of 80 columns, even in documentation. I know, it's archaic but it's how we roll. #80ColumnsLikeGodIntended.

retrohacker: Heh, I actually follow the same rule. I failed to run `gq}` before closing vim, embarrassing!

Issue #1305, opened 2017-04-13
Node v6 can throw startup warnings on OSX under certain configurations


hetul:
I am on OSX El Capitan 10.11.2, using restify 4.3.0 with Node v6.10.2. When running under this configuration, I get the following errors related to `dtrace-provider` on OSX for my node app:

```
(node) v8::ObjectTemplate::Set() with non-primitive values is deprecated
(node) and will stop working in the next major release.

Security context: 0x252991dcfb39 <JS Object>#0#
    1: .node [module.js:597] [pc=0x36ccf4ff3e44] (this=0x2f41a4d8b0c1 <an Object with map 0x23f1fd11b2d9>#1#,module=0x16230ded361 <a Module with map 0x23f1fd11c6c9>#2#,filename=0x16230debdd9 <String[101]: /Users/hetul/git/outlier-2/api/node_modules/dtrace-provider/build/Release/DTraceProviderBindings.node>)
    2: load [module.js:~478] [pc=0x36ccf51ae818] (this=0x16230ded361 <a Module with map 0x23f1fd11c6c9>#2#,filename=0x16230debdd9 <String[101]: /Users/hetul/git/outlier-2/api/node_modules/dtrace-provider/build/Release/DTraceProviderBindings.node>)
    3: tryModuleLoad(aka tryModuleLoad) [module.js:446] [pc=0x36ccf4f4145d] (this=0x252991d04381 <undefined>,module=0x16230ded361 <a Module with map 0x23f1fd11c6c9>#2#,filename=0x16230debdd9 <String[101]: /Users/hetul/git/outlier-2/api/node_modules/dtrace-provider/build/Release/DTraceProviderBindings.node>)
    4: _load [module.js:438] [pc=0x36ccf4f35422] (this=0x2f41a4d8b079 <JS Function Module (SharedFunctionInfo 0x2f41a4d399a9)>#3#,request=0x16230dea789 <String[38]: ./build/Release/DTraceProviderBindings>,parent=0x16230de8501 <a Module with map 0x23f1fd11c6c9>#4#,isMain=0x252991d04271 <false>)
    5: require [module.js:~494] [pc=0x36ccf4ff630c] (this=0x16230de8501 <a Module with map 0x23f1fd11c6c9>#4#,path=0x16230dea789 <String[38]: ./build/Release/DTraceProviderBindings>)
[...]

PR #1318, opened 2017-04-27
Update version of the dtrace-provider optionalDependency

hetul:
Fixes https://github.com/restify/node-restify/issues/1305

I ran the unit tests and they still pass. I did not actually test that dtrace still works as expected with the new version, so someone will need to verify that (or show me how that should be checked).

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "vasync": "^1.6.4"
   },
   "optionalDependencies": {
-    "dtrace-provider": "^0.6.0"
+    "dtrace-provider": "^0.8.1"
   },
   "devDependencies": {
     "cover": "^0.2.9",


retrohacker:
Thanks for the quick turn around on this! :heart:

I've never used dtrace either. I looked through the tests and it doesn't appear that any of the dtrace logic is tested, though I may be overlooking something.

I'll spend some time tonight figuring out how to test this :smile:

retrohacker:
I managed to get dtrace installed on debian, but it isn't picking up on restify's logs. I'm going to circle back to this one soon :tm:

yunong:
@retrohacker thanks for your efforts. Just a heads up that DTrace is unlikely to work on GNU/Linux, but does work on Mac OS X, Illumos, Solaris, and BSD.

If you have access to any of those systems you can try this section in the docs to see if it works:

http://restify.com/#dtrace

PR #1319, opened 2017-04-27
Close connection on error

### LINKED ISSUE #1308: uncaughtException to close connection
Just want to bring the idea of closing the tcp connection on an uncaught exception.

If an uncaught exception occurs, that current tcp connection is killed and could cause clients to hang when trying to connect in subsequent requests.

This might make the library too opinionated though since it's adding it's own http standards - happy to discuss and make a PR.

Current way to keep subsequent client connections from hanging when an uncaughtException is thrown:
```
server.on('uncaughtException', (req, res, route, err) => {
  res.setHeader('Connection', 'close');
  res.send(err);
});
```

retrohacker:
See #1308 for context

Fixes #1308

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -41,6 +41,11 @@ function createServer(options) {
                 return (false);
             }
 
+            // In the event an exception was thrown and not caught, the domain
+            // belonging to the active TCP connection will no longer be usable,
+            // the client should create a new TCP connection on the next
+            // request.
+            res.setHeader('Connection', 'close');
             res.send(new InternalError(e, e.message || 'unexpected error'));
             return (true);
         });
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -660,6 +660,10 @@ test('GH-77 uncaughtException (default behavior)', function (t) {
     CLIENT.get('/', function (err, _, res) {
         t.ok(err);
         t.equal(res.statusCode, 500);
+
+        if (res.headers) {
+            t.equal(res.headers.connection, 'close', 'close connection');
+        }
         t.end();
     });
 });
@@ -676,6 +680,12 @@ test('GH-77 uncaughtException (with custom handler)', function (t) {
     CLIENT.get('/', function (err, _, res) {
         t.ifError(err);
         t.equal(res.statusCode, 204);
+
+        // We don't close the connection for our user if they specify their own
+        // handler
+        if (res.headers) {
+            t.equal(res.headers.connection, 'keep-alive');
+        }
         t.end();
     });
 });


yunong:
Since domains support is deprecated at this point in 5.x -- my initial thoughts are leaning towards not merging this. Thoughts?

retrohacker:
Ah, I wasn't aware that this was being deprecated. I agree with that :smile: domains can be problematic, and some of our open issues reflect the false sense of security they provide.

Since this change is only on the execution pathway when users opt-in to domains, I think its a safe change to introduce. It fixes a rather hard-to-find bug for clients that attempt to re-use a TCP connection when a domain goes upside down.

DonutEspresso:
Deprecation doesn't mean it shouldn't work correctly, IMHO? Given that the work and tests are already done, I think we should merge this given it's pretty low effort and low cost to maintain.

yunong:
Is this the expected behaviour in HTTP? Since the protocol is stateless, the client shouldn't be relying on state passed back from the server to determine how to handle the next connection.

retrohacker:
We make it stateful by passing `Connection: keep-alive`, this instructs the client to re-use the TCP connection on the next request. When we crash, we want to explicitly tell the client to use a new TCP connection on it's next request, otherwise the packets (seem) to make it to the Node.js process but never get a response, and the client sits around waiting for the connection to timeout.

retrohacker:
Something that was keeping me up last night:

With the deprecation of domains, are we loosing the ability to reliably trace an `uncaughtException` back to the HTTP request that caused it? Without liberal use of `try`/`catch` (an anti-pattern IMO), can users determine which connection needs to be sent a 500? If not, draining active connections will need a timeout since that one open socket will never run out of "work" on the EventLoop and the process will never gracefully exit. Possibly worth opening an issue for?

yunong:
@retrohacker 
> We make it stateful by passing Connection: keep-alive, this instructs the client to re-use the TCP connection on the next request. When we crash, we want to explicitly tell the client to use a new TCP connection on it's next request, otherwise the packets (seem) to make it to the Node.js process but never get a response, and the client sits around waiting for the connection to timeout.
> 
I would argue that having the client timeout and retry is the expected behaviour. When an uncaught exception occurs on the server, there's no guarantee of correctness at that point; we can't guarantee we can even set headers and send responses back correctly. 

There's no expectation taht even if a server sets `Connection: keep-alive` that the server can't misbehave. In the scope of a distributed system, the onus is always on the client to be tolerant of failures.

yunong:
@retrohacker 
> With the deprecation of domains, are we loosing the ability to reliably trace an uncaughtException back to the HTTP request that caused it? Without liberal use of try/catch (an anti-pattern IMO), can users determine which connection needs to be sent a 500? If not, draining active connections will need a timeout since that one open socket will never run out of "work" on the EventLoop and the process will never gracefully exit. Possibly worth opening an issue for?

With the deprecation of Domains, we're setting the expectation that the default behaviour on an UncaughException is that the server aborts and dumps core. This usually means that the stack trace of that exception is dumped to the logs, and a core dump is captured to disk. 

With a core dump, we're able to get the full state of the process heap when the app crashed. We can't expect folks to go snooping around in core dumps on every crash. To your point, if we crashed before we finished the request, the `audit` log wouldn't be generated, and we would lose visibility into the request that caused the crash.

We should probably log an `incomingRequest` audit log on each request, which then will make it clear the request that caused the UncaughtException. However, there's still no guarantee that these logs will even make it to disk, since I think most logging systems buffer, and so on a process crash the last bit of logs will never got flushed to disk.

retrohacker:
> the onus is always on the client to be tolerant of failures.

Absolutely, but if we can avoid a timeout I would argue it is best to avoid the timeout. Servers should be as self correcting as possible, only requiring the client take defensive measures when absolutely necessary (and clients should assume that at some point it will always be necessary, which I think we are on the same page about). Attempting to send the headers may be unsuccessful, but it if is successful it will avoid a stalled connection. 30 second (or more) timeouts can lead to poor user experiences, and while the browser or load-blancer can detect that timeout and reroute/retry, having to avoid that timeout entirely seems ideal.

That being said, this change is super minor, spending much more time discussing it is probably investing more than the overall savings of introducing it. I'm staying :+1: on the change but we can close this in favor of #1320 that shows how to gracefully handle shutdown if possible, especially since that will be applicable outside of domains as well.

That being said, I'm not sure how useful a coredump will be if you gracefully cleanup before shutting down and only `throw` in `process.on('beforeExit'`

yunong:
@retrohacker I appreciate the discussion 👍 . One thing to be mindful of which is a subtle but important distinction is 

>  if you gracefully cleanup before shutting down 

This is generally not recommended if you wish to capture the process state via a core dump and debug post-mortem. If you're performing additional work before shutdown on an UncaughtException, it means at least a few things. First there's no way of knowing if any of the work you're doing is correct; second, performing additional work after the error means you're not mutating the process state, which makes it much harder to debug and get root cause on what actually caused the error.

retrohacker:
Ah, I agree. Gracefully shutting down sounded nice on the surface, but it ruins a core dump.

I'd like to see something along these lines in a "best practices" guide at some point, even if its just a link to the joyent documentation. Might try taking another shot at that soon.

PR #1320, opened 2017-04-27
Uncaught exception documentation

### LINKED ISSUE #1308: uncaughtException to close connection

retrohacker:
This PR introduces a "Best Practices" guide, it is written from the perspective of running restify in production.

Complements #1319 

Fixes #1308

### DIFF
--- a/docs/pages/guides/server-best-pracitces.md
+++ b/docs/pages/guides/server-best-pracitces.md
@@ -1,0 +1,48 @@
+# Server Best Practices
+
+This guide contains a collection of best practices to follow when using
+restify. We recommend that you only deviate from these practices with good
+reason and much thought. This advice is given from the perspective of running
+restify in production at scale.
+
+## Error Handling
+
+It is important to differentiate between two classes of errors:
+
+* Expected errors
+* Exceptions
+
+Expected errors are the code paths you identify upfront as possibly failing,
+and your server should contain logic that gracefully handles those errors. A
+great example of this is when your server makes an REST request to another
+server. Network calls are not always successful, and a failed network call
+doesn't necessarily mean there is a problem. This guide wont focus in one how
+to gracefully handle this class of error, since every case will be different.
+That being said, the rule of thumb should be to have your server self correct
+if possible, and only expose errors to the client when absolutely necessary.
+
+The second class of error, exceptions, are problems that arise due to logic
+errors. For problems that the process can not recover from, an error will be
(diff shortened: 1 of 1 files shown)


yunong:
@DonutEspresso is our resident renaissance man when it comes to documentation, tagging him for this review.

retrohacker:
Ah! This appears to have included  the content from #1319 😞

retrohacker:
Found out in #1319 that domains are being deprecated. A large chunk of this has become irrelevant, `server.on('uncaughtException'` needs to be changed to `process.on('uncaughtException'` and graceful cleanup needs to happen there. Will revisit tomorrow.

DonutEspresso:
This is a great summary of the Joyent docs! That said, I think I'd be inclined to just reference those docs given the pending deprecation, and simply have a one or two liner like "we recommend following the recommended best practices [link]. If you choose to use the uncaughtException feature, please be aware of the trade offs [here][here][here]". I'm not even sure we need to get into `process.on('uncaughtException'`, at least from the perspective of restify docs. Thoughts?

retrohacker:
Thanks for looking at this @DonutEspresso,

I was planning on revisiting this today :smile: I agree that a collection of links may contain enough information for a user to learn proper best practices for running a Node.js `net` or `tls` server in production, but I think collecting them into an opinionated guide would help restify establish it's ethos as a framework designed for running reliable REST services in production. I'm not sure if the time investment (I'm estimating ~45 minutes) would justify that though.

If we decide to push forward with this, I was planning on putting in a quick example showing how you can stop accepting new incoming connections in restify, and how to let your process shut itself down with a proper exit code (not having to explicitly call `process.exit`) once all existing connections have drained, so users know how to avoid having a restify service stuck between `uncaughtException` and a clean exit when the EventLoop never runs out of work to do. Also, possibly an example that immediately and cleanly closes all active connections and shuts down instead of letting the connections drain (to prevent clients from hanging).

Thoughts?

retrohacker:
Based on the conversation over in #1319, I think its safe to close this PR.

PR #1329, opened 2017-04-30
Set Copyright to it's own line

theel0ja:
Currently it looks like this, the Copyright line is in the same line with the *The MIT License (MIT)*:

image


retrohacker:
LGTM :+1:

Thank you for this!

PR #1330, opened 2017-05-01
Handle the case where content type is array

serkanyersen:
When JSONP is enabled, sometimes content type is received as an array

```
[ 'application/javascript', 'application/json' ]
```

Current system does not handle this. it falls in to the next `else if` where indexOf is used wrongly.

`} else if (type.indexOf(';') !== '-1') {` compares against a string. which always ends up being True.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -123,7 +123,7 @@ Response.prototype._findFormatter = function _findFormatter(callback) {
         }
     } else if (Array.isArray(type)) {
         type = type[0];
-    } else if (type.indexOf(';') !== '-1') {
+    } else if (type.indexOf(';') !== -1) {
         type = type.split(';')[0];
     }
 


serkanyersen:
is it possible to get this fix applied to both 4x and 5x?

serkanyersen:
to recreate this issue try this.
```javascript
api.use(restify.jsonp());
api.get('/api/test', (req, res, next) => {
    res.json({
        yes: true
    })
});
```

When you make the request with `?callback=test` it fails with
```
type = type.split(';')[0];
                    ^
TypeError: type.split is not a function
```

if you change the route with this
```javascript
api.get('/api/settings', (req, res, next) => {
    res.send(200, {
        yes: true
    })
});
```

It works correctly.

retrohacker:
Hey @serkanyersen,

Thank you for taking the time to find the bug, create a repro case, _and_ open a PR! We really appreciate your effort here :heart:

I'm tagging this for now, and will try to circle back soon to take a pass at this!

retrohacker:
Took a quick look for high level feedback for you. Generally we require tests to be implemented before we consider merging a PR, the tests should reveal the bug when your changes aren't included, and they should provide convincing and thorough assertions that the implemented behaviour is correct.

Totally understand holding off on implementing tests until you get feedback on the current work though :heart: just keep in mind that request is on the horizon

retrohacker:
Hey @serkanyersen,

I'm taking a look at this now and I'm struggling to recreate a case where `contentType` is an array. Can you create a simple repro case where the incoming request results in `contentType` being an array? This change looks simple enough, so I'm not opposed to merging with proper tests.

From our new issue template regarding repro cases:

> Please include a simple and concise example reproducing this bug. Please _do
> not_ just dump your application here. By either not providing a repro case or
> by providing an overly complicated repro case, you are offloading the work of
> isolating your bug to other developers, many of which are here voluntarily.
> Good repro cases are single file Node.js applications, where the only logic
> present is logic necessary to expose the undesired behaviour. You will often
> find that when creating your repro case, you will solve the problem yourself!

This repro case could (and probably should!) come in the form of a test for this PR so we can get it merged for you :smile:

In regards to v4, this can be back ported if needed. That being said, v5 should be shipping soon (only have a few things left to do) so it may be best to wait it out. Let me know what you think!

retrohacker:
Hey @serkanyersen, I'm closing this issue as stale for now due to inactivity. If this problem persists and we can get a repro case to help track it down on our end I would be happy to re-open and get this merged :heart:

PR #1351, opened 2017-05-15
Opinionated cleanup of argumentsToChain

retrohacker:
* Improve comments and JSDoc
* Eliminate need for edge case handling
* Single purpose function (don't conflate converting `arguments` to `Array` with flattening handlers)


retrohacker:
@mcavage I know its been a super long time since this code was written (<6 years) but what was the rationale for supporting a nested arrays of handlers for `_addRoute` and `use`?

retrohacker:
I would like to remove the recursive flattening here, but we can kick that can down the road. I think the comments clarifying this chunk of code are good for now. Thoughts on merging @rajatkumar @DonutEspresso?

lib/server.js
```diff
@@ -61,41 +61,40 @@ function _reqClosed(req) {
...
- * helper function to help verify and flatten an array of arrays.
- * takes an arguments object and an index frmo which to slice, then
- * merges that into a single array.
+ * Verify and flatten a nested array of request handlers.
+ *
  * @private
  * @function argumentsToChain
  * @throws   {TypeError}
- * @param    {Object} args  pass through of funcs from server.[method]
- * @param    {Number} start index of args at which to start working with
```
DonutEspresso: Was `start` ever used with a value of not `1`?

lib/server.js
```diff
@@ -651,12 +654,15 @@ Server.prototype.rm = function rm(route) {
  * You can pass in any combination of functions or array of functions.
  * @public
  * @function use
+ * @param {...*} handlers - A variable number of handler functions * and/or a
```
DonutEspresso: Whoa, is this a JSDoc thing? What does it mean?

lib/server.js
DonutEspresso: Oh, looks like it was only ever 1 or not used at all.

DonutEspresso:
LGTM

lib/server.js
retrohacker: Its been a bit since I wrote this, but I believe I'm using the wildcard implying any type is valid. I don't believe there is a way of specifying multiple types beyond this. I would like to be able to remove the nested array of handlers logic but haven't heard back from @mcavage as to why this was included or if it needs to stick around.

The `...` I believe implies a variadic function.

lib/server.js
retrohacker: Yeah, I'm not sure if it was used at some point or was included to support something they would do in the future, but it doesn't need to be there.

PR #1352, opened 2017-05-15
Create unit tests for sanitizePath

retrohacker:
* Create unit tests for sanitizePath


DonutEspresso:
Hmm is there overlap between this and the plugins sanitize path plugin?

retrohacker:
There is, whats more is that it doesn't look like the `util.sanitizePath` function is used anywhere. I'm :+1: on moving this implementation over to the plugin and migrating the tests to that API. The plugin's implementation is done with a for loop.

retrohacker:
@alex ready for another review :smile:

DonutEspresso:
LGTM! Rebase and let's get this in. 🎉

Issue #1354, opened 2017-05-15
emitRouteError doesn't wait for all handlers before closing request


retrohacker:
- [x] Used appropriate template for the issue type
- [x] Searched both open and closed issues for duplicates of this issue
- [x] Title adequately and _concisely_ reflects the feature or the bug

# Bug Report

## Restify Version

5.x

## Node.js Version

All

## Expected behaviour

When a route error occurs, all handlers should be allowed to complete before cleaning up the request.

## Actual behaviour

The current implementation uses `once` to protect against the cleanup logic running multiple times, but this means the cleanup will happen after the first handler.

## Cause

https://github.com/restify/node-restify/blob/5.x/lib/server.js#L212-L215

## Are you willing and able to fix this?

Yes :tada:

retrohacker:
A majority of the work has been done for this in #1355, just needs tests

PR #1355, opened 2017-05-15
Opinionated cleanup of emitRouteError

retrohacker:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #1354 

# Changes

* Adding comments
* Simplify edge conditions
* Use `listenerCount` as opposed to creating a copy of the listeners array
* Wait for all event handlers to run before executing cleanup

Alternatively, instead of relying on callbacks, we can assume all event handlers are synchronous and close the request as soon as execution reaches the next line. I believe this would be a significant breaking change. Pushing the requirement onto the user to call a callback on completion seems preferable.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -220,7 +220,7 @@ function emitRouteError(server, req, res, err) {
     function cb() {
         // Check to see if we have received as many callbacks as there are
         // listeners
-        if (invoked++ === server.listenerCount()) {
+        if (invoked++ === server.listenerCount(name)) {
             cleanup();
         }
     }
--- a/lib/server.js
+++ b/lib/server.js
@@ -208,14 +208,26 @@ function emitRouteError(server, req, res, err) {
 
     req.log.trace({name: name, err: err}, 'entering emitRouteError');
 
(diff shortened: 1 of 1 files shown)


retrohacker:
It doesn't appear that there is a test covering this behahviour. Do not merge :smile:

lib/server.js
```diff
@@ -208,15 +208,27 @@ function emitRouteError(server, req, res, err) {
...
+    function cleanup() {
         res.send(err);
         server._finishReqResCycle(req, res, null, err);
     }
+
+    var invoked = 0;
+    function cb() {
+        // Check to see if we have received as many callbacks as there are
+        // listeners
+        if (invoked++ === server.listenerCount(name)) {
```
DonutEspresso: I think this is a breaking change, i.e., the response won't get flushed up until all listener callbacks are invoked. I don't have opinions about this opinionated cleanup (badum psh) but probably worth documenting in the migration guide.

DonutEspresso:
With tests, LGTM.

retrohacker:
Closing until this irks me again

PR #1369, opened 2017-06-09
fix(audit): use public APIs for accessing response headers

montmanu:
## Pre-Submission Checklist

- [x] Opened an issue discussing these changes before opening the PR
- [x] Ran the linter and tests via `make prepush`
- [ ] Included comprehensive and convincing tests for changes

## Issues

Closes:

* Issue #1370

# Changes

- `res._headers` has been deprecated and accessing it in `auditResponseSerializer` generates exception (RangeError: Maximum call stack size exceeded)
- `res.getHeaders` returns an object that does not inherit from `Object`, so is unsuitable for use here
- falls back to `res._headers` if public APIs (introduced in v7.7.0) do not exist
- see https://nodejs.org/api/deprecations.html#deprecations_dep0066_outgoingmessage_headers_outgoingmessage_headernames
- see https://nodejs.org/api/http.html#http_response_getheaders

PR #1376, opened 2017-06-19
Node8 response headers

retrohacker:
Supersedes #1369

### DIFF
--- a/lib/plugins/audit.js
+++ b/lib/plugins/audit.js
@@ -7,6 +7,27 @@ var bunyan = require('bunyan');
 var HttpError = require('restify-errors').HttpError;
 var VError = require('verror');
 
+/**
+ * Utility to get response headers from a given response.
+ * Manually generates a POJO from `res.getHeaderNames` and `res.getHeader`,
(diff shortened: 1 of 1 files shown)


retrohacker:
Failing tests are unrelated to PR

vhenriet:
Could we hae this fix also for 4.x?

avimar:
The upgrade guide looks a little long, I wasn't planning on upgrading yet.
http://restify.com/docs/4to5/

Not that this is the best place to ask, but is there anything more to do to upgrade to 6.x?

retrohacker:
Hey @avimar :smile:

We use semantic versioning and conventional commits so it should be pretty easy to track what has changed between releases by looking at our commit history on master. https://github.com/restify/node-restify/commits/master

There is a lot to unpack there when moving from 4->5, which is why the migration guide exists, but 5-6 doesn't contain many changes to the API. Most changes were backwards compatible (unless you depended on internals or undocumented behaviour) and the ones that aren't are labeled `BREAKING CHANGE` in the commit history. Most of the breaking changes for 5->6 were trivial but we erred on the side of caution.

Hope this helps :heart:

Issue #1347, opened 2017-05-13
Remove Async Formatters


retrohacker:
# Feature Request

## Use Case

Currently the 5.x branch supports async formatters, however internally there is no way to support them and maintain the current restify API. After exploring possible solutions (i.e. #885), we have decided to drop support for async formatters prior to shipping 5x. The implications of this are that users will need to handle all asynchronous rendering outside of the restify stack, providing `res.send` et. al. values that are capable of being serialized synchronously.

## Example API

Identical to restify 4.x formatters

## Are you willing and able to implement this?

Yes :smile:

retrohacker:
Merged :tada:

PR #1377, opened 2017-06-21
Revert async formatters

retrohacker:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #1347 

Async formatters were introduced as a way to handle formatting an response body by binding it to data coming from an asyncronous data source. This introduced complexities that were difficult to reconsile with the rest of the code base. By requiring all formatters to be syncronous we simplify the codebase and can :clap: **ship 5.x** :clap: 

# Changes

> What does this PR do?

* Removes asynchronous formatters
* Subjectively simplifies `__send` implementation

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -410,6 +410,8 @@ Response.prototype.__send = function __send() {
         type = mime.lookup(type);
     }
 
+    // If we were unable to derive a valid type, default to treating it as
+    // arbitrary binary data per RFC 2046 Section 4.5.1
     if (!self.formatters[type] && self.acceptable.indexOf(type) === -1) {
         type = 'application/octet-stream';
     }
--- a/lib/response.js
+++ b/lib/response.js
@@ -255,7 +255,7 @@ Response.prototype.__send = function __send() {
     var code, body, headers, format;
 
     // derive arguments from types, one by one
-    var index = 0
+    var index = 0;
     // Check to see if the first argument is a status code
     if (typeof arguments[index] === 'number') {
         code = arguments[index++];
@@ -282,7 +282,7 @@ Response.prototype.__send = function __send() {
     // optional arguments was of an invalid type or we were provided with
     // too many arguments
     assert(arguments[index] === undefined,
-      `Unknown argument: ${arguments[index]}\nProvided: ${arguments}`)
+      'Unknown argument: ' + arguments[index] + '\nProvided: ' + arguments);
 
     // Now lets try to derive values for optional arguments that we were not
     // provided, otherwise we choose sane defaults.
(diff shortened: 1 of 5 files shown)


retrohacker:
Unit tests need to be updated to validate the new behaviour, but would like a review first in case there are any major changes :smile:

rajatkumar:
On the surface this looks good but I would like to see the unit tests as unit tests help me with understanding the code better.

lib/response.js
```diff
@@ -269,13 +209,11 @@ Response.prototype.link = function link(l, rel) {
  * @param    {Number} code http status code
  * @param    {Object | Buffer | Error} body the content to send
  * @param    {Object} headers  any add'l headers to set
- * @param    {Function} callback a callback for use with async formatters
- * @returns  {Object | undefined} when async formatter in use, returns null,
```
hetul: You may have already planned to do this, but consider updating the `response.md` documentation to be consistent with the new JSDoc, just so you don't forget to do it later.

retrohacker: Good catch! I hadn't looked at documentation, would have missed this :heart:

retrohacker:
Tests caught a bug where `response.send` and `response.sendRaw` didn't have optional parameters so the argument parser in `__send` was broken. Fixed using `Array.prototype.push`.

Removed all async formatter specific tests.

Only tests failing now are node 8 specific.

Ready for another review, assuming everyone is :+1: on this it's ready for a merge ^_^

hetul:
I suppose fixing node 8 support is on another ticket? In any case, LGTM! 👍

lib/response.js
```diff
@@ -284,114 +224,134 @@ Response.prototype.send = function send(code, body, headers, callback) {
...
 
+    // Set sane defaults for optional arguments if they were not provided and
+    // we failed to derive their values
+    code = code || 200;
     headers = headers || {};
 
+    // Populate our response object with the derived arguments
+    self.statusCode = code;
+    self._body = body;
+    Object.keys(headers).forEach(function (k) {
```
rajatkumar: Good to see the headers are being set here instead of `_flush()`

lib/response.js
```diff
@@ -404,68 +364,75 @@ function __send(maybeCode, maybeBody, maybeHeaders, maybeCallback, format) {
...
-            // async. invoke callback in that case. return null when no
-            // callback because eslint wants consistent-return.
-            return (callback) ? callback() : null;
-        }
-    });
+    if (!self.formatters[type] && type.indexOf('/') === -1) {
+        type = mime.lookup(type);
+    }
+
+    if (!self.formatters[type] && self.acceptable.indexOf(type) === -1) {
```
rajatkumar: Can we add a code comment explaining why we set type to `application/octet-stream`

retrohacker:
Only tests failing on CI are v8 related. Squashing and merging :tada:

Issue #1370, opened 2017-06-09
Use public APIs for accessing response headers


montmanu:
- [x] Used appropriate template for the issue type
- [x] Searched both open and closed issues for duplicates of this issue
- [x] Title adequately and _concisely_ reflects the feature or the bug

# Bug Report
`auditResponseSerializer` generates exception (`RangeError: Maximum call stack size exceeded`) when attempting to access deprecated `res._headers` property.

## Restify Version
5.0.0-beta-8.0

## Node.js Version
v8.0.0

## Expected behaviour
Log the request without generating an exception

## Actual behaviour
generates `RangeError`

<details>
<summary>bunyan: ERROR: Exception thrown from the "res" Bunyan serializer. This should never happen. This is a bug in that serializer function.</summary>
<pre><code>
RangeError: Maximum call stack size exceeded
   at ServerResponse.getHeaders (/root/app/node_modules/restify/lib/response.js:176:52)
   at ServerResponse.get (_http_outgoing.js:130:17)
   at ServerResponse.getHeaders (/root/app/node_modules/restify/lib/response.js:177:17)
   at ServerResponse.get (_http_outgoing.js:130:17)
   at ServerResponse.getHeaders (/root/app/node_modules/restify/lib/response.js:177:17)
   at ServerResponse.get (_http_outgoing.js:130:17)
   at ServerResponse.getHeaders (/root/app/node_modules/restify/lib/response.js:177:17)
   at ServerResponse.get (_http_outgoing.js:130:17)
   at ServerResponse.getHeaders (/root/app/node_modules/restify/lib/response.js:177:17)
   at ServerResponse.get (_http_outgoing.js:130:17)
</code></pre>
</details>

## Repro case
```javascript
// using node v8.0.0
var restify = require('restify');
var bunyan = require('bunyan');
var server = restify.createServer();

server.on('after', restify.plugins.auditLogger({
  log: bunyan.createLogger({
    name: 'audit',
    stream: process.stdout,
  }),
}));

server.get('/', function (req, res, next) {
  res.send(200, 'OK');
  return next();
});

server.listen(8080);
// GET http://127.0.0.1:8080/
```

## Cause

[Accessing](https://github.com/restify/node-restify/blob/5.x/lib/plugins/audit.js#L102) deprecated `res._headers` property.

## Are you willing and able to fix this?

https://github.com/restify/node-restify/pull/1369

PR #1378, opened 2017-06-22
Backport "Node8 response headers" to 4.x

vhenriet:
- `res._headers` has been deprecated and accessing it in `auditResponseSerializer` generates exception (RangeError: Maximum call stack size exceeded)
- `res.getHeaders` returns an object that does not inherit from `Object`, so is unsuitable for use here
- falls back to `res._headers` if public APIs (introduced in v7.7.0) do not exist
- see https://nodejs.org/api/deprecations.html#deprecations_dep0066_outgoingmessage_headers_outgoingmessage_headernames
- see https://nodejs.org/api/http.html#http_response_getheaders

## Pre-Submission Checklist

- [ ] Opened an issue discussing these changes before opening the PR
- [ ] Ran the linter and tests via `make prepush`
- [ ] Included comprehensive and convincing tests for changes

## Issues

Closes:

* Issue #
* Issue #
* Issue #

> Summarize the issues that discussed these changes

# Changes

> What does this PR do?

PR #1379, opened 2017-06-22
fix(audit): use public APIs for accessing response headers

gfpacheco:
Cloned from #1369 

## Pre-Submission Checklist

## Issues

Closes:

* Issue #1370 

# Changes

- res._headers has been deprecated and accessing it in auditResponseSerializer generates exception (RangeError: Maximum call stack size exceeded)
- res.getHeaders returns an object that does not inherit from Object, so is unsuitable for use here
- falls back to res._headers if public APIs (introduced in v7.7.0) do not exist
see https://nodejs.org/api/deprecations.html#deprecations_dep0066_outgoingmessage_headers_outgoingmessage_headernames
- see https://nodejs.org/api/http.html#http_response_getheaders

### DIFF
--- a/lib/plugins/audit.js
+++ b/lib/plugins/audit.js
@@ -7,6 +7,29 @@ var bunyan = require('bunyan');
 
 var HttpError = require('../errors').HttpError;
 
+/**
+ * Utility to get response headers from a given response.
+ * Manually generates a POJO from `res.getHeaderNames` and `res.getHeader`,
+ * if available, falling back to deprecated `res._headers`, otherwise.
+ * Intentionally does not use `res.getHeaders` to avoid deserialization
+ * issues with object returned by that method.
+ * @param {http.ServerResponse} res the OutgoingMessage
+ * @private
+ * @function getResponseHeaders
+ * @returns {object} map from header name to header value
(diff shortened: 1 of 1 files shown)


retrohacker:
This LGTM :+1:

I'm okay merging this w/o tests since the 5.x release is just around the corner and the behaviour is validated there. Thoughts @yunong?

retrohacker:
CC: @DonutEspresso @rajatkumar

retrohacker:
Also, I think this is a duplicate of #1378

gfpacheco:
I wouldn't mind back porting the tests, but the current tests are already throwing errors if ran on Node 8. The problem is that it's just [printing the errors to the stderr and passing the test anyway](https://gist.github.com/gfpacheco/6478fe1f29303567f43421f2fb481275#file-output-from-running-node-restify-branch-4-x-tests-on-node-8-L119).

I tried but couldn't find a way to make the tests fail. Could you give me a hint?

retrohacker:
@gfpacheco ah! I was misremembering that PR. Yes, the test is already failing, this PR should get it to pass. Let me clone and check.

Issue #1380, opened 2017-06-23
Versioned route matching throws TypeError when "maxV" is not set


kusor:
# Bug Report

When we have two routes applicable to the same request, one typed and the other versioned and not typed, a TypeError exception is thrown.

## Restify Version

Every one since `v2.8.5`, exactly since commit [Route matching should only prefer later routes if version is greater](https://github.com/restify/node-restify/commit/0a45bd380a45cd7fd62d62d3285e7a707e04bc5d)

## Node.js Version

Any.

## Expected behaviour

Exceptions should not be thrown

## Actual behaviour

```
TypeError: Invalid Version: undefined
    at new SemVer (/Users/pedropc/work/node_modules/restify/node_modules/semver/semver.js:281:11)
    at SemVer.compare (/Users/pedropc/work/node_modules/restify/node_modules/semver/semver.js:348:13)
    at compare (/Users/pedropc/work/node_modules/restify/node_modules/semver/semver.js:571:31)
    at Function.gt (/Users/pedropc/work/node_modules/restify/node_modules/semver/semver.js:600:10)
    at /Users/pedropc/work/node_modules/restify/lib/router.js:497:38
    at Array.forEach (native)
    at Router.find (/Users/pedropc/work/node_modules/restify/lib/router.js:493:24)
    at Server._route (/Users/pedropc/work/node_modules/restify/lib/server.js:753:21)
    at routeAndRun (/Users/pedropc/work/node_modules/restify/lib/server.js:711:14)
    at Server._handle (/Users/pedropc/work/node_modules/restify/lib/server.js:731:9)
```

## Repro case

```
var restify = require('restify');

var server = restify.createServer({
    name: 'myapp',
    version: '1.0.0'
});
server.use(restify.plugins.acceptParser(server.acceptable));
server.use(restify.plugins.queryParser());
server.use(restify.plugins.bodyParser());


server.post({
    path: '/:param_one/path/[0-f]+/:param_two/word',
    contentType: 'application/json'
}, function (req, res, next) {
    req.log.warn(req.headers);
    res.setHeader('Content-Length', '0');
    res.send(204);
    next();
});


server.post({
    path: '/:param_one/path/[0-f]+/:param_two/:param_three'
}, function (req, res, next) {
    req.log.warn(req.headers);
    res.setHeader('Content-Length', '0');
    res.send(204);
    next();
});

server.listen(8080, '127.0.0.1', function () {
    console.log('%s listening at %s', server.name, server.url);
});

[...]

PR #1381, opened 2017-06-23
GH-1380 versioned route matching should not throw TypeError

kusor:
## Issues

Closes:

* Issue ##1380

# Changes

Avoid exceptions being thrown when we have two routes applicable to the same request, one typed and the other versioned and not typed.


retrohacker:
Merging at EOD :smile:

PR #1383, opened 2017-06-26
fix(static): avoid user-provided data in Error messages being interpreted as sprintf codes

### LINKED ISSUE #1382: static plugin ResourceNotFoundError throws on sprintf-like percent code in req.path
Restify version: 4.x (I'm running 4.3.0).
Node.js version: v4.8.0.

## Expected behaviour

If using the restify server is using the static plugin and the client requests a path that doesn't exist, you'd expect a 404:

```
% curl -i http://10.99.99.21/docs/no-such-file.html
HTTP/1.1 404 Not Found
Content-Type: application/json
Content-Length: 72
Date: Mon, 26 Jun 2017 22:35:38 GMT
Connection: keep-alive

{
  "code": "ResourceNotFound",
  "message": "/docs/no-such-file.html"
}
```

## Actual behaviour

However, if the given request path includes percent codes (for URL escaping), that *also* look like printf codes, then you'll get an InternalError:

```
 { [InternalError: too few args to sprintf]
  jse_shortmsg: '',
  jse_info: {},
  message: 'too few args to sprintf',
  statusCode: 500,
  body: { code: 'InternalError', message: 'too few args to sprintf' },
  restCode: 'InternalError',
  name: 'InternalError' }
```

(or a crash if your restify server is configured to not use an 'uncaughtException' handler).
This is because:

- The 'static' plugin calls: `next(new ResourceNotFoundError(err, req.path()));`
- ResourceNotFoundError is a RestError that is based on verror.WError, which ultimately calls `extsprintf.sprintf(<path>)`.
- This blows up because:

```
> new restify.ResourceNotFoundError('/docs/not-%22such-file.html not found')
Error: too few args to sprintf
    ...
```

## Repro case

I'll have a test case in the coming PR.

trentm:
The 'static' plugin had a few cases where the path in a request would be
passed as the first ("message") field to a RestError constructor.
RestError uses verror.WError, which uses extsprintf to render the given
arguments. If the "message" includes "%...s" or similar printf codes,
then it will error output.

Also bump to 4.3.1.
Fixes #1382.


trentm:
gah, I meant for this to be vs 4.x branch. I'll re-create a separate PR.

PR #1385, opened 2017-06-27
Fixes the broken test for node version 8

retrohacker:
@rajatkumar's fix for failing tests in v8


retrohacker:
Holy crapola! Green CI! :man_dancing:

rajatkumar:
Yep, moving to error codes is good idea, but it is only in version 8 do we have specific `Nodejs Error Codes`.

retrohacker:
Small fix and is blocking Node.js version 8 support.

Merging at EOD.

retrohacker:
This was fixed in another PR

PR #1391, opened 2017-06-28
test: add 'static' plugin tests of 404s for #1382

trentm:


retrohacker:
This LGTM :+1:

Not going to cut a release off of this since it's only tests.

trentm:
> Not going to cut a release off of this since it's only tests.

Totally. Thanks.

PR #1405, opened 2017-07-08
Add node 7-8 travis support

simonepri:
## Pre-Submission Checklist

# Changes
Restify is fully compatible with both node 7 and 8 and they are quite used in production also if they aren't LTS at the moment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ language: node_js
 node_js:
    - "4"
    - "6"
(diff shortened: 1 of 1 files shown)


simonepri:
Some steps fails on Travis, but since you already support the `stable` version (That at the moment installs node 8) I think they are problem unrelated to this PR.

retrohacker:
Thanks @simonepri

Issue #1407, opened 2017-07-11
Restify 5.0 Response.js - Maximum call stack size exceeded - Node 8.1.4


ianvonholt:
# Bug Report

## Restify Version
`restify@5.0.0`

## Node.js Version
8.1.4

## Expected behaviour

Calling `res.getHeaders()` would return the appropriate header object.

## Actual behaviour

```
./node_modules/restify/lib/response.js:119
Response.prototype.getHeaders = function getHeaders() {
                                                   ^

RangeError: Maximum call stack size exceeded
    at ServerResponse.getHeaders (./node_modules/restify/lib/response.js:119:52)
    at ServerResponse.get (_http_outgoing.js:130:17)
    at ServerResponse.getHeaders (./node_modules/restify/lib/response.js:120:17)
    at ServerResponse.get (_http_outgoing.js:130:17)
    at ServerResponse.getHeaders (./node_modules/restify/lib/response.js:120:17)
    at ServerResponse.get (_http_outgoing.js:130:17)
    at ServerResponse.getHeaders (./node_modules/restify/lib/response.js:120:17)
    at ServerResponse.get (_http_outgoing.js:130:17)
    at ServerResponse.getHeaders (./node_modules/restify/lib/response.js:120:17)
    at ServerResponse.get (_http_outgoing.js:130:17)
```
## Repro case

```
const restify = require('restify');
const server = restify.createServer();
server.get('/', function (req, res, next) {
  res.send(200, res.getHeaders());
  next();
});
server.listen(8080, function () {});
```

## Cause

The use of `_headers` is deprecated and it appears that it is currently a wrapper for `getHeaders()` Public API in Node 8.1.4.

https://nodejs.org/api/deprecations.html#deprecations_dep0066_outgoingmessage_headers_outgoingmessage_headernames

https://github.com/nodejs/node/blob/v8.1.4/lib/_http_outgoing.js#L128-L144

The problem is the inheritance that Restify 5.0 displays with `http.ServerResponse`. At https://github.com/restify/node-restify/blob/5.x/lib/response.js#L22 `Response` is set as the appropriate object. 

https://github.com/restify/node-restify/blob/5.x/lib/response.js#L119-L121 Then goes to override the Public API function and causes an infinite loop. 

## Are you willing and able to fix this?

Simply commenting out https://github.com/restify/node-restify/blob/5.x/lib/response.js#L119-L121 would fix the issue in Node 8 environments as the method already exists.

[...]

PR #1408, opened 2017-07-11
Fix for #1407

ianvonholt:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #1407 

# Changes

In Node 7.7+ we use the `http.ServerResponse.getHeaders()` instead of creating a new prototype function.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -116,9 +116,11 @@ Response.prototype.get = function get(name) {
  * @function getHeaders
  * @returns  {Object}
  */
-Response.prototype.getHeaders = function getHeaders() {
-    return (this._headers || {});
-};
+if (typeof Response.prototype.getHeaders !== 'function') {
+    Response.prototype.getHeaders = function getHeaders() {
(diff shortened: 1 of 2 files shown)


kokarn:
This would be much appreciated, fixes an issue with apicache as well.

lib/response.js
```diff
@@ -116,9 +116,11 @@ Response.prototype.get = function get(name) {
  * @function getHeaders
  * @returns  {Object}
  */
-Response.prototype.getHeaders = function getHeaders() {
-    return (this._headers || {});
-};
+if (typeof Response.prototype.getHeaders !== 'function') {
```
yunong: Could you please move the JS Doc down into the if block on top of the function? Additionally If you could then add a comment before the if block about why this is necessary, that'd be great. Once we get the comments fixed we should be ready to land this. Thanks so much!

ianvonholt: @retrohacker beat me to it. Thank you for the quick response.

PR #1420, opened 2017-07-27
Fix: emit restifyError event even for router errors

### LINKED ISSUE #1419: `restifyError` event does not fire for router errors

# Bug Report
Router errors that are proxied and emitted through the server, i.e., NotFound/MethodNotAllowed/VersionNotAllowed are not emitted via the generic restifyErrors event. This was first reported by @mauron85 in #944.

## Restify Version
5.0.1

## Node.js Version
All versions. 

## Expected behaviour
On a router errors like NotFound (404), `restifyError` event should be emitted on the server object.

## Actual behaviour
`restifyError` event is not emitted 

## Repro case
```
'use strict';

const server = restify.createServer({ name: 'foo' });

    res.send(200);
});

server.on('restifyError', function(req, res, err, cb) {
    console.warn('restifyError should fire', err);
    return cb();
});

server.listen(3000);
```
Then curl localhost:3000/dne

## Cause
It's complicated.

Yes

DonutEspresso:
Fixes #1419 

This also removes the old FormatterError used in async formatters, which is no longer applicable.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -186,43 +186,9 @@ function ifError(n) {
 
 
 /**
- * when an error occurrs, this is used to emit an error to consumers
- * via EventEmitter.
- * @private
- * @function emitRouteError
- * @param    {Object} server the server object
- * @param    {Object} req    the request object
- * @param    {Object} res    the response object
- * @param    {Object} err    an error object
- * @returns  {undefined}
- */
-function emitRouteError(server, req, res, err) {
-
-    var name;
-
-    if (err.name === 'ResourceNotFoundError') {
-        name = 'NotFound';
-    } else if (err.name === 'InvalidVersionError') {
-        name = 'VersionNotAllowed';
-    } else {
-        name = err.name.replace(/Error$/, '');
-    }
-
-    req.log.trace({name: name, err: err}, 'entering emitRouteError');
-
-    if (server.listeners(name).length > 0) {
-        server.emit(name, req, res, err, once(function () {
-            res.send(err);
-            server._finishReqResCycle(req, res, null, err);
-        }));
-    } else {
-        res.send(err);
-        server._finishReqResCycle(req, res, null, err);
-    }
-}
-
-
-/**
+ * TODO: this is relevant for CORS only. Should move this out eventually to a
+ * userland middleware?
+ *
  * returns true if an error generated is for an options request.
  * @private
  * @function optionsError
@@ -232,15 +198,31 @@ function emitRouteError(server, req, res, err) {
  * @returns  {Boolean}
  */
 function optionsError(err, req, res) {
-    var code = err.statusCode;
-    var ok = false;
-
-    if (code === 404 && req.method === 'OPTIONS' && req.url === '*') {
-        res.send(200);
-        ok = true;
+    if (err.statusCode === 404 && req.method === 'OPTIONS' && req.url === '*') {
+        return true;
     }
+    return false;
+}
 
-    return (ok);
+
+/**
+ * map an Error's .name property into the actual event name that is emitted
+ * by the restify server object.
+ * @function
+ * @private getErrorName
+ * @param {Object} err an error object
+ * @returns {String} an event name to emit
+ */
+function getErrorName(err) {
+    if (err.name === 'ResourceNotFoundError') {
+        // remap the name for router errors
+        return 'NotFound';
+    } else if (err.name === 'InvalidVersionError') {
+        // remap the name for router errors
+        return 'VersionNotAllowed';
+    } else {
+        return err.name.replace(/Error$/, '');
+    }
 }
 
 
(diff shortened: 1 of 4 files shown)


lib/server.js
```diff
@@ -789,33 +772,46 @@ Server.prototype._handle = function _handle(req, res) {
...
         cb = name;
         name = null;
 
-        this.router.find(req, res, function onRoute(err, route, ctx) {
+        return this.router.find(req, res, function onRoute(err, route, ctx) {
             var r = route ? route.name : null;
 
             if (err) {
+                // TODO: if its a 404 for OPTION method (likely a CORS
+                // preflight), return OK. This should move into userland.
```
retrohacker: Why would we return 200 on an invalid preflight route?

lib/server.js
```diff
@@ -889,73 +885,18 @@ Server.prototype._run = function _run(req, res, route, chain, cb) {
 
         if (arg) {
             if (arg instanceof Error) {
```
retrohacker: Not part of this PR, but doesnt `arg instanceof Error === true` imply `arg`? This second check seems redundant and adds a level of indentation. Same with `typeof`

lib/server.js
```diff
@@ -889,73 +885,18 @@ Server.prototype._run = function _run(req, res, route, chain, cb) {
...
-                    });
-                    emittedError = true;
-                } else {
+                // the emitting of the error events are async, so we can not
+                // complete this invocation of run() until it returns. setting
+                // the emittedError flag is flag that is returned sync -
+                // setting this flag will allow us to to not call cb()
+                // automatically later in this function.
+                emittedError = self._emitErrorEvents(req, res, route, arg,
+                function emitErrorsDone() {
```
retrohacker: Did we lint this? It looks like this should have an extra level of indentation.

lib/server.js
```diff
@@ -1125,6 +1066,65 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
 };
 
 
+/**
+ * emit error events when errors are encountered either while attempting to
+ * route the request (via router) or while executing the handler chain.
+ * @private
+ * @function emitErrorEvents
```
retrohacker: This is missing the `_` prefix

lib/server.js
```diff
@@ -1125,6 +1066,65 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
...
+        }
+    }, function (nullErr, results) { // eslint-disable-line handle-callback-err
+        // vasync will never return error here. callback with the original
+        // error to pass it on.
+        return cb(err);
+    });
+
+    // while async listeners are firing, return a boolean indicating whether or
+    // not we had any listeners to fire.
+    return (errEvtNames.length > 0) ? true : false;
```
retrohacker: This block of code is super clever! I dig it! :clap: poetry :clap:

lib/server.js
```diff
@@ -186,43 +186,9 @@ function ifError(n) {
...
-        }));
-    } else {
-        res.send(err);
-        server._finishReqResCycle(req, res, null, err);
-    }
-}
-
-
-/**
+ * TODO: this is relevant for CORS only. Should move this out eventually to a
```
DonutEspresso: 👍  on this idea! I think it might be useful to have a new label then, something like.. `internal` for these non user facing clean up tasks/refactoring.

lib/server.js
DonutEspresso: I updated the TODO in the `optionsError()` method to reflect what's going on here. This behavior isn't actually new - just moved it out of the `optionsError()` method. Here's the comment I added:

```
/** 
 * returns true if the router generated a 404 for an options request.
 * TODO: this is relevant for CORS only. Should move this out eventually to a
 * userland middleware? This also seems a little like overreach, as there is no
 * option to opt out of this behavior today.
```

lib/server.js
DonutEspresso: `arg` can also be a boolean, or more specifically, a value of `false` which tells restify to stop executing the handler chain.

lib/server.js
DonutEspresso: Yep! Lint doesn't complain because function invocations don't have to be indented even if they are > 1 line.

I actually prefer this style, as it gives us an extra 4 spaces for this last arg. Open to suggestions on how to better do this and still keep within 80cols though.

lib/server.js
DonutEspresso: 🎉

DonutEspresso:
Did a little clean up of the _run function (scope creep!) but also added a ton of comments to better breakdown what's going on.

lib/server.js
retrohacker: cc @rprieto

lib/server.js
```diff
@@ -789,51 +774,78 @@ Server.prototype._handle = function _handle(req, res) {
  */
 Server.prototype._route = function _route(req, res, name, cb) {
     var self = this;
+    // helper function to, when on router error, emit error events and then
+    // flush the err.
+    var errResponse = function errResponse(err) {
+        return self._emitErrorEvents(req, res, null, err, function () {
+            res.send(err);
```
edvinerikson: all `res.send()` requires to check `res.headersSent` otherwise node will throw `Can't set headers after they are sent.`

Issue #1344, opened 2017-05-10
Place throttle limits in response header


retrohacker:
Migrated from https://github.com/restify/plugins/issues/80

eps1lon:
I did some work implementing this (which was pretty straight forward).

On which branch do you accept pull requests?

retrohacker:
Hey @eps1lon,

Thanks for doing this :heart:! PRs should be opened against the latest major release, and backported if necessary. Currently, that is 5.x :smile:

retrohacker:
Merged in https://github.com/restify/node-restify/pull/1453

PR #1425, opened 2017-08-01
added optional limit, rate and remaining headers when using throttle

eps1lon:
As request in #1344 this exposes headers that specify how the api is throttled. This will not happen unless explicitly set in `setHeaders`. 

Added headers are: 
* `X-RateLimit-Remaining` for remaining tokens
* `X-RateLimit-Limit` for burst 
* `X-RateLimit-Rate` for the refresh reate

I did some quick research and i only found that Twitter uses `x-rate-limit-*` instead of camelCase. 

Most changes were made in the test since i had to setup a new server for that test. As far es I can tell there is no way to change plugins options after they were added.

### DIFF
--- a/lib/plugins/throttle.js
+++ b/lib/plugins/throttle.js
@@ -204,6 +204,8 @@ TokenTable.prototype.get = function get(key) {
  *                   - {Boolean} ip (optional).
  *                   - {Boolean} username (optional).
  *                   - {Boolean} xff (optional).
+ *                   - {Boolean} setHeaders: Set response headers for rate,
+ *                               limit (burst) and remaining. Default is false.
  *                   - {Object} overrides (optional).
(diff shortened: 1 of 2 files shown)


lib/plugins/throttle.js
```diff
@@ -278,7 +280,16 @@ function throttle(options) {
         req.log.trace('Throttle(%s): num_tokens= %d',
             attr, bucket.tokens);
 
-        if (!bucket.consume(1)) {
+        var tooManyRequests = !bucket.consume(1);
+
+        // set throttle headers after consume which changes the remaining tokens
+        if (options.setHeaders) {
```
retrohacker: At the top of this function, we should assert `options.setHeaders` is an optional boolean.

PR #1431, opened 2017-08-03
Create inflightRequestThrottle plugin

retrohacker:
Plugin for limiting inflight requests

### DIFF
--- a/docs/api/plugins.md
+++ b/docs/api/plugins.md
@@ -156,8 +156,7 @@ event, e.g., `server.on('after', plugins.metrics());`:
 The module includes the following plugins to be used with restify's `pre` event:
 * `inflightRequestThrottle(options)` - limits the max number of inflight requests
   * `options.limit` {Number} the maximum number of inflight requests the server will handle before returning an error
-  * `options.res` {Error} An error that will be passed to `res.send` when the limit is reached.
-  * `options.res.statusCode` {Number} The status code to return when the limit is reached.
+  * `options.err` {Error} opts.err A restify error used as a response when the inflight request limit is exceeded
   * `options.server` {Object} The restify server that this module will throttle
 
 ## Accept Parser
@@ -463,7 +462,7 @@ the maximum number of inflight requests your server is able to handle. This
 is a simple heuristic for protecting against event loop contention between
 requests causing unacceptable latencies.
 
-The custom response is optional, and allows you to specify your own response
+The custom error is optional, and allows you to specify your own response
 and status code when rejecting incoming requests due to too many inflight
 requests. It defaults to `503 ServiceUnavailableError`.
 
--- a/docs/api/plugins.md
+++ b/docs/api/plugins.md
@@ -155,10 +155,10 @@ event, e.g., `server.on('after', plugins.metrics());`:
 
 The module includes the following plugins to be used with restify's `pre` event:
 * `inflightRequestThrottle(options)` - limits the max number of inflight requests
-  * `options.limit` {Number} the maximum number of simultaneous connections the server will handle before returning an error
-  * `options.resp` {Error} An error that will be passed to `res.send` when the limit is reached.
-  * `options.resp.statusCode` {Number} The status code to return when the limit is reached.
-  * `options.server.inflightReuqests` {Function} Should return the number of active connections to the server
+  * `options.limit` {Number} the maximum number of inflight requests the server will handle before returning an error
+  * `options.res` {Error} An error that will be passed to `res.send` when the limit is reached.
+  * `options.res.statusCode` {Number} The status code to return when the limit is reached.
+  * `options.server` {Object} The restify server that this module will throttle
 
 ## Accept Parser
 
@@ -449,8 +449,12 @@ with a `String` key, and an `Object` value.
 ## Inflight Request Throttling
 
 ```js
+var errors = require('restify-errors');
+var restify = require('restify');
+
+var server = restify.createServer();
 const options = { limit: 600, server: server };
-options.resp = new require('restify-errors').InternalServerError();
+options.res = new errors.InternalServerError();
 server.pre(restify.plugins.inflightRequestThrottle(options));
 ```
 
--- a/docs/api/plugins.md
+++ b/docs/api/plugins.md
@@ -156,9 +156,9 @@ event, e.g., `server.on('after', plugins.metrics());`:
 The module includes the following plugins to be used with restify's `pre` event:
 * `inflightRequestThrottle(options)` - limits the max number of inflight requests
   * `options.limit` {Number} the maximum number of simultaneous connections the server will handle before returning an error
-  * `opts.resp` {Error} An error that will be passed to `res.send` when the limit is reached.
-  * `opts.resp.statusCode` {Number} The status code to return when the limit is reached.
-  * `opts.server.inflightReuqests` {Function} Should return the number of active connections to the server
+  * `options.resp` {Error} An error that will be passed to `res.send` when the limit is reached.
+  * `options.resp.statusCode` {Number} The status code to return when the limit is reached.
+  * `options.server.inflightReuqests` {Function} Should return the number of active connections to the server
 
 ## Accept Parser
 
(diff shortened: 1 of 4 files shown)


docs/api/plugins.md
```diff
@@ -153,6 +153,13 @@ event, e.g., `server.on('after', plugins.metrics());`:
   * `res` {Object} the response obj
   * `route` {Object} the route obj that serviced the request
 
+The module includes the following plugins to be used with restify's `pre` event:
+* `inflightRequestThrottle(options)` - limits the max number of inflight requests
+  * `options.limit` {Number} the maximum number of simultaneous connections the server will handle before returning an error
```
DonutEspresso: `options` vs `opts`?

docs/api/plugins.md
```diff
@@ -439,6 +446,23 @@ uniform request distribution.  To enable this, you can pass in
 `options.tokensTable`, which is simply any Object that supports `put` and `get`
 with a `String` key, and an `Object` value.
 
+## Inflight Request Throttling
+
+```js
+const opts = { limit: 600, server: server };
+opts.error = new Error('Resource Exhausted');
+error.statuscode = 420;
```
DonutEspresso: Should that be camelCase `statusCode`? Also, I might suggest adding an example using restify-errors constructors.

docs/api/plugins.md
```diff
@@ -439,6 +446,23 @@ uniform request distribution.  To enable this, you can pass in
 `options.tokensTable`, which is simply any Object that supports `put` and `get`
 with a `String` key, and an `Object` value.
 
+## Inflight Request Throttling
+
+```js
+const opts = { limit: 600, server: server };
+opts.error = new Error('Resource Exhausted');
```
DonutEspresso: Is it `opts.error` or `opts.resp`?

lib/plugins/inflightRequestThrottle.js
```diff
@@ -0,0 +1,67 @@
...
+function onRequest (req, res, next) {
+    var self = this;
+    var inflightRequests = self._server.inflightRequests();
+
+    if (inflightRequests > self._limit) {
+        req.log.trace({
+            plugin: 'inflightRequestThrottle',
+            inflightRequests: inflightRequests,
+            limit: self._limit,
+            message: 'maximum inflight requests exceeded, rejecting request'
```
DonutEspresso: Whoops, sorry, I should have been clearer about my last comment on this. I meant using the logger API itself, so something like this:

```js
req.log.trace({
  plugin: 'inflightRequestThrottle',
  inflightRequests: inflightRequests,
  limit: self._limit
}, 'maximum inflight requests exceeded, rejecting request');
```

retrohacker:
@DonutEspresso one more time?

docs/api/plugins.md
yunong: s/simultaneous connections/inflight requests.

docs/api/plugins.md
```diff
@@ -153,6 +153,13 @@ event, e.g., `server.on('after', plugins.metrics());`:
   * `res` {Object} the response obj
   * `route` {Object} the route obj that serviced the request
 
+The module includes the following plugins to be used with restify's `pre` event:
+* `inflightRequestThrottle(options)` - limits the max number of inflight requests
+  * `options.limit` {Number} the maximum number of simultaneous connections the server will handle before returning an error
+  * `options.resp` {Error} An error that will be passed to `res.send` when the limit is reached.
```
yunong: what does `resp` stand for? Response? If so can use the convention of `res`?

docs/api/plugins.md
```diff
@@ -439,6 +446,24 @@ uniform request distribution.  To enable this, you can pass in
 `options.tokensTable`, which is simply any Object that supports `put` and `get`
 with a `String` key, and an `Object` value.
 
+## Inflight Request Throttling
+
+```js
+const options = { limit: 600, server: server };
```
yunong: The `server` property is not in docs. Do we need to pass in `server` or should we be only passing in the specific properties that the throttle plugin needs?

docs/api/plugins.md
```diff
@@ -439,6 +446,24 @@ uniform request distribution.  To enable this, you can pass in
...
+options.resp = new require('restify-errors').InternalServerError();
```
yunong: Since people copy paste snippets all the time, we should hoist the require up to the top and then call `new` on the constructor. Additionally, I thought this was a response? Why is it now an error object?

lib/plugins/index.js
```diff
@@ -25,6 +25,7 @@ module.exports = {
     serveStatic: require('./static'),
     throttle: require('./throttle'),
     urlEncodedBodyParser: require('./formBodyParser'),
+    inflightRequestThrottle: require('./inflightRequestThrottle'),
```
yunong: We generally sort these alphabetically -- so please do that. And yes it'd be awesome if we had a lint rule for this.

lib/plugins/inflightRequestThrottle.js
```diff
@@ -0,0 +1,66 @@
...
+ * @param {Error} opts.resp the error object to pass to res.send when the
+ *    maximum concurrency is exceeded
+ * @param {Number} opts.resp.statusCode the status code to return when the
+ *    maximum concurrency is exceeded
+ * @param {Function} opts.server.inflightRequests A function that, when
+ *    invoked, returns the number of requests currently being handled by the
+ *    server.
+ * @returns {Function} middleware to be registered on server.pre
+ */
+function inflightRequestThrottle (opts) {
```
yunong: Does this class need a constructor? Generally, the pattern for plugins in restify has been to use a closure, and obviating the need for a constructor. Can we keep it consistent and do that here? We can of course, debate closure vs constructor trade offs at another point in time. Here's an example: https://github.com/restify/node-restify/blob/5.x/lib/plugins/metrics.js#L23

docs/api/plugins.md
yunong: Second this -- let's keep it consistent :)

docs/api/plugins.md
retrohacker: Errors get serialized by `res.__send`

retrohacker:
@yunong one more time?

docs/api/plugins.md
```diff
@@ -153,6 +153,13 @@ event, e.g., `server.on('after', plugins.metrics());`:
   * `res` {Object} the response obj
   * `route` {Object} the route obj that serviced the request
 
+The module includes the following plugins to be used with restify's `pre` event:
+* `inflightRequestThrottle(options)` - limits the max number of inflight requests
+  * `options.limit` {Number} the maximum number of inflight requests the server will handle before returning an error
+  * `options.res` {Error} An error that will be passed to `res.send` when the limit is reached.
```
yunong: Sorry to go one more time on this -- now that I understand this is an actual error, can we just name it options.err instead? That makes it a lot more clear, at least to me. Thoughts? And totally my bad for not realizing this in the first place.

test/plugins/inflightRequestThrottle.test.js
```diff
@@ -0,0 +1,119 @@
+'use strict';
+
+var assert = require('chai').assert;
+var restify = require('../../lib/index.js');
+var restifyClients = require('restify-clients');
+var P = restify.plugins.inflightRequestThrottle;
```
yunong: Why P?

docs/api/plugins.md
```diff
@@ -153,6 +153,13 @@ event, e.g., `server.on('after', plugins.metrics());`:
...
+  * `options.res.statusCode` {Number} The status code to return when the limit is reached.
```
yunong: In terms of StatusCode here -- I almost think we should have the user just pass in a restify error of their choosing, and eliminate having to side channel the status code here. If they want to override the status code, they can just pass in an error they construct which contains the statuscode.

docs/api/plugins.md
sean3z: Agreed, `options.res` doesn't feel intuitive.

retrohacker: Yup, I totally agree. This grew organically since the original implementation expected this to be an array of parameters that would be passed to `res.send.apply`. Now that we expect a restify error, its an awkward fit.

lib/plugins/inflightRequestThrottle.js
```diff
@@ -0,0 +1,61 @@
+'use strict';
+
+var assert = require('assert-plus');
+var ServiceUnavailableError = require('restify-errors').ServiceUnavailableError;
+var defaultResponse = new ServiceUnavailableError('resource exhausted');
+
+/**
+ * inflightRequestThrottle
```
yunong: You may want to specify both here and in docs that users should register this plugin *first* as order matters and you want the throttle check to occur as early as possible.

retrohacker:
@yunong one more time?

retrohacker:
:heart:

Issue #1429, opened 2017-08-02
Setting the status code with res.status has no effect


svozza:
- [x] Used appropriate template for the issue type
- [x] Searched both open and closed issues for duplicates of this issue
- [x] Title adequately and _concisely_ reflects the feature or the bug

# Bug Report

## Restify Version
5.0.1

## Node.js Version
Node 8

## Expected behaviour

When using `res.status` in a route, I expect the status code set there to be the one returned when
hitting the route.

## Actual behaviour

The route always returns a 200.

## Repro case

Here's the test I have in my branch:

```js
test('GH-1429: setting code with res.status not respected', function (t) {
    SERVER.get('/404', function (req, res, next) {
        res.status(404);
        res.send(null);
    });

    CLIENT.get(join(LOCALHOST, '/404'), function (err, _, res) {
        t.equal(res.statusCode, 404); // always returns a 200
        t.end();
    });
});
```

## Cause

The offending code is here:

https://github.com/restify/node-restify/blob/5.x/lib/response.js#L307
https://github.com/restify/node-restify/blob/5.x/lib/response.js#L311

On line 307 if `code` is undefined it gets set to 200 and then on line 311 that overwrites `self.statusCode` even if it has already been set.

## Are you willing and able to fix this?

Yes. In fact, I've already fixed it on my fork.

kokarn:
Encountering this issue as well

retrohacker:
This was me :disappointed:

kokarn:
PR: https://github.com/restify/node-restify/pull/1440

PR #1440, opened 2017-08-08
fix: respect when status code is set with res.status (GH-1429)

svozza:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #1429

# Changes

Before setting the default status code to 200 if there's been no error, it checks to see if `res.status` has been set and uses the value there if it has.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -304,7 +304,7 @@ Response.prototype.__send = function __send() {
 
     // Set sane defaults for optional arguments if they were not provided and
     // we failed to derive their values
-    code = code || 200;
+    code = code || self.statusCode || 200;
     headers = headers || {};
 
     // Populate our response object with the derived arguments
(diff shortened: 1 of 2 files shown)


kwhitley:
@svozza - is this PR required for the [apicache](https://www.npmjs.com/package/apicache) release?

Also, huge thanks for spearheading this - I've been swamped!

kwhitley:
Ahhh nm @svozza - have tested the fix myself, it works! 👍 

Pretty please to the authors for a quick merge and npm patch for this... my users will be very grateful if I don't have to drop restify support to properly cover Node 7.7, 8+! :)

kwhitley:
@svozza You see the out of date notice?  Looks like a merge needed now...

kwhitley:
@retrohacker, any idea on a merge & publish timeline on this?  [Apicache](https://www.npmjs.com/package/apicache) is waiting on a pretty big release that requires this fix!

🙏

sean3z:
Thanks for submitting this @svozza! 👍 
 
I'm unsure whether this route was the original intent coming out of #1377 
However, since it does solve the issue, this seems to be the best solution until a more long-term fix can be put in place

svozza:
Thank you, much appreciated! :smiley:

PR #1441, opened 2017-08-08
update request expiry to expose an isExpired() function for consumers

yunong:


lib/plugins/requestExpiry.js
```diff
@@ -38,31 +38,45 @@ function requestExpiry(options) {
     var timeoutHeaderKey = options.timeoutHeader;
 
     return function requestExpirationCheck(req, res, next) {
-        var expiryTime;
+        /*
+         * Add check expiry API to to req if it doesn't already exist. We only
+         * add this the first time this handler is run, since this handler
+         * could be used in multiple places in the handler chain.
+         */
+        if (!req._expiry) {
```
DonutEspresso: Don't know that we've been consistent about appending these directly on req. Is intention to have audit log consume this at some point? If so, then this is probably okay. If not, just use the closure?

lib/plugins/requestExpiry.js
```diff
@@ -38,31 +38,45 @@ function requestExpiry(options) {
...
-            // to arrive at the expiration time
-            var startTime = req.headers[startHeaderKey];
-            var timeout = req.headers[timeoutHeaderKey];
+            if (useAbsolute) {
+                req._expiry._expiryTime =
+                    Number(req.headers[absoluteHeaderKey]);
+            } else {
+                // Use the start time header and add the timeout header to it
+                // to arrive at the expiration time
+                var startTime = req.headers[startHeaderKey];
```
DonutEspresso: I don't think there is an assertion here to make sure that both `startHeaderKey` and `timeoutHeaderKey` exist (they're asserted independently).

lib/plugins/requestExpiry.js
```diff
@@ -38,31 +38,45 @@ function requestExpiry(options) {
...
-        if (useAbsolute) {
-            expiryTime = Number(req.headers[absoluteHeaderKey]);
-        } else {
-            // Use the start time header and add the timeout header to it
-            // to arrive at the expiration time
-            var startTime = req.headers[startHeaderKey];
-            var timeout = req.headers[timeoutHeaderKey];
+            if (useAbsolute) {
+                req._expiry._expiryTime =
+                    Number(req.headers[absoluteHeaderKey]);
```
DonutEspresso: Is Number intentional, vs parseInt? What if `req.headers[absoluteHeaderKey]` returns NaN?

lib/plugins/requestExpiry.js
DonutEspresso: Same here, what if `req.headers[startHeaderKey]` or `req.headers[timeoutHeaderKey]` return NaN?  

It kind of magically works, since NaN + NaN = NaN, and any comparison against NaN always returns false (in this case, will never expire) but this seems extremely implicit. Not sure I feel super comfortable with the implicitness, though.

lib/plugins/requestExpiry.js
```diff
@@ -38,31 +38,45 @@ function requestExpiry(options) {
...
-                expiryTime = Number(startTime) + Number(timeout);
+                if (startTime && timeout) {
+                    req._expiry._expiryTime =
+                        Number(startTime) + Number(timeout);
+                }
             }
-        }
 
-        if (expiryTime) {
+            req.isExpired = req._expiry.isExpired;
```
DonutEspresso: Seems a little roundabout to have both this function referenced in two places, since you're not using `req._expiry` to encapsulate any state (since you aren't using `this` inside of it either). Any reason not to just assign `req.isExpired` directly at the beginning?

lib/plugins/requestExpiry.js
```diff
@@ -38,31 +38,45 @@ function requestExpiry(options) {
...
+        }
 
+        if (req.isExpired()) {
             // The request has expired
-            if (Date.now() > expiryTime) {
-                return next(new GatewayTimeoutError('Request has expired'));
-            }
+            return next(new GatewayTimeoutError('Request has expired'));
```
DonutEspresso: Do you think there is value in adding how expiry was determined here to log output? i.e., use the `context` feature to add the absolute/relative headers and values

DonutEspresso: Btw, `context` being the restify-errors `context` option (similar to verror's info)

lib/plugins/requestExpiry.js
yunong: Not sure -- this is from the old implementation -- but nice catch.

lib/plugins/requestExpiry.js
yunong: Will context be serialized out to the response? If so I'm not sure we should add it since it'll leak internal details out to clients.

yunong:
@DonutEspresso one more time?

lib/plugins/requestExpiry.js
DonutEspresso: The default `toJSON()` includes only message and code, no internal details. Goes to logs only.

Issue #1435, opened 2017-08-05
Audit log timer values should accumulate not overwrite


lrowe:
# Bug Report

## Restify Version

5.x

## Node.js Version

6.x

## Expected behaviour

Audit log timers should accumulate their value when startHandlerTimer/endHandlerTimer is called more than once for the same timer name.

## Actual behaviour

Audit log timer is overwritten.

## Repro case

To do.

## Cause

https://github.com/restify/node-restify/blob/5169db7b1d2c9979e534b2c27912f5be398bcbca/lib/plugins/audit.js#L85-L90

## Are you willing and able to fix this?

Yes.

retrohacker:
Hey @lrowe,

Could you give some context around why the proposed behavior is desirable? I don't know much about how audit log timers are used.

lrowe:
@retrohacker I'm using handler timers to provide insight into where time is spent within my endpoint handler (since almost all the cpu time is spent in the endpoint handler rather than middleware.)

Let's say I have an expensive function, 'collapse' and I call `req.startHandlerTimer("collapse")` / `req.endHandlerTimer("collapse")` around its invocation to record the time spent within a request. I collapse gets called multiple times in a single request its audit timer no longer shows what proportion of the request was spent executing collapse, only the duration of the final execution.

Ideally each individual time might be logged to an array, though that would involve downstream logging changes.

PR #1443, opened 2017-08-10
fix: audit timers of same name should accumulate (#1435)

lrowe:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #1435

Audit log timers should accumulate their value when startHandlerTimer/endHandlerTimer is called more than once for the same timer name.

# Changes

Accumulates audit log timers by timer name.

### DIFF
--- a/lib/plugins/audit.js
+++ b/lib/plugins/audit.js
@@ -86,7 +86,7 @@ function auditLogger(opts) {
                 (req.timers || []).forEach(function (time) {
                     var t = time.time;
                     var _t = Math.floor((1000000 * t[0]) + (t[1] / 1000));
-                    timers[time.name] = _t;
+                    timers[time.name] = (timers[time.name] || 0) + _t;
                 });
                 return ({
                     // account for native and queryParser plugin usage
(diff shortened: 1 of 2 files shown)


yunong:
@lrowe thanks for the submission. How does this affect a handler chain with multiple handlers with the same name? In this case, we want each of the identically named handlers to have separate timers, which I believe is the current behaviour.

DonutEspresso:
Q: is this technically a breaking change?

lrowe:
This isn't breaking change since the API remains compatible. I don't think it even meets the bar for a feature since there's no additional functionality exposed.

No code will break as a result of this change. The only thing existing users might see is a discontinuity in statistics.

DonutEspresso:
That's true, but audit log output has now changed, so any tooling that was consuming req.timers in the logs will now be seeing something different. While the field name itself hasn't changed, the method of calculating that output has changed (i.e., behavioral change) - I could see an argument both ways - or am I off my rocker?

lrowe:
It is a behavioural change, but I really do consider the previous behaviour of overwriting same named timings errant. Pragmatically, I don't expect Restify consumers would need to change any code as a result this fix, so bracketing its release in with breaking changes doesn't seem worthwhile.

PR #1444, opened 2017-08-11
BREAKING CHANGE: bump restify-errors version to v5

retrohacker:
Updating dependencies. This changes the format of custom errors when they are serialized as a string.

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
     "negotiator": "^0.6.1",
     "once": "^1.3.0",
     "qs": "^6.2.1",
-    "restify-errors": "^4.2.3",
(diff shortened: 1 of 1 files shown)


DonutEspresso:
Is the conventional changelog `breaking change:` or just `breaking:` ? Otherwise, LGTM.

retrohacker:
According to the website, it is `BREAKING CHANGE` though we may be using another flavor that I'm not aware of: http://conventionalcommits.org/

retrohacker:
Going to wait until we get 5.x merged into master for this one, since it bumps us to 6.x

Would also be nice to ship all the semver patch/minor stuff to the 5.x release before doing a major bump.

PR #1454, opened 2017-08-17
New: server `strictNext` option throws when calling next more than once

DonutEspresso:
## Pre-Submission Checklist

# Changes

This introduces a new `strictNext` option that causes the server option to throw when next is called more than once. This defaults to true (breaking change for 6.x).

This uncovered a bug during the scenario when `next(err)` is invoked, and the server is emitting error events. It was in fact calling the internal completion callback twice (with no side effects, AFAIK).


lib/server.js
```diff
@@ -253,8 +253,9 @@ function Server(options) {
     this.router = options.router;
     this.routes = {};
     this.secure = false;
-    this.versions = options.versions || options.version || [];
     this.socketio = options.socketio || false;
+    this._once = (options.strictNext === false) ? once : once.strict;
```
yunong: at this point since you're in here anyway, might want to drop the `once` dependency and just use the function provided by `lodash`

lib/server.js
```diff
@@ -919,7 +920,9 @@ Server.prototype._run = function _run(req, res, route, chain, cb) {
                 // flag is true, we will avoid automatically calling cb at the
                 // end of this function, which causes server to move on to the
                 // next handler in the chain.
-                emittedError = self._emitErrorEvents(req, res, route, arg,
+                emittingErrors = true;
```
yunong: Why is this change here necessary? Don't we just need to change the name?

lib/server.js
```diff
@@ -851,9 +852,9 @@ Server.prototype._route = function _route(req, res, name, cb) {
  * for case #3, emitted error events are async and serial. this means the
  * automatic invocation of cb() cannot occur:
  *
- * 1) `emittedError` - this boolean is set to true when next is called with an
- * error. this var is used to avoid automatic invocation of cb(), which is
- * delayed until all async error events are fired.
+ * 1) `emittingErrors` - this boolean is set to true when the server is still
+ * emitting error events. this var is used to avoid automatic invocation of
```
yunong: I'm missing some context here about this change, mind elaborating?

lib/server.js
DonutEspresso: lodash's version doesn't offer a strict implementation

lib/server.js
DonutEspresso: There's a race condition here due to the way the code is structured. Basically, restify's internal cb is called in two places. 

When you `next(err)`, assuming you have error listeners attached, we will async fire the events serially, flush response, then call cb (L923-L925). 

If no listeners are attached, then `_emitErrorEvents` will return false, and cb is fired down at L1030. After `_emitErrorEvents` completes and calls its callback on L923, it will then call cb again at L928. Calling cb twice will throw, as intended. 

We are using the `emittingErrors` closure var to basically tell one pathway to not call cb since it will be called later on. This is not great and really hard to grok, but not something to solve in this PR.

yunong: Alright, I'll take your word for it. :)

DonutEspresso:
@retrohacker speak now or forever hold your peace?  :)

retrohacker:
Failures are unrelated to this PR. Merging.

PR #1460, opened 2017-08-24
cpuUsageThrottle

retrohacker:
## Pre-Submission Checklist

# Changes

Adds a CPU usage throttle which begins dropping requests at the front door when a process begins using more than the defined limit. This helps prevent unacceptable latencies from requests backing up.


retrohacker:
Failing tests are unrelated to this PR: https://github.com/nodesecurity/nsp/issues/185

docs/api/plugins.md
```diff
@@ -468,6 +474,31 @@ requests. It defaults to `503 ServiceUnavailableError`.
...
+  max: 1,
+  interval: 250,
+  halfLife: 500,
+  err: new errors.InternalServerError()
+}
+
+server.pre(restify.plugins.cpuUsageThrottle(options));
+```
+
+Have restify reject a variable number of requests at the front door based on the process' current CPU usage. This plugin works by keeping a running average of the CPU usage over time and rejecting a percentage of requests based on the user specified CPU limit, maximum CPU limit, and the current value. This module will reject between 0% and 100% of all incomming requests when attempting to bring CPU usage back to a reasonable level. The method we use for calculating the CPU usage will work across platforms. This is an effective module for preventing high latencies caused by an over saturated restify server. This module is most effective when paired with an auto-scale policy at an orechestration layer.
```
DonutEspresso: s/orechestration/orchestration

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,139 @@
...
+ * Have restify reject a variable number of requests at the front door based on
+ * the process' current CPU usage. This plugin works by keeping a running
+ * average of the CPU usage over time and rejecting a percentage of requests
+ * based on the user specified CPU limit, maximum CPU limit, and the current
+ * value. This module will reject between 0% and 100% of all incomming requests
+ * when attempting to bring CPU usage back to a reasonable level. The method
+ * we use for calculating the CPU usage will work across platforms. This is
+ * an effective module for preventing high latencies caused by an over saturated
+ * restify server. This module is most effective when paired with an auto-scale
+ * policy at an orechestration layer.
```
DonutEspresso: s/orechestration/orchestration

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,139 @@
...
+ *    cpu usage limit is exceeded
+ * @returns {Function} middleware to be registered on server.pre
+ */
+function cpuUsageThrottle (opts) {
+
+    // Scrub input and populate our configuration
+    assert.object(opts, 'opts');
+    assert.number(opts.limit, 'opts.limit');
+    assert.optionalNumber(opts.max, 'opts.max');
+    assert.number(opts.interval, 'opts.interval');
```
DonutEspresso: Is it possible to have a sane default for this? Just wondering we propose folks to figure out what this value should be.

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,139 @@
...
+ * we use for calculating the CPU usage will work across platforms. This is
+ * an effective module for preventing high latencies caused by an over saturated
+ * restify server. This module is most effective when paired with an auto-scale
+ * policy at an orechestration layer.
+ *
+ * @param {Object} opts Configure this plugin.
+ * @param {Number} opts.limit The point at which restify will begin rejecting a
+ *    % of all requests at the front door. This value is a percentage like you
+ *    would see when running `top` on linux. For example .8 === 80%.
+ * @param {Number} opts.max The point at which restify will reject 100% of all
```
DonutEspresso: Use `[]` to indicate optional options in your JSDoc

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,139 @@
...
+    self._reject = 0;
+
+    // updateReject should be called on an interval, it checks the average CPU
+    // usage between two invocations of updateReject.
+    function updateReject() {
+        pidusage.stat(process.pid, function (e, stat) {
+            // If we fail to stat for whatever reason, err on the side of
+            // accepting traffic, while failure probably indicates something
+            // horribly wrong with the server, identifying this error case and
+            // appropriately responding is way outside the scope of this plugin
```
DonutEspresso: Is it worth emitting an error in this case? Might be a way to hook up metrics for when the plugin is misbehaving for whatever reason.

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,139 @@
...
+            if (e) {
+                // When cpu is NaN, we should never reject
+                stat = { cpu: NaN };
+            }
+
+            // Divide by 100 to match linux's top format
+            self._cpu = stat.cpu / 100;
+            // Update reject with the % of traffic we should be rejecting. This
+            // is safe since max > limit so it can never be division by 0. If
```
DonutEspresso: divided?

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,139 @@
...
+            // the current cpu usage is less that the limit, _reject will be
+            // negative and we will never shed load
+            self._reject =
+                (self._cpu - self._limit) / (self._max - self._limit);
+            self._timeout = setTimeout(updateReject, self._interval);
```
DonutEspresso: Under heavy load `updateReject()` might execute later than you'd like it to - is it valuable to incorporate that delta into the algo?

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,139 @@
...
+        });
+    }
+
+    // Kick off updating our _reject value
+    updateReject();
+
+    function onRequest (req, res, next) {
+        // Check to see if this request gets rejected
+        if (self._reject > Math.random()) {
```
DonutEspresso: Some comments around `Math.random()` and why it exists would be helpful

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,139 @@
...
+            // If so, log some metadata and return an error
+            req.log.trace({
+                plugin: 'cpuUsageThrottle',
+                cpuUsage: self._cpu,
+                limit: self._limit
+            }, 'maximum cpu usage exceeded, rejecting request');
+            res.send(self._err);
```
DonutEspresso: When working with errors, let's use `return next(self._err)`

DonutEspresso: This is so the err events emitted by server will work correctly, but should also make your tests a little easier so you don't have to mock res.send

test/plugins/cpuUsageThrottle.test.js
```diff
@@ -0,0 +1,132 @@
...
+        stat: function (pid, cb) {
+            return cb(ERROR, { cpu: CPU });
+        }
+    }
+});
+
+var MR = Math.random;
+describe('cpuUsageThrottle', function () {
+
+    it('Setup: stub math.random', function (done) {
```
DonutEspresso: This could be a `before()`

test/plugins/cpuUsageThrottle.test.js
```diff
@@ -0,0 +1,132 @@
...
+                assert.equal(e.name,
+                    'InternalServerError', 'Default err returned');
+                assert.equal(res.statusCode, 555,
+                    'Default shed status code returned');
+                clearTimeout(plugin._timeout);
+            });
+        });
+    });
+
+    it('Teardown: Reset Math.random', function (done) {
```
DonutEspresso: `after()`

lib/plugins/cpuUsageThrottle.js
retrohacker: Probably right... I keep coming back to 500ms, so recheck every 1/2 second. Though I'm not sure if that is aggressive enough.

lib/plugins/cpuUsageThrottle.js
retrohacker: The plugin reads the metrics from either stdout or a file descriptor, the error we would get here would be the same as a read from the filesystem failing. Something horribly wrong is happening, and more than likely is showing up somewhere else that is in a better position to handle it. I'd be open to logging this, though I don't believe we have access to the log object at this point. We would need to either accept it in the constructor or cache it on the first request.

lib/plugins/cpuUsageThrottle.js
retrohacker: Yup, EWMA accounts for this :-)

retrohacker: It is the reason for the crazy magic maths here: https://github.com/ReactiveSocket/ewma/blob/master/index.js#L55

lib/plugins/cpuUsageThrottle.js
retrohacker: Oh cool, didn't know `next` handled sending when an error was passed in! https://github.com/restify/node-restify/blob/master/lib/server.js#L925-L926

lib/plugins/cpuUsageThrottle.js
retrohacker: Since this is a super low overhead function and results in reading from a single file descriptor, I'm going to have it default to checking every 1/4 of a second.

lib/plugins/cpuUsageThrottle.js
yunong: Even tho ewma accounts for this in the algorithm, the interval is a parameter configured by the user, and we should strive to stick to said interval. I think it still makes sense to keep track of the delta here, since changes to the interval will affect the throttling algorithm.

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,149 @@
...
+        // can use Math.random() (which picks from a uniform distribution in
+        // [0,1)) to give us a `self._reject`% chance of dropping any given
+        // request. This is a stateless was to drop approximatly `self._reject`%
+        // of traffic.
+        if (Math.random() >= self._reject) {
+            return next() // Don't reject this request
+        }
+
+        // If we are rejecting the request, log some metadata and return error
+        req.log.trace({
```
yunong: Instead of logging trace here, you should just put these parameters in a context object on the error, and that way the audit log will automatically emit it. This has the additional benefit of consumers being able to listen to the 'after' event and emit a metric based on the fact that this got throttled.

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,138 @@
...
+    // self._reject represents the % of traffic that we should reject at the
+    // current point in time based on how much over our limit we are. This is
+    // updated on an interval by updateReject().
+    self._reject = 0;
+
+    // updateReject should be called on an interval, it checks the average CPU
+    // usage between two invocations of updateReject.
+    function updateReject() {
+        pidusage.stat(process.pid, function (e, stat) {
+            // If we fail to stat for whatever reason, err on the side of
```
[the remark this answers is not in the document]
retrohacker: https://github.com/restify/node-restify/pull/1460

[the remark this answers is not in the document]
retrohacker: https://github.com/restify/node-restify/pull/1460#discussion_r137375716

lib/plugins/cpuUsageThrottle.js
retrohacker: After a conversation with @DonutEspresso, we settled on exposing this module's state through https://github.com/restify/node-restify/pull/1460/files#diff-01a30fd8ce166416dd691d3498c746c7R144

This allows for tooling outside of this plugin/repo to check and log/monitor/etc. the behavior of this plugin. When  fails, we can can check `cpu === NaN` in that module and log it out.

lib/plugins/cpuUsageThrottle.js
retrohacker: > Even tho ewma accounts for this in the algorithm, the interval is a parameter configured by the user, and we should strive to stick to said interval.

Agreed. That is what I'm trying to accomplish with this code, though there may be a better way. Since `pid.stat` is async and it's performance is not necessarily bound to process saturation, its possible the first call to it could take a few minutes to return. With `setInterval` we have the potential to queue a whole bunch of `pid.stat` calls up while that first blocks. Then it ends up being a race to burn through that queue. This tries to ensure we call `pid.stat` exactly once at a time and that the time between it returning back to us and us calling it again is as close to the interval the user requested as possible.

retrohacker: > keep track of the delta

That would be valuable, though I'm not sure how we could reliably expose that metric.

Some ideas:

* We could attach it to `self`, but whatever logs `self` at the end of the day won't be bound to this interval.
* We could emit an event, but that would increase the surface area of our plugin API and would probably be something we want to generalize. (also worth noting that `plugin.state` in this PR is doing this already)

retrohacker: > changes to the interval will affect the throttling algorithm

While it may change the calculation slightly, if it has a meaningful impact on the behaviour of throttling we will need revisit this implementation. It is currently designed to be stable in the face of unevenly spaced data points, and since our data points actually represent several million values (either 1 or 0 indicating whether Node used a cycle of the CPU) averaged over the interval, every "point" is still meaningfully represented in the calculation irrespective of latency. The decay of all of those "points" represented in a single average will be the same as if they all occurred at the same moment in time. That being said, from an external perspective, this _should_ result in nearly identical behaviour. Like I said, if this proves to not be the case, we will need to re-think this implementation.

Worth noting that in order to measure whether this has any "meaningful impact" we need to log the delta somewhere :-P

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,138 @@
...
+    self._err = opts.err || defaultResponse;
+    self._limit = opts.limit;
+    self._max = opts.max || 1;
+    self._interval = opts.interval;
+    self._halfLife = (typeof opts.halfLife === 'number') ?
+      opts.halfLife : opts.interval;
+    self._timeout = null;
+    assert.ok(self._max > self._limit, 'limit must be less than max');
+
+    // We start at 0 to compensate for the initial spike in CPU usage a server
```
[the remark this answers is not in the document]
retrohacker: Possibly. In practice I've seen values that are arguably incorrect (i.e. 400% CPU utilization) when we check on a fresh process start. The intent here was that the 0 value would reduce how much those values skewed the results giving us a more sane EWMA value moving forward.

Alternatively, taking those initial outliers as the initial value would have the effect of incrementally accepting traffic at boot which may be desirable. The impact of this decision is greater when the interval and halfLife are larger.

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,138 @@
...
+ *    unit is in ms. The default is to set this to the same value as interval.
+ * @param {Error} [opts.err] A restify error used as a response when the
+ *    cpu usage limit is exceeded
+ * @returns {Function} middleware to be registered on server.pre
+ */
+function cpuUsageThrottle (opts) {
+
+    // Scrub input and populate our configuration
+    assert.object(opts, 'opts');
+    assert.number(opts.limit, 'opts.limit');
```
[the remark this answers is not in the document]
retrohacker: I'm not confident we can have a sane default here. A server that can handle 2RPS binding large data sets to large templates may want a very different limit than a server that does 8k RPS validating auth for login. I may be surprised.

lib/plugins/cpuUsageThrottle.js
retrohacker: :+1:

retrohacker: Actually, this module is miss-using errors. I should be creating a new error on every rejected request instead of returning the same error. Wil.l have to change that if I want to pass a context through.

lib/plugins/cpuUsageThrottle.js
retrohacker: It is probably worth clarifying why `interval` is configurable if there isn't a significant impact of a high delta between when the user requested interval and the actual time between checks due to resource contention.

While I don't believe contention increasing this delta will have a meaningful impact on the behaviour of load shedding, I do believe that, during periods where there isn't contention for resources, a more frequent check may be desirable and will have a meaningful impact. So essentially, the user may want to _lower_ this value to be able to measure CPU spikes with a higher resolution.

retrohacker:
@yunong @DonutEspresso one more time?

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,183 @@
...
+    self._err = opts.err || defaultResponse;
+    self._limit = opts.limit;
+    self._max = opts.max || 1;
+    self._interval = opts.interval || 250;
+    self._halfLife = (typeof opts.halfLife === 'number') ?
+      opts.halfLife : opts.interval;
+    self._timeout = null;
+    assert.ok(self._max > self._limit, 'limit must be less than max');
+
+    self._ewma = new EWMA(self._halfLife);
```
retrohacker: 0_0 why am I not using this?

retrohacker:
Rebased against master

lib/plugins/cpuUsageThrottle.js
DonutEspresso: What about just exposing the delta for now then (through EE or similar) and we can keep track of it as we first deploy? As we gather data it should inform us just how much of an impact there will be.

DonutEspresso: In other words, no need to consume it in a meaningful way now, but at least we have some visibility into what's going on. Although, now that I think about it - would something like this already be captured in the event loop metrics?

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,235 @@
...
+            // accepting traffic, while failure probably indicates something
+            // horribly wrong with the server, identifying this error case and
+            // appropriately responding is way outside the scope of this plugin
+            if (e) {
+                // When cpu is NaN, we should never reject
+                stat = { cpu: NaN };
+            }
+
+            // Divide by 100 to match linux's top format
+            self._ewma.insert(stat.cpu / 100);
```
DonutEspresso: What happens when you insert NaN?

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,235 @@
...
+    // Allow the app to clear the timeout for this plugin if necessary, without
+    // this we would never be able to clear the event loop when letting Node
+    // shut down gracefully
+    function close () {
+        clearTimeout(self._timeout);
+    }
+    onRequest.close = close;
+
+    // Expose internal plugin state for introspection
+    onRequest.state = self;
```
DonutEspresso: Are any of the exposed fields actually public? They all look like _ properties

docs/api/plugins.md
```diff
@@ -525,6 +531,46 @@ requests. It defaults to `503 ServiceUnavailableError`.
...
...ver 250ms intervals. Due to contention for resources, the duration of each average may be wider or narrower than 250ms.  To compensate for this, we use an exponentially weighted moving average. The EWMA algorithm is provided by the ewma module. The parameter for configuring the EWMA is halfLife. This value controls how quickly each load average measurment decays to half it's value when being represented in the current average. For example, if you have an interval of 250, and a halfLife of 250, you will take the previous ewma value multiplied by  .5 and add it to the new CPU utilization average measurement multiplied by .5. The previous value and the new measurement would each represent 50% of the new value. A good way of thinking about the halfLife is in terms of how responsive this plugin will be to spikes in CPU utilization. The higher the halfLife, the longer CPU utilization will have to remain above your defined limit before this plugin begins rejecting requests and, converserly, the longer it will have to drop below your limit before the plugin begins accepting requests again. This is a knob you will want to with play when trying to determine the ideal value for your use case.
```
yunong: Love the new summary. Instead of saying `linux's top`, we could be more precise and link load average to this: http://www.brendangregg.com/blog/2017-08-08/linux-load-averages.html

docs/api/plugins.md
```diff
@@ -525,6 +531,46 @@ requests. It defaults to `503 ServiceUnavailableError`.
...
+
+For a better understanding of the EWMA algorithn, refer to the documentation for the ewma module.
+
+Params:
+  * `max` - The point at which restify will reject 100% of all requests at the front door. This is used in conjunction with limit to determine what % of traffic restify needs to reject when attempting to bring the average load back within tolerable thresholds. Since Node.js is single threaded, the default for this is 1. In some rare cases, a Node.js process can exceed 100% CPU usage and you will want to update this value.
```
yunong: Where is `limit` defined?

docs/api/plugins.md
```diff
@@ -525,6 +531,46 @@ requests. It defaults to `503 ServiceUnavailableError`.
...
+  * `interval` - How frequently to check if we should be rejecting/accepting connections. It's best to keep this value as low as possible without creating a significant impact on perfomance in order to keep your server as responsive to change as possible.
```
yunong: this might be clearer if it was `how frequently we update the load average, which drives if we should be throttling connections`. `updating the load average is not cheap, and keeping this value as low as possible reduces the performance impact`

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,235 @@
...
+ * longer it will have to drop below your limit before the plugin begins
+ * accepting requests again. This is a knob you will want to with play when
+ * trying to determine the ideal value for your use case.
+ *
+ * For a better understanding of the EWMA algorithn, refer to the documentation
+ * for the ewma module.
+ *
+ * @param {Object} opts Configure this plugin.
+ * @param {Number} opts.limit The point at which restify will begin rejecting a
+ *    % of all requests at the front door. This value is a percentage like you
```
yunong: I think here it would be less confusing and more precise to just use the temr `CPU utilization`, and link to documentation: http://www.brendangregg.com/blog/2017-05-09/cpu-utilization-is-wrong.html

lib/plugins/cpuUsageThrottle.js
yunong: Sorry, I think there's been some confusion. I am suggesting we keep track of how much time has elapsed so that we can update `interval` appropriately. e.g. if the interval is 500ms, but it took us 300ms to go through this current interval, we should then set timeout to 200ms, so that we're ensuring we're firing the interval every 500ms. Otherwise in this example we would be running the next interval 800ms since the last interval.

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,235 @@
...
+    updateReject();
+
+    function onRequest (req, res, next) {
+        // Check to see if this request gets rejected. Since, in updateReject,
+        // we calculate a percentage of traffic we are planning to reject, we
+        // can use Math.random() (which picks from a uniform distribution in
+        // [0,1)) to give us a `self._reject`% chance of dropping any given
+        // request. This is a stateless was to drop approximatly `self._reject`%
+        // of traffic.
+        var draw = Math.random();
```
yunong: maybe `probabilityDraw`?

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,235 @@
...
+    /**
+     * cpuUsageThrottle.update
+     *
+     * Allow the plugin's configuration to be updated during runtime.
+     *
+     * @param {Object} opts The opts object for reconfiguring this plugin, it
+     *    follows the same format as the constructor for this plugin.
+     */
+    onRequest.update = function update(newOpts) {
+        assert.object(newOpts, 'newOpts');
```
yunong: For this function you might actually want to log/emit that we've updated the parameters, to ease debugging later on.

docs/api/plugins.md
```diff
@@ -525,6 +531,46 @@ requests. It defaults to `503 ServiceUnavailableError`.
...
+  max: 1,
+  interval: 250,
+  halfLife: 500,
+  err: new errors.InternalServerError()
+}
+
+server.pre(restify.plugins.cpuUsageThrottle(options));
+```
+
+cpuUsageThrottle is a middleware that rejects a variable number of requests (between 0% and 100%) based on a historical view of CPU utilization of a Node.js process. Essentially, this plugin allows you to define what constitutes a saturated Node.js process via CPU utilization and it will handle dropping a % of requests based on that definiton. This is useful when you would like to keep CPU bound tasks from piling up causing an increased per-request latency.
```
yunong: For bonus points might want to link to the ewma paper, but if you're already doing that in the ewma module, then maybe not. I leave that up to you :)

lib/plugins/cpuUsageThrottle.js
retrohacker: Don't have access to the logger here :-\

I can start caching the logger on the first request though.

docs/api/plugins.md
retrohacker: Done in the EWMA module :smile: the :rabbit: :hole: is deep :wink:

lib/plugins/cpuUsageThrottle.js
retrohacker: There could be latency introduced by the file descriptor read on Linux or spinning up the subprocess on windows (inside or `pidusage.stat`). I'll export it as a value off of `state`

retrohacker: Also adding it to the error context when we shed load

lib/plugins/cpuUsageThrottle.js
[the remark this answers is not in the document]
retrohacker: After talking it through, we settled on a default of .75 for right now. We will open a PR later if we find a better default.

docs/api/plugins.md
retrohacker: It's actually not using linux's top's formula. Used the word "similar" here since it isn't the same, but it gives people a point of reference when trying to understand the metric. The actual algorithm used is: https://stackoverflow.com/questions/16726779/how-do-i-get-the-total-cpu-usage-of-an-application-from-proc-pid-stat/16736599#16736599

I could link to it, but if someone is diving that deep, I would suspect they would refer to the `pidusage` documentation which does define the algorithm.

docs/api/plugins.md
retrohacker: :blush:

lib/plugins/cpuUsageThrottle.js
retrohacker: Agreed this is confusing. I removed the reference to linux top for the reasons detailed in https://github.com/restify/node-restify/pull/1460#discussion_r137885294

retrohacker:
@DonutEspresso @yunong one more time?

lib/plugins/cpuUsageThrottle.js
```diff
@@ -0,0 +1,268 @@
...
+        }
+
+        if (newOpts.interval !== undefined) {
+            self._interval = newOpts.interval;
+        }
+
+        if (newOpts.halfLife !== undefined) {
+            self._halfLife = newOpts.halfLife;
+            // update our ewma with the new halfLife, we use the previous known
+            // state as the initial state for our new halfLife in liue of
```
DonutEspresso: s/liue/lieu

DonutEspresso:
Chatted offline re: removing the `err` option and just relying on restify's built in `server.on('restifyError'...` capability to do error customization.

PR #1462, opened 2017-08-24
WIP: inflight request handler data

DonutEspresso:
WIP, just a prototype. Need to store this up here for now.


yunong:
It occurs to me that we could also add an additional field, which is the rate of change of a specific counter/route. That way, we will be able to tell if there are potential leaks by looking at the derivative.

Issue #1459, opened 2017-08-22
restify documentation appears to be missing HTTP verb methods


jclulow:
I was reading the Restify server documentation, i.e., http://restify.com/docs/server-api/, and it seems that the route handler mounting methods (e.g., `.get()`, `.post()`, etc) are completely missing from the server API documentation.  They are covered to some extent in the _Getting Started_ section, but appear to be absent entirely from the reference portion of the documentation.

hekike:
Fixed by JSDoc based docs generation.

PR #1463, opened 2017-08-24
Chore: restore server guide from old restify.com

DonutEspresso:
This restores the server guide from the old restify.com. Also addresses #1459

### DIFF
--- a/docs/api/server.md
+++ b/docs/api/server.md
@@ -12,6 +12,8 @@ var restify = require('restify');
 var server = restify.createServer();
 ```
 
+## createServer()
+
 The `createServer()` method takes the following options:
 
 |Option|Type|Description|
@@ -29,7 +31,7 @@ The `createServer()` method takes the following options:
 |strictRouting|Boolean|(Default=`false`). If set, Restify will treat "/foo" and "/foo/" as different paths.|
 
 
-# Properties
+## Properties
 
 A restify server has the following properties on it:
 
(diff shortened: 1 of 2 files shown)


docs/api/server.md
```diff
@@ -41,25 +43,58 @@ A restify server has the following properties on it:
...
+// a static route
+server.get('/foo', function(req, res, next) {});
+// a parameterized route
+server.get('/foo/:bar', function(req, res, next) {});
+// a regular expression
+server.get(/^\/([a-zA-Z0-9_\.~-]+)\/(.*)/, function(req, res, next) {});
+// an options object
+server.get({
+    path: '/foo',
+    version: ['1.0.0', '2.0.0]
```
retrohacker: Hmm something weird with the markdown here?

docs/api/server.md
```diff
@@ -41,25 +43,58 @@ A restify server has the following properties on it:
...
+```
+
+### address()
 
 Wraps node's [address()](http://nodejs.org/docs/latest/api/net.html#net_server_address).
 
-## listen(port, [host], [callback])
+### listen(port, [host], [callback])
 
 Wraps node's [listen()](http://nodejs.org/docs/latest/api/net.html#net_server_listen_path_callback).
```
retrohacker: It doesn't look like GitHub's markdown likes `'`?

docs/api/server.md
sean3z: Missing a `'` after `2.0.0`
```
['1.0.0', '2.0.0']
```

docs/api/server.md
sean3z: Correcting the missing `'` above will fix this 😃

DonutEspresso: Thanks @sean3z!

DonutEspresso:
BTW the intention for this is to get something back up that at least contains the information - we can work on the copy iteratively.

PR #1468, opened 2017-08-28
test(plugin): avoid using 127.0.0.2 in throttle tests

hekike:
# Changes

- avoid using loopback interface IP 127.0.0.2 in the plugin/throttle tests, it doesn't work on OSX by default

### DIFF
--- a/test/plugins/throttle.test.js
+++ b/test/plugins/throttle.test.js
@@ -189,38 +189,46 @@ describe('throttle plugin', function () {
         });
     });
 
-    it('should expose headers on options set', function (done) {
-        // setup a new server with headers set to true since we cant
-        // change throttle options after init
-        setupClientServer('127.0.0.2', {
-            burst: 17,
-            rate: 0.1,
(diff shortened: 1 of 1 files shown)


test/plugins/throttle.test.js
```diff
@@ -189,38 +189,46 @@ describe('throttle plugin', function () {
...
+        before(function (done) {
+            // close global server before creating a new to avoid port conflicts
+            CLIENT.close();
+            SERVER.close(done);
+        });
+
+        it('should expose headers on options set', function (done) {
+            // setup a new server with headers set to true since we cant
+            // change throttle options after init
+            setupClientServer('127.0.0.1', {
```
hekike: Hint: Changed from `127.0.0.2` to `127.0.0.1`

DonutEspresso:
Thanks @hekike! I was wondering why my tests were failing locally - I suspect not many folks have 127.0.0.2 loopback setup.

PR #1471, opened 2017-09-01
Inflight request throttle fix

retrohacker:
* Always call next
* Abort middleware chain by passing `false` when shedding load
* Fix error messages
* Remove accidental onRequest on prototype


retrohacker:
Failing tests are nsp shrinkwrap issues which should be resolved upstream. Merging.

PR #1474, opened 2017-09-03
docs: more detailed plugin docs

DonutEspresso:


docs/api/plugins.md
```diff
@@ -32,21 +32,141 @@ server.use(restify.plugins.throttle({
...
+Exmaple:
+
+```js
+server.pre(restify.plugins.pre.dedupeSlashes());
+server.get('/hello/:one', function(req, res, next) {
+    res.send(200);
+    return next();
+});
+
+// the server will will now convert requests to /hello//jake => /hello/jake
```
sean3z: > will will

😄

docs/api/plugins.md
```diff
@@ -32,21 +32,141 @@ server.use(restify.plugins.throttle({
...
+  }
+}
+```
+
+`req.username` will also be set, and defaults to 'anonymous'.  If the scheme
+is unrecognized, the only thing available in `req.authorization` will be
+`scheme` and `credentials` - it will be up to you to parse out the rest.
+
+
+### `restify.plugins.dateParser(sec)
```
sean3z: Missing the closing ` here

DonutEspresso:
Thanks all! 🎉

PR #1475, opened 2017-09-06
fix: exclude shrinkwrap from project

retrohacker:
A `package-lock.json` file found its way into the project. Also, `npm shrinkwrap --dev` is causing our ci/cd to fail, `nsp` doesn't appear to require it from what I can tell.


retrohacker:
https://github.com/nodesecurity/nsp/issues/185

retrohacker:
package-lock.json still should not be committed

Issue #1464, opened 2017-08-25
Optional parameter issue with NON-trailing double slash


lordbah:
# Bug Report
Double slash confuses parameter parser.
## Restify Version
5.1.0
## Node.js Version
8.2.1
## Expected behaviour
Don't match this pattern.

## Actual behaviour
Matched, and assigned later part of URL to earlier parameters.

## Repro case
const restify = require('restify');
const server = restify.createServer();

server.put('/products/:catalog/:item', function(req, res, next) {
    console.log(\`URL=${req.url}\`);
    console.log(\`catalog='${req.params.catalog}' item='${req.params.item}'\`);
});

server.listen(8989, () => {
    console.log('listening');
});

Then used Postman to send some PUT requests with these results:
URL=/products/one/two
catalog='one' item='two'
URL=/products//two
catalog='' item='two'
URL=/products//two/three
catalog='two' item='three'

The first two results are what I expect. In the third case I expect no match.

/products/one/two/three was of course not matched by this pattern. The third case should be the same except that it should match an empty string to catalog. The second case shows that it is capable of matching an empty string.

## Cause
I don't know.

## Are you willing and able to fix this?
No. I mean, at my current level of Javascript experience I don't think you want me messing around in there.

DonutEspresso:
Hi @lordbah, thanks for the report! This certainly does appear to be unexpected behavior. I'll have to take some time to look into what's happening. 

In the meantime, a possible work around for you might be to opt-in to URL sanitization by using the `dedupeSlashes` plugin:

```js
server.pre(restify.plugins.pre.dedupeSlashes());
```

This will turn `/products//two` into `/products/two`, and `/products//two/three` into `/products/two/three`, which will allow you to entirely bypass this somewhat ambiguous behavior.

hekike:
@DonutEspresso here is a fix: https://github.com/restify/node-restify/pull/1478
But non-trailing slashes were tested before for skipping, we should agree about the expected behavior first.

I believe with this change it would be more consistent.
What do you think?

retrohacker:
Closing this as resolved by #1512 

If this is not the case, please comment and we can re-open :smile:

PR #1478, opened 2017-09-09
fix(router): handle non-trailing slashes correctly

hekike:
## Issues

Closes:

https://github.com/restify/node-restify/issues/1464

Non-trailing slash leads to incorrect routing match (and parameter parsing).

# Changes

Fix router regexp matching with non-trailing slash(es)

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -134,7 +134,7 @@ function compileURL(options) {
     }
 
     if (!options.strict) {
-        pattern += '[\\/]*';
+        pattern += '[\\/]?';
     }
 
     if (pattern === '^') {
--- a/lib/router.js
+++ b/lib/router.js
@@ -103,7 +103,7 @@ function compileURL(options) {
             return (false);
         }
 
-        pattern += '\\/{1}';
+        pattern += '\\/';
 
         if (frag.charAt(0) === ':') {
             var label = frag;
--- a/lib/router.js
+++ b/lib/router.js
@@ -103,7 +103,7 @@ function compileURL(options) {
             return (false);
         }
 
-        pattern += '\\/+';
+        pattern += '\\/{1}';
 
         if (frag.charAt(0) === ':') {
             var label = frag;
--- a/test/router.test.js
+++ b/test/router.test.js
@@ -210,13 +210,26 @@ test('Default non-strict routing ignores trailing slash(es)', function (t) {
 
     var trailing = server.router.routes.GET[0];
     t.ok(trailing.path.test('/trailing/'));
-    t.ok(trailing.path.test('//trailing//'));
+    t.ok(trailing.path.test('/trailing//'));
     t.ok(trailing.path.test('/trailing'));
 
     var noTrailing = server.router.routes.GET[1];
     t.ok(noTrailing.path.test('/no-trailing'));
-    t.ok(noTrailing.path.test('//no-trailing//'));
+    t.ok(noTrailing.path.test('/no-trailing//'));
     t.ok(noTrailing.path.test('/no-trailing/'));
+
+    t.end();
+});
+
+test('Default non-strict routing doesn\'t match non-trailing slash(es)',
+  function (t) {
+    var server = restify.createServer();
+    function noop () {}
+
+    server.get('/no-trailing', noop);
+
+    var noTrailing = server.router.routes.GET[0];
+    t.notOk(noTrailing.path.test('//no-trailing'));
 
     t.end();
 });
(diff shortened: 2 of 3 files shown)


test/router.test.js
```diff
@@ -210,17 +210,30 @@ test('Default non-strict routing ignores trailing slash(es)', function (t) {
 
     var trailing = server.router.routes.GET[0];
     t.ok(trailing.path.test('/trailing/'));
-    t.ok(trailing.path.test('//trailing//'));
+    t.ok(trailing.path.test('/trailing//'));
     t.ok(trailing.path.test('/trailing'));
 
     var noTrailing = server.router.routes.GET[1];
     t.ok(noTrailing.path.test('/no-trailing'));
-    t.ok(noTrailing.path.test('//no-trailing//'));
```
retrohacker: I'm not sure I have enough context to understand why this change is necessary. Why did we change an existing test as opposed to adding a separate test for a single leading `/`?

hekike: The original test case is not related to the code what it tests. This test should focus to double no-trailing slashes.

hekike: Sry, I mean it should focus to slashes at the end of the path :P

DonutEspresso:
@hekike thanks for taking this on! I think you're right, the behavior should be to not remove double slashes by default - that should be taken care of by the dedupeSlashes plugin. This is going to be a breaking change, perhaps we can slide this in before we cut 6.x @retrohacker?

hekike:
Apologize I didn't explain very well the context of this pr.
To be on the same page, the current PR:

- Allows only one slash at the middle to prevent route pram parsing issues
- Allows **zero or more** slashes at the end

I believe what @DonutEspresso proposes is a one step further and wouldn't allow, more than one slash at the end neither.

The new behavior after this proposal would be:

- Allows only one slash at the middle to prevent route pram parsing issues
- Allows **zero or one** slash at the end

retrohacker:
Thanks for tackling this @hekike! 

I'm :+1: on leaving the path exactly as is, and allowing the user to specify the dedupe behavior through a plugin. My main motivation is that we should do as little "magic" as possible, and changing paths before they get to the user seems like magic. I think our plugin logic should behave like [`merge_slashes` in nginx](http://nginx.org/en/docs/http/ngx_http_core_module.html#merge_slashes). If users need a different behavior, they can specify that as their own plugin.

retrohacker:
So, tl;dr, I propose:

1. We remove all url scrubbing logic from the router
2. We update our dedupe plugin to match NGinx's behavior
3. We document how you can specify your own url transformer through `server.pre`.

retrohacker:
As for holding v6, I'd suggest we go ahead and ship. We've been in a state of perpetually staged breaking changes before (as @yunong pointed out yesterday) and it wasn't a great place to be.

Instead, I'd suggest we create a new branch called `next` that we merge `BREAKING CHANGE`s into until we think we have enough for a new major release. Then we can cut from there. This gives us two advantages:
1. We don't have to block bug/feature fixes because `BREAKING` is on the current `master` (like we are doing now)
2. Users who need the new `BREAKING` changes can opt into tracking the `next` branch by using `restify/node-restify#next` in their `package.json`.

hekike:
@retrohacker I removed the double slash merging magic from the router, see my latest commit.
The `restify.plugins.pre.dedupeSlashes()` plugin is already documented.

I also refactored the `dedupeSlashes` test to cover cases with more than two slashes.
If I understand correctly `restify`'s dedupleSlashes plugin already worked like `nginx`'s `merge_slashes` option, just the test didn't point it out earlier:

> Enables or disables compression of two or more adjacent slashes in a URI into a single slash. - [merge_slashes](http://nginx.org/en/docs/http/ngx_http_core_module.html#merge_slashes)

lib/router.js
```diff
@@ -103,7 +103,7 @@ function compileURL(options) {
             return (false);
         }
 
-        pattern += '\\/+';
+        pattern += '\\/{1}';
```
retrohacker: I think this quantifier is unnecessary

lib/router.js
```diff
@@ -134,7 +134,7 @@ function compileURL(options) {
     }
 
     if (!options.strict) {
-        pattern += '[\\/]*';
+        pattern += '[\\/]?';
```
retrohacker: I think this list is unnecessary

retrohacker:
This shouldn't merge into `master` unless we want to cut a new release since it is going to be a breaking change. Can we open this PR against a new branch `next`?

hekike:
@retrohacker I've no permission to create a `next` branch in this repo, could you create one? I cannot open a PR to a non-existent branch.

hekike:
Closed by the favor of https://github.com/restify/node-restify/pull/1512
cc @retrohacker

PR #1482, opened 2017-09-10
Do not use shrinkwrap file to run nsp check

hekike:
# Issues

The current master is broken by shrinkwrap's "extraneous" issues.

TravisCI output:

```
npm ERR! Linux 4.11.6-041106-generic
npm ERR! argv "/home/travis/.nvm/versions/node/v4.8.4/bin/node" "/home/travis/.nvm/versions/node/v4.8.4/bin/npm" "shrinkwrap" "--dev"
npm ERR! node v4.8.4
npm ERR! npm  v2.15.11
npm ERR! Problems were encountered
npm ERR! Please correct and try again.
npm ERR! extraneous: abbrev@1.1.0 /home/travis/build/restify/node-restify/node_modules/nsp/node_modules/abbrev
```

# Changes

Do not use `shrinkwrap` file to run `nsp check`.

Shrinkwrap file is only required with `--offline` flag, which is not recommended anymore, see: https://github.com/nodesecurity/nsp/blob/56f9848b6950ea821e4b92d1f1fb66ffc93bd79e/README.md#offline-mode

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,9 +31,8 @@ NPM		:= npm
 #
 DOC_FILES	 = index.restdown
 JS_FILES	 = '.'
-SHRINKWRAP	 = npm-shrinkwrap.json
 
-CLEAN_FILES	+= node_modules $(SHRINKWRAP) cscope.files
+CLEAN_FILES	+= node_modules cscope.files
 
 include ./tools/mk/Makefile.defs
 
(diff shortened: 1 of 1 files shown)


DonutEspresso:
We weren't using offline mode, the purpose of using shrinkwrap in this case is to verify that the full transitive dependency tree is not causing problems either. 

The problem with extraneous deps I believe is actually a nsp problem:
https://github.com/nodesecurity/nsp/issues/185

We should probably revert this change once that nsp fix lands.

hekike:
@DonutEspresso I think you are right. I only checked the docs of `nsp`. 
If I understand it correctly `nsp` can also [work from `package-lock.json`](https://github.com/nodesecurity/nsp/blob/master/lib/check.js#L130),  shouldn't we move to that direction instead of generating the shinkwrap on CI?

DonutEspresso:
@hekike yes! I think now that npm5 has been out for some time we should move in that direction. I don't think the repo currently has a lockfile checked in, but if nsp supports it then we should absolutely do that.

Issue #853, opened 2015-06-29
Http2 support


yunong:

AartiKriplani:
Is there an ETA for this?

Maxwell2022:
+1

nicolaspeixoto:
+1

sean3z:
+1

NitroGhost:
+1

FabianHinz:
Yes, please! +1

NicolaOrritos:
+1, multiple times :-)

willin:
+1

sebs:
+1 ;)

jscMR:
+1 :)

NagarajNachiappan:
+1

cusspvz:
Isn't HTTP2 already supported by spdy?

sebs:
![mooh](http://bestanimations.com/Signs&Shapes/Hearts/heart-rainbow-animation21.gif)

diaakasem:
+1

mingard:
+1

Atsman:
+1

rozaydin:
+1

retrohacker:
https://github.com/nodejs/http2/blob/master/doc/api/http2.md

industral:
https://github.com/restify/node-restify/blob/master/examples/http2/http2.js

PR #1489, opened 2017-09-12
feat(http2): add HTTP/2 support

hekike:
**DO NOT MERGE IT!**

## Issues

Related:

* https://github.com/restify/node-restify/issues/853

Add HTTP/2 support to `restify`

# Changes

* Adds HTTP/2 support to `server`
* Uses patch method to decorate `Request` and `Response` objects
* Example for HTTP/2 server with `restify`

# TODO

- [] New Node.js release (> 8.4.0) that contains the https://github.com/nodejs/node/pull/14690 change (exposes `Http2ServerRequest` and `Http2ServerRespone` objects)
- [] Tests for HTTP/2 *(probably we should run all of the tests with both `http2` and `spdy` options as well)*

# How to run

At the moment you need to [build Node.js](https://github.com/nodejs/node/blob/master/BUILDING.md) with the latest master and using the `--expose-http2` flag to run this branch.

After that:

```
../node/node --expose-http2 examples/http2/http2.js 
```


retrohacker:
image

yunong:
🎉  I've assigned a few reviewers to look at this, but this is amazing work so far.

hekike:
@DonutEspresso @yunong @jclulow @retrohacker Node.js `v8.8.0` is out which means we have everything that's needed to ship this feature, [see changelog](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V8.md#2017-10-24-version-880-current-mylesborins):

- no more conflict with the `path` getter
- `http2` is now exposed by default without the need for a flag
- However `http2` stability is still experimental in the Node core

I rebased this PR with the `master` and added a simple test case.  
Do you want to wait more or can we ship it?

yunong:
Looks great. What do you think about adding some documentation to the home page about how we support HTTP/2, but that it's still experimental? I think this would give us tremendous PR!

yunong:
What happens if we're on a version of Node that doesn't support `http2`, does the current code handle this path?

hekike:
@yunong I added an assert to the [server.js](https://github.com/restify/node-restify/pull/1489/files#diff-c945a46d13b34fcaff544d966cffcabaR141) `http2` option.

The `http2` related tests will be skipped with older versions:

```
HTTP2 module is not available
Node.js version >= v8.8.8 required, current: 6.9.4

OK: 0 assertions (4ms)
```

hekike:
@yunong regarding docs and landing I can create some hands-on guide for `restify`'s `http2` support, some ideas:

- HTTP/2 Server Push for: like I did here https://blog.risingstack.com/node-js-http-2-push/ *(maybe a new static middleware with push capabilities)*
- HTTP/1 fallback demonstration (we could even add helper functions for protocol detection)
- other?

yunong:
@hekike That blog post is great, maybe we can link to some articles instead of having to maintain HTTP/2 specific documentation.

I was more thinking about just adding docs to say that we support HTTP/2 now, with some pointers to code snippets or examples, which you already have added.

hekike:
@yunong I agree. What do you think, should we link the articles it in this PR?
I believe we could focus on the marketing after the feature is landed.

yunong:
Yeah we should land and then follow up with some docs :)

hekike:
Breaks because of: https://github.com/restify/node-restify/issues/1545

PR #1498, opened 2017-09-18
Correct doc: inflightRequests() returns a number, not a string.

djcsdy:
The attached pull request makes a trivial correction to the inline documentation in restify.

`server.inflightRequests()` was incorrectly documented as returning a `String`. It actually returns a `number`.


retrohacker:
:tada: thanks for contributing this @djcsdy!

PR #1500, opened 2017-09-19
fix(server): error in pre handler triggers after event

retrohacker:
Currently, aborting a middleware chain inside a `pre` handler will cause restify to drop the chain on the floor. This results in the `after` event not firing.

An effect of this is that `after` handlers do not get access to requests rejected by throttles.

This PR fixes that.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -755,7 +755,7 @@ Server.prototype._handle = function _handle(req, res) {
         self._run(req, res, null, self.before, function (err) {
             // Like with regular handlers, if we are provided an error, we
             // should abort the middleware chain and fire after events.
-            if (err !== undefined) {
+            if (err === false || err instanceof Error) {
                 self._finishReqResCycle(req, res, null, err);
                 return;
             }
--- a/lib/server.js
+++ b/lib/server.js
@@ -753,17 +753,14 @@ Server.prototype._handle = function _handle(req, res) {
     // run pre() handlers first before routing and running
     if (self.before.length > 0) {
         self._run(req, res, null, self.before, function (err) {
-            // check for return false here - like with the regular handlers,
-            // if false is returned we already sent a response and should stop
-            // processing.
-            if (err === false) {
-                self._finishReqResCycle(req, res);
+            // Like with regular handlers, if we are provided an error, we
+            // should abort the middleware chain and fire after events.
+            if (err !== undefined) {
+                self._finishReqResCycle(req, res, null, err);
                 return;
             }
 
-            if (!err) {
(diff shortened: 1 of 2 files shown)


test/server.test.js
```diff
@@ -2957,3 +2957,22 @@ test('calling next twice should throw', function (t) {
         t.ifError(err);
     });
 });
+
+test('aborting pre should still call after', function (t) {
```
DonutEspresso: Can we put this at/around L2312 to group it with the other pre handler test that is testing `next(false)`? And maybe rename the test to `next(err) should still trigger after event` to be a little more precise since "abort" doesn't really mean much in context of existing restify docs or nomenclature.

lib/server.js
```diff
@@ -753,17 +753,14 @@ Server.prototype._handle = function _handle(req, res) {
...
     if (self.before.length > 0) {
         self._run(req, res, null, self.before, function (err) {
-            // check for return false here - like with the regular handlers,
-            // if false is returned we already sent a response and should stop
-            // processing.
-            if (err === false) {
-                self._finishReqResCycle(req, res);
+            // Like with regular handlers, if we are provided an error, we
+            // should abort the middleware chain and fire after events.
+            if (err !== undefined) {
```
DonutEspresso: This behavior is slightly inconsistent compared to how its implemented in the `use` chain, where we are explicitly checking for `false` or `err`. The result is that calling `return next({})` could trigger an early exit in pre where it wouldn't in use. I don't mind revisiting what nexting with a POJO means, but probably worth tackling in a separate PR?

DonutEspresso:
Awesome - great work tracking this down. 🎉

lib/server.js
retrohacker: Yup, was on the fence about this. The explicitness bit us in the butt here :smile: so I erred on the side of assuming everything was a cause to abort the chain. Though I guess if someone were to do `next(true)` or `next(null)` this would drop the chain which may not be desirable.

lib/server.js
```diff
@@ -753,17 +753,14 @@ Server.prototype._handle = function _handle(req, res) {
...
-            if (err === false) {
-                self._finishReqResCycle(req, res);
+            // Like with regular handlers, if we are provided an error, we
+            // should abort the middleware chain and fire after events.
+            if (err === false || err instanceof Error) {
+                self._finishReqResCycle(req, res, null, err);
                 return;
             }
 
-            if (!err) {
```
yunong: How come this got dropped? I think if you remove this there is the edge case where the "err" object is not false or a function?

PR #1501, opened 2017-09-19
feat: genericThrottle

retrohacker:
# Changes

This introduces a genericThrottle plugin that allows you to throttle traffic based on any arbitrary metric. It also converts cpuUsageThrottle to being a simple wrapper around the genericThrottle.


retrohacker:
Bump @yunong

PR #1503, opened 2017-09-24
fix(formatter): text formatter crash when body is null/undefined

AVVS:
note: String() will invoke .toString() in most cases and wont crash when .toString() is not defined for some reason


lib/formatters/text.js
```diff
@@ -17,9 +17,7 @@
 function formatText(req, res, body) {
 
     // if body is null, default to empty string
-    var data = '';
-
-    data = body.toString();
+    var data = body == null ? '' : String(body);
```
retrohacker: Great catch!

I like the `== null` style myself for both `null` and `undefined`, but it fails linting in this project.

Could you provide a set of tests for this change? Would like to see
1) A test for when `.toString` is not defined
2) A test for when body is no provided
3) A test for when body is a string

retrohacker:
Marking this as help wanted for implementing tests

PR #1504, opened 2017-09-26
fix: dont include interval in lag

retrohacker:
## Pre-Submission Checklist

# Changes

> What does this PR do?

Currently the `lag` metric for `cpuUsageThrottle` includes the interval. For example, an interval of `50` firing at the exact right moment will result in a lag of `50` as opposed to `0`. This fixes that.

Also, this PR removes all timeouts after the `cpuUsageThrottle` tests.

### DIFF
--- a/test/plugins/cpuUsageThrottle.test.js
+++ b/test/plugins/cpuUsageThrottle.test.js
@@ -19,6 +19,8 @@ var cpuUsageThrottle = proxyquire('../../lib/plugins/cpuUsageThrottle.js', {
 var MR = Math.random;
 describe('cpuUsageThrottle', function () {
 
+    var plugin = undefined
+
     before('Setup: stub math.random', function (done) {
         Math.random = function () {
             return 0;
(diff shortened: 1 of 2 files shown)


retrohacker:
Hmm, make prepush works for me locally. Removing arrow function.

test/plugins/cpuUsageThrottle.test.js
```diff
@@ -30,7 +30,7 @@ describe('cpuUsageThrottle', function () {
         var opts = { limit: 0, interval: 500 };
         var plugin = cpuUsageThrottle(opts);
         function next (cont) {
-            clearTimeout(plugin._timeout);
+            plugin.close();
```
DonutEspresso: Should this just be in an `afterEach` block?

retrohacker: Is there a way to do that without maintaining state outside the test?

DonutEspresso: Not that I know of - usually just use a `plugin` var within the closure.

PR #1506, opened 2017-09-29
docs(README): remove NSP badge

hekike:
The current `nsp` badge is static, it's always green.
However, it's probably correct at the moment when a new PR is merged to master, as CI checks it.
It's not necessarily valid at any point in the time, like now.

# Changes

- remove "always green" `nsp` badge from README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,6 @@
 [![Dependency Status](https://david-dm.org/restify/node-restify.svg)](https://david-dm.org/restify/node-restify)
 [![devDependency Status](https://david-dm.org/restify/node-restify/dev-status.svg)](https://david-dm.org/restify/node-restify#info=devDependencies)
 [![bitHound Score](https://www.bithound.io/github/restify/node-restify/badges/score.svg)](https://www.bithound.io/github/restify/node-restify/master)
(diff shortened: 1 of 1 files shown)


DonutEspresso:
Hm, there is a `./tools/nspBadge.js` script that updates the badge whenever we run `make nsp`. Does that work for you?

hekike:
@DonutEspresso yes but it's misleading, as it's calculated when CI runs, and a vulnerability can be discovered later.

Imagine the following scenario:

1. PR is merged, CI generates a green nsp badge
2. Vulnerability is discovered a day later
3. You check restify's Readme -> you think everything is cool

DonutEspresso:
Hm that's true - are there any ways to keep this "live"? Or are we better off not showing it at all? I figured something was better than nothing, but that's probably not great if it's misleading. :(

hekike:
Snyk has a badge, but it showed zero when we had a known vulnerability based on `nsp` :/
https://snyk.io/test/github/restify/node-restify/badge.svg

Would be nice to have something for nsp as well.

retrohacker:
I'm :+1: on removing this, it is very misleading since `nsp`'s database is dynamic and the image is static. There will never be a time where we merge a non-green `nsp` badge.

retrohacker:
Opening another PR to remove the makefile stuffs

PR #1513, opened 2017-10-09
Restify v7

retrohacker:
Branch for staged breaking changes that have not been released.

To use the features in this branch use: `npm install --save restify/node-restify#next`


retrohacker:
This PR should only be used for changes that are ready to be released, this is not a WIP branch.

PR #1519, opened 2017-10-13
new: audit logger plugin now supports custom context objects

yunong:
Add a context option to the audit logger plugin to support arbitrary user added context to the audit log payload.

### DIFF
--- a/docs/api/plugins.md
+++ b/docs/api/plugins.md
@@ -231,6 +231,11 @@ event, e.g., `server.on('after', plugins.auditLogger());`:
 * `auditLogger(options)` - an audit logger for recording all handled requests
   * `options.event` {String} The name of the event, one of `pre`, `routed`, or `after`
   * `options.log` {Object} bunyan logger
+  * `[opts.context]` {Function} The optional context function of signature
+   f(req, res, route, err).  Invoked each time an audit log is generated. This
+   function can return an object that customizes the format of anything off the
+   req, res, route, and err objects. The output of this function will be
+   available on the `context` key in the audit object.
   * `[options.server]` {Object} restify server. if passed in, causes server to
      emit 'auditlog' event after audit logs are flushed
   * `[options.printLog]` {Boolean} when true, prints audit logs. defaults to true.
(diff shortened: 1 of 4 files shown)


lib/plugins/audit.js
```diff
@@ -41,8 +41,13 @@ function getResponseHeaders(res) {
  * @param {Object} opts.log The logger.
  * @param {String} opts.event The event from the server which initiates the
  * log, one of 'pre', 'routed', or 'after'
+ * @param {Context} [opts.context] The optional context function of signature
```
DonutEspresso: The JSDoc here looks incorrect, should be {type} not the name of the parameter?

lib/plugins/audit.js
```diff
@@ -41,8 +41,13 @@ function getResponseHeaders(res) {
...
+ * f(req, res, route, err).  Invoked each time an audit log is generated. This
+ * function can return an object that customizes the format of anything off the
+ * req, res, route, and err objects. The output of this function will be
+ * available on the `context` key in the audit object.
  * @param {Server} [opts.server] The restify server, used to emit the audit log
- * object programttically
+ * object programtically
```
DonutEspresso: programmatically

docs/api/plugins.md
```diff
@@ -231,6 +231,11 @@ event, e.g., `server.on('after', plugins.auditLogger());`:
 * `auditLogger(options)` - an audit logger for recording all handled requests
   * `options.event` {String} The name of the event, one of `pre`, `routed`, or `after`
   * `options.log` {Object} bunyan logger
+  * `[opts.context]` {Context} The optional context function of signature
```
DonutEspresso: {Object}

lib/plugins/audit.js
```diff
@@ -41,8 +41,13 @@ function getResponseHeaders(res) {
  * @param {Object} opts.log The logger.
  * @param {String} opts.event The event from the server which initiates the
  * log, one of 'pre', 'routed', or 'after'
+ * @param {Context} [opts.context] The optional context function of signature
+ * f(req, res, route, err).  Invoked each time an audit log is generated. This
+ * function can return an object that customizes the format of anything off the
+ * req, res, route, and err objects. The output of this function will be
+ * available on the `context` key in the audit object.
  * @param {Server} [opts.server] The restify server, used to emit the audit log
```
DonutEspresso: Not technically your commit but since you're here {Object} not {Server}

PR #1522, opened 2017-10-18
Remove unleash dependency

### LINKED ISSUE #1521: unleash should be moved from dependencies to devDependencies
- [x] Used appropriate template for the issue type
- [x] Searched both open and closed issues for duplicates of this issue
- [x] Title adequately and _concisely_ reflects the feature or the bug

# Bug Report

## Restify Version

6.2.0

## Node.js Version

8.7.0

## Expected behaviour

`npm install restify` should not install unleash and its dependencies.

## Actual behaviour

`npm install restify` installs unleash and its dependencies.

## Repro case

`npm install restify`

## Cause

https://github.com/restify/node-restify/blob/09066d833c5431b2f4a33758a1856a953aae57be/package.json#L113

Yes

boennemann:
## Pre-Submission Checklist

- [x] Closes #1521
- [ ] ~Included comprehensive and convincing tests for changes~

# Changes

With the release of 6.2.0 (https://github.com/restify/node-restify/commit/09066d833c5431b2f4a33758a1856a953aae57be) the unleash package got added as a dependency of this project. I suppose this happened accidentally, as it's not a real runtime dependency of this project.

Aside from downloading unnecessary bytes this has a very unfortunate consequence.
With the unleash package both "npm-check" and "npm" end up in the users `node_modules/.bin` folder.

Now if you run `npx` inside a project that uses restify, or `npx npm-check -u` for that matter, you end up with `unleash`'s outdated npm-check version, and even worse with an npm@2.

This lead to verify confusing and seemingly unrelated error messages – until I could track it down to this slip-up in restify.

I hope my changes are in the right format, so you can merge this PR directly and release a new version.

I also removed `lodash` from devDependencies, because it's a dependency already. This gets rid of an npm warning.

Thanks a lot!


EvgenyOrekhov:
I would call this a fix because unleash and npm@2 broke all my apps. See more info here: https://github.com/npm/npm/issues/18874.

retrohacker:
Hey all! Sorry for this, looks like npm's new "default to --save" behavior caught one of our contributors by surprise. Thank you for catching this and filing an issue so quickly @boennemann! :heart: sorry we broke your builds!

Also looks like we failed to semver bump our npm-shrinkwrap file, thank you for doing this in this PR.

boennemann:
@EvgenyOrekhov You're right. Updated the commit message.

@retrohacker Don't worry, mistakes happen :) In order to keep them from happening again I added another commit. It changes the unleash instructions to use npx. That way nobody has to install unleash explicitly.

retrohacker:
TIL: `npx` is a thing! That is awesome!

boennemann:
🎉

Issue #1315, opened 2017-04-26
Convert JSCS to eslint


retrohacker:
JSCS [merged with eslint](https://medium.com/@markelog/jscs-end-of-the-line-bc9bf0b3fdb2#.x0o1xqkk9) a little while ago. It may make sense to migrate our JSCS config over to eslint since we are using both projects.

https://www.npmjs.com/package/polyjuice/ can automate this for us

yunong:
@retrohacker Thanks for the tip. Since the projects are merging, it definitely makes sense to migrate. However, I would propose that we explicitly separate the style configs from the lint configs. It makes it cleaner and more obvious.

My recommendation would be that we keep both of our `make` tasks of `style` and  `lint`, and either have two `.eslintrc` files, or separate sections in the `.eslintrc` file for style and lint.

retrohacker:
Sounds good to me!

hekike:
@retrohacker @yunong do you still want to keep both `jscs` and `eslint`?

Currently, I'm facing the following issue with JSCS: https://github.com/jscs-dev/node-jscs/issues/1939 while I'm working on JSDoc.

I'm wondering if it's still necessary to have both in the project. ESLint handles spaces in JSDoc perfectly.

Also `jscs` is not maintained anymore and is merged with `eslint` see https://medium.com/@markelog/jscs-end-of-the-line-bc9bf0b3fdb2#.x0o1xqkk9

hekike:
Closed by: https://github.com/restify/node-restify/pull/1527

PR #1527, opened 2017-10-18
Remove JSCS as it has merged with ESLint

hekike:
## Pre-Submission Checklist

## Issues

Closes:

- https://github.com/restify/node-restify/issues/1315

JSCS has merged with ESLint: 
https://medium.com/@markelog/jscs-end-of-the-line-bc9bf0b3fdb2

# Changes

- Remove JSCS
- Migrate JSCS rules to ESLint
- Enable uncommented stricter ESLint rules
- Separate make command for linting and styling
- Auto-fix indentation issues
- Fix `self` assignment issues (`self` "keyword" was used for other than `this`)


retrohacker:
Yes! Please! :heart:

hekike:
@yunong if we separate the two things into multiple `.eslintrc` files the IDE cannot make live linting based on the `.eslintrc`. Which would be a huge comfort killer.

What would be the benefit of keeping them separate?

hekike:
For the record: 
We discussed the benefits of having a separate `lint` and `style` commands and come up with a solution that is based on the `.eslintrc.js` file and uses environment variables to disable a specific group of rules in certain cases. This concept also allows us to have one ESLint config file that works well with most of the popular IDEs.

"Unfortunately" ESLint's indentation checker is much stricter that caused a massive and evil (really: `+666 −666`) line changes. I moved it to a separate commit.

I think this PR is ready for merging in the case you agree with me.

cc @yunong @retrohacker

retrohacker:
> I'm generally happy do this, however I'd like to still keep the .eslintrc files for style and lint separate 

> We discussed the benefits of having a separate lint and style 

Can I get some more context on this @yunong? What is the value add?

retrohacker:
I'm :+1: on merging this either way since it unblocks @hekike's work on other PRs, and I would just like to see JSCS leave the codebase.

I'd like to at least revisit the decision to differentiate between linting and style rules via custom tooling (is there a distinction? Are there rules that fit both? Is the cognitive overhead of separating them worth it? What value do we loose by combining them? What does supporting this custom tooling moving forward look like, and is that worth it vs. a plain old .eslintrc file? Does this increase a barrier to entry for non-core members contributing linting rules? Does this translate easily to other projects we maintain?) but we can move that to another Issue/PR if @yunong is :+1: on landing this.

lib/plugins/cpuUsageThrottle.js
```diff
@@ -86,39 +86,41 @@ function cpuUsageThrottlePlugin (opts) {
...
     // current point in time based on how much over our limit we are. This is
     // updated on an interval by updateReject().
-    self._reject = 0;
+    plugin._reject = 0;
 
-    // self._timeout keeps track of the current handle for the setTimeout we
+    // plugin._timeout keeps track of the current handle for the setTimeout we
     // use to gather CPU load averages, this allows us to cancel the timeout
     // when shutting down restify.
-    self._timeout = null;
```
yunong: Why is it changing all the `self`s to `plugin`?

tools/mk/Makefile.targ
```diff
@@ -122,16 +122,12 @@ check-bash: $(BASH_FILES:%=%.bashchk) $(BASH_FILES:%=%.bashstyle)
...
-check-jscs::
-	$(JSCS) $(JS_FILES)
-
 .PHONY: check-eslint
 check-eslint::
 	$(ESLINT) $(JS_FILES)
 
 .PHONY: check
-check: check-eslint check-jscs check-bash
+check: check-eslint check-bash
```
yunong: Did you get a chance to add back the style target?

lib/plugins/cpuUsageThrottle.js
hekike: `consistent-this` rule forces to use `self` for `this`

hekike: Self is not the `this` here.

tools/mk/Makefile.targ
hekike: Without the environment variables, it runs all of the rules both lint and style. Do you want to separate hem in the `Makefile.targ` like:  `check-lint` and `check-style`? I would keep `check-eslint` for the default behavior.

hekike: Probably I should rename these: https://github.com/restify/node-restify/pull/1527/files#diff-b67911656ef5d18c4ae36cb6741b7965R52

hekike: Could be something like this:

```
.PHONY: check-eslint
check-eslint::
	$(ESLINT) $(JS_FILES)

.PHONY: check-lint
check-lint::
	NO_STYLE=true $(ESLINT) $(JS_FILES)

.PHONY: check-style
check-style::
	NO_LINT=true $(ESLINT) $(JS_FILES)

.PHONY: check
check: check-lint check-style check-bash
	@echo check ok
```

yunong: Yes, if we can keep two separate targets it'll keep the makefile as it was, but all on eslint which would be awesome.

hekike: Fixed

retrohacker:
Leaving a comment here for future Will with the context we talked about offline:

We have two classes of rules that we would like to keep separated: correctness and style. We have an opinionated set of rules that we believe help make a Node.js application "correct" by removing certain classes of errors. These rules fall under "linting." We then have an opinionated set of styles like indentation that don't necessarily improve the correctness of the code but help improve it's readability.

By keeping these separate, we can easily copy the "correctness" rules to our other projects that we maintain while allowing flexibility for different styles in those code bases.

Keeping this delineation right now didn't introduce significant complexity. If we find this breaks forwards compatibility in the future, we may revisit this decision.

PR #1531, opened 2017-10-18
chore(docs): remove watch

hekike:
Docs watch seemed to be a good idea but the output is far from the Jekyll output which makes it pretty much useless. Docs watch input configuration also needs a maintainence, which is easy to forget. After spending a couple of hours with writing docs I believe it's totally unnecessary. `docs-build` with the markdown output is perfectly enough.

# Changes

- Remove make `docs-watch`


yunong:
LGTM, feel free to merge once tests pass as you have the commit bits now :)

PR #1535, opened 2017-10-21
chore(eslint): ESLint JSDoc rules, JSDoc improvements

hekike:
Add ESLint rules for JSDoc and improve JSDoc.

# Changes

- Add ESLint rules for JSDoc
- Improve JSDoc

PR #1537, opened 2017-10-21
chore(prettier): use prettier for style checking and fixing

hekike:
Use prettier for style checking and fixing instead of ESLint rules.
Prettier provides an opinionated style with an auto-fix capability for existing codebase.

> Prettier enforces a consistent code style (i.e. code formatting that won't affect the AST) across your entire codebase because it disregards the original styling* by parsing it away and re-printing the parsed AST with its own rules that take the maximum line length into account, wrapping code when necessary. - https://github.com/prettier/prettier

# Changes

- [x] integrate `prettier`
- [x] replace ESLint style related rules with `prettier`
- [x] run `fix-style` (merge https://github.com/restify/node-restify/pull/1535 first)

# Makefile

- `make check-lint` uses ESLint with manually defined linting rules
- `make check-style` uses ESLint with `prettier` rules via npm package
- `make fix-style` uses `prettier` to re-print the code


DonutEspresso:
👍  from here, assuming we go into this using the existing styles and rules?

yunong:
This looks good. I'm down to use an automagic style formatter as long as it uses semicolons and the column width is 80 as God intended.

hekike:
@DonutEspresso @yunong @retrohacker I finalized this PR.
I would recommend reviewing only the first commit as the second one's content is auto-generated by `prettier`.

Most of the changes come from the different spacing around `function` keyword, sadly it's not configurable as prettier is an opinionated pre-configured formatter.

yunong:
I'm fine with this. What does everyone else think? @rajatkumar @micahr @retrohacker

DonutEspresso:
LGTM!

package.json
```diff
@@ -124,13 +124,16 @@
     "documentation": "5.3.2",
     "eslint": "4.9.0",
     "eslint-plugin-jsdoc": "3.1.3",
+    "eslint-config-prettier": "2.6.0",
+    "eslint-plugin-prettier": "2.3.1",
```
DonutEspresso: What are these two deps for? Also any reason we aren't floating on these?

hekike: Sorry, it's my bad, my global `.npmrc` dropped id.

`eslint-config-prettier` and `eslint-plugin-prettier` integrate ESLint with `prettier`, so ESLint can detect rules against prettier and IDEs can recognize it.

Read more about it: https://prettier.io/docs/en/usage.html#eslint

hekike: Okay, I just realized that with this I killed the separate linting and style-checking, this PR needs a little bit more work.

hekike: I fixed it, the new version uses `ESLint` with the environment variable trick to separate linting and style checking. The `make check-style` uses ESLint with `prettier` rules. The `make fix-style` uses `prettier` to re-print the code.

retrohacker:
Rebased

Issue #1545, opened 2017-11-01
TravisCI breaks with EADDRNOTAVAIL


hekike:
# Bug Report

`master` tests pass on a local machine but break on TravisCI:

**server.test.js / gh-751 IPv4/IPv6 server URL**

```
Error: listen EADDRNOTAVAIL ::1:36480
    at Object.exports._errnoException (util.js:1020:11)
    at exports._exceptionWithHostPort (util.js:1043:20)
    at Server._listen2 (net.js:1249:19)
    at listen (net.js:1298:10)
    at net.js:1408:9
    at _combinedTickCallback (internal/process/next_tick.js:83:11)
    at process._tickDomainCallback (internal/process/next_tick.js:128:9)
```

## Node.js Version

all

## Repro case

See: https://travis-ci.org/restify/node-restify/jobs/295430142

## Are you willing and able to fix this?

Yes, but I need a hint

hekike:
Seems to be that the current version of TravisCI lacks the IPv6 support:

> The virtual machines in the Legacy environment running the tests have IPv6 enabled. They do not have any external IPv4 address but are fully able to communicate with any external IPv4 service. The container-based, OS X, and GCE (both Precise and Trusty) builds do not currently have IPv6 connectivity. - https://docs.travis-ci.com/user/reference/precise/#Networking

I'm not sure why we have this issue now.

PR #1549, opened 2017-11-02
test(server): fix ERR_SERVER_ALREADY_LISTEN issue with Node v9.0.0

hekike:
## Pre-Submission Checklist

Node v9.0.0 doesn't allow to call server.listen() multiple times.
(It's not clear why it was necessary earlier)

See: https://nodejs.org/api/errors.html#errors_err_server_already_listen

# Changes

- remove duplicated server.listen() calls from tests

### DIFF
--- a/test/plugins/query.test.js
+++ b/test/plugins/query.test.js
@@ -225,12 +225,10 @@ describe('query parser', function() {
             res.send(req.params);
         });
 
-        SERVER.listen(8080, function() {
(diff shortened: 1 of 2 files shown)


hekike:
It won't pass on TravisCI because of the IPv6 issue: https://github.com/restify/node-restify/issues/1545

I tested it on my local machine.

PR #1550, opened 2017-11-02
Update minor dependencies

hekike:
Update minor  and non-semver dependencies.

# Changes

- update minor dependencies
- update non-semver dependencies (nodeunit)


hekike:
Travis broke after merge with styling issues, some linter dependency didn't follow SemVer :(
Here is the fix: https://github.com/restify/node-restify/pull/1559

PR #1553, opened 2017-11-03
BREAKING CHANGE: dropping Node.js support < v4.0.0

hekike:
Drop Node.js support below `v4.0.0`.
Revisit TravisCI Node.js versions, only keep the supported releases: 

- Maintenance LTS (4.x)
- Current LTS
- Latest release

# Changes

- package.json engines
- TravisCI config

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,15 @@
 sudo: false
 language: node_js
 node_js:
-   - "4"
-   - "6"
-   - "7"
-   - "8"
-   - "stable"
+   - "4"                    # Maintenance LTS release
+   - "lts/*"                # Active LTS release
+   - "node"                 # Latest stable release
 env:
-  - TEST_SKIP_IP_V6=true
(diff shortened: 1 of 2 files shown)


hekike:
Jobs are broken on 0.10 because the lack of ES6 support that's used by our dependencies.

We haven't run tests on `< 4.0.0` since December 2016 (see: https://github.com/restify/node-restify/commit/2f0f1771e10483ef6dc0326c232ce55ce371aeb8#diff-354f30a63fb0907d4ad57269548329e3).

Shouldn't we drop < 4.0.0 support from `restify` if we don't test it continuously?

DonutEspresso:
Agreed - moving forward we should feel comfortable following Node's LTS schedule - if it's outside LTS I don't think we have the bandwidth to support it.

hekike:
@DonutEspresso I changed the PR to drop support < `v4.0.0` as `4.x` this is the current LTS Maintenance release.
@retrohacker should it go to the `next` branch? It's a breaking change.

retrohacker:
Woah, looks like next needs to be rebased :joy:

PR #1554, opened 2017-11-03
Performance improvement: benchmark suite and accept content-type optimization

hekike:
Benchmark and improve performance.

# Changes

- Add benchmark suite based on https://github.com/fastify/benchmarks
- Set Content-Type to `application/json` when `res.send` is called with an Object instead of calling `res.json`
- Call `req.accepts` only once when not provided type is derived
- Eliminate some closures
- Optimize internal set header calls: use raw `res.setHeader`

# Improvement

Dumb servers:

```sh
$ npm start

> restify-benchmark@0.0.0 start /node-restify/benchmark
> node .

? Do you want to compare HEAD with latest release? Yes
? How many connection you need? 100
? How many pipelining you need? 10
? How long does it takes? 30
---- response-json ----
response-json throughput:
{
    "significant": "***",
    "equal": false,
    "wins": "head",
    "diff": "21.3%"
}

---- response-text ----
response-text throughput:
{
    "significant": "***",
    "equal": false,
    "wins": "head",
    "diff": "7.3%"
}

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ nbproject
 deps/javascriptlint
 deps/jsstyle
 package-lock.json
+benchmark/results
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,6 +42,12 @@ make docs-build
 
 To add a new page, simply give it a [permalink](https://github.com/restify/node-restify/blob/94fe715173ffcebd8814bed7e17a22a24fac4ae8/docs/index.md) and then update [docs.yml](https://github.com/restify/restify.github.io/blob/master/_data/docs.yml) with the new permalink.
 
+## Running a benchmark
+
+```
+make benchmark
+```
+
 ## Cutting a release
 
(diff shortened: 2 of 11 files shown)


benchmark/README.md
```diff
@@ -0,0 +1,5 @@
+# restify-benchmark
+
+```sh
+npm start
```
yunong: Can this be a make task instead? I'd rather not split our build tasks between npm tasks and makefiles.

hekike: Yes, the current benchmark is living in a separate folder: it has it's own npm package file as well. I can move these dependencies to the main devDependencies and create a make command.

What do you think? Should I remove the separate `package.json`?

yunong: Yeah probably makes sense to make those a devDependency in the main `package.json`

hekike: Done, I merged it into the main package.json, also added make command and mentioned in the `CONTRIBUTION.md`

PR #1561, opened 2017-11-09
7.x: new router and middleware system

hekike:
Replace the router and middleware logic with a more performant solution.
Goal: zero or almost zero API/functionality change with significant performance improvement.

Pt 1. (https://github.com/restify/node-restify/pull/1554) was nice for dumb servers and synthetic tests; this should be more useful for real-world applications, especially ones with many routes.

# Changes

- Replaces the router with the Radix Tree based [find-my-way](https://github.com/delvedor/find-my-way)
- New middleware system called `chain.js`, similar to [connect](https://github.com/senchalabs/connect)
- Move versioning and contentType routing into a `conditionalHandler` plugin, see:

```js
server.use(restify.plugins.conditionalHandler({
   contentType: 'application/json',
   version: '1.0.0'
   handler: function (req, res, next) {
       next();
   })
});

server.get('/hello/:name', restify.plugins.conditionalHandler([
  {
     version: '1.0.0',
     handler: function(req, res, next) { res.send('1.x') }
  },
  {
     version: ['1.5.0', '2.0.0'],
     handler: function(req, res, next) { res.send('1.5.x, 2.x') }
  },
  {
     version: '3.0.0',
     contentType: ['text/html', 'text/html']
     handler: function(req, res, next) { res.send('3.x, text') }
  },
  {
     version: '3.0.0',
     contentType: 'application/json'
     handler: function(req, res, next) { res.send('3.x, json') }
  }
]);
```

Hints for review:

- `response.js`: 100% new, do not review it by diff
- `server.js`: the complete request handling logic is new, I recommend to review it as a new file

# TODO

- [x] replacing router
- [x] replacing middleware system
- [x] content-type based routing *(moved to `conditioalHandler` plugin)*
- [x] router versioning, see: https://github.com/delvedor/find-my-way/issues/39 *(moved to `conditioalHandler` plugin)*
- [x] fix plugin tests *(114 passing,  39 failing)*
- [x] ~~revisit fix response tests port *(why did it fail with absolut url?)*~~
- [x] add back DTrace support to server.js
- [x] `router` tests
- [x] `chain` tests
- [x] re-visit `once` for handler `next`
- [x] router unmount, see: https://github.com/delvedor/find-my-way/issues/40 *(tests are disabled temporary)*
- [x] Write migration guide

# Breaking Changes

- Nonstrict-routing is not available
- Limited `RegExp` usage in router path see https://github.com/delvedor/find-my-way
- Remove already deprecated `next.ifError`
- disable DTrace probes by default *( I know it's painful, but most of the people don't know/use it, it causes ~10% throughput improvement in all benchmark tests)*, `createServer({ dtrace: true })`
- Router versioning and contentType routing moved to `conditioalHandler` plugin, see above
- `once ` is not applied automatically for multiple `next()` calls, use the new `onceNext: true` server option, `strictNext` is the same

# Results

Test machine: Mid 2014 MBP, 2.6 i5 16GB RAM 
(~30% improvement compared to `6.3.2` (https://github.com/restify/node-restify/pull/1554), ~50% compared to `6.3.1`)

## Result: dumb server

*Note: Most of the performance improvements in these benchmarks come from the new middleware logic and other optimizations, the router performance is not tested here as it uses LRU cache.*

```
---- response-json ----
response-json throughput:
{
    "significant": "***",
    "equal": false,
    "wins": "head",
    "diff": "30.21%"
}

---- response-text ----
response-text throughput:
{
    "significant": "***",
    "equal": false,
    "wins": "head",
    "diff": "34.18%"
}
```

## Result: middleware

```
---- middleware ----
middleware throughput:
{
    "significant": "***",
    "equal": false,
    "wins": "head",
    "diff": "44.68%"
}
```

## Result: heavy router

*Note: Disabling (mocking out) cache in a stable version: it's not fair as it aims to the worst case, cache hit ratio is 0%. However, it's still better than the worst case scenario as it least doesn't require extra time to maintain the LRU cache.
There is no other way to simulate 100+ different endpoint calls with the current benchmark suite.*

```
---- router-heavy ----
router-heavy throughput:
{
    "significant": "***",
    "equal": false,
    "wins": "head",
    "diff": "3086.8%"
}
```

### DIFF
--- a/benchmark/benchmarks/router-heavy.js
+++ b/benchmark/benchmarks/router-heavy.js
@@ -1,0 +1,132 @@
+'use strict';
+
+var restify = process.argv.includes('version=head')
+    ? require('../../lib')
+    : require('restify');
+
+var server = restify.createServer();
+var path = '/whiskeys/scotch/islay/lagavulin/16-years/50';
+var methods = ['post', 'put', 'get', 'del', 'patch'];
+var _ = require('lodash');
+var port = 3000;
+
+// Disabling cache: it's not fair as it aims to the worst case, when
+// cache hit ratio is 0%. However, it's still better than the worst
+// as it doesn't require extra time to maintain the LRU cache.
+// There is no other way to simulate 100+ different endpoint
+// calls with the current benchmark suite.
+if (!process.argv.includes('version=head')) {
+    server.router.cache = {
+        get: function get() {
+            return null;
+        },
+        set: function get() {
+            return null;
+        },
+        dump: function get() {
+            return [];
+        }
+    };
+}
+
+module.exports = {
+    url: 'http://localhost:' + port + path
+};
+
+var routes = {
+    beers: {
+        ale: {
+            'pale-ale': {
+                'american-pale-ale': [],
+                'indian-pale-ale': []
+            },
+            lambic: [],
+            stout: {
+                'american-porter': [],
+                'imperial-stout': [],
+                'irish-stout': []
+            }
+        },
+        lager: {
+            'german-lager': {
+                marzen: []
+            },
+            pilsner: {
+                'german-pilsner': []
+            }
+        }
+    },
+
+    whiskeys: {
+        american: {
+            bourbon: {
+                kentchuky: {
+                    'jim-beam': ['jim-beam', 'bookers', 'old-crow'],
+                    'makers-mark': ['makers-mark'],
+                    'woodford-reserve': ['woodford-reserve']
+                },
+                tennessee: {
+                    'jack-daniels': ['jack-daniels']
+                }
+            },
+            rye: {
+                'beam-suntory': ['jim-beam-rye', 'knob-creek']
+            }
+        },
+        irish: {
+            'single-malt': {
+                bushmills: ['bushmills'],
+                connemare: ['connemare']
+            },
+            'single-pot': {
+                redbreast: ['redbreast'],
+                jameson: ['jameson-15-year']
+            }
+        },
+        japanese: {
+            nikka: ['coffeey-malt', 'blended', 'from-the-barrel'],
+            hibiki: ['japanese-harmony'],
+            yamazakura: ['blended']
+        },
+        scotch: {
+            islay: {
+                bruichladdich: ['25-years', 'islay-barley-2009'],
+                octomore: ['7.2', 'islay-barley-8.3'],
+                laphroaig: ['lore', '15-years', 'four-oak'],
+                lagavulin: ['distillers-edition', '8-years', '16-years']
+            }
+        }
+    }
+};
+
+function handler(req, res) {
+    res.send('hello');
+}
+
+function attachRoute(parent, routeConfig) {
+    _.map(routeConfig, function map(route, routeKey) {
+        var pathChunk = _.isString(routeKey) ? routeKey : route;
+        var routePath = parent + '/' + pathChunk;
+
+        methods.forEach(function forEach(method) {
+            server[method](routePath, handler);
+        });
+
+        if (_.isObject(route) || _.isArray(route)) {
+            attachRoute(routePath, route);
+        }
+        if (_.isString(route)) {
+            for (var i = 0; i <= 100; i++) {
+                methods.forEach(function forEach(method) {
+                    server[method](routePath + '/' + i, handler);
+                });
+            }
+        }
+    });
+}
+
+attachRoute('', routes);
+
+if (!module.parent) {
+    server.listen(port);
+}
--- a/docs/guides/6to7guide.md
+++ b/docs/guides/6to7guide.md
@@ -1,0 +1,157 @@
+---
+title: restify 6.x to 7.x migration guide
+permalink: /docs/6to7/
+---
+
+## Introduction
+
+restify `7.x` comes with a completely new router and middleware logic that
+brings significant performance improvement to your application.
+From `v7.0.0` restify uses the Radix Tree based
+[find-my-way](https://github.com/delvedor/find-my-way) package as a router
+backend.
+
+## Breaking Changes
+
+### Server returns `RequestCloseError` instead of `RequestAbortedError`
+
+Server returns `RequestCloseError` instead of `RequestAbortedError` in the case
+of the request was terminated by the client for some reason.
+
+The new version of restify never returns `RequestAbortedError`.
+
+### Non-strict routing is gone
+
+Option `strictRouting` is removed `createServer({ strictRouting: false })`.
+Strict routing is the new default.
+
+### Different `RegExp` usage in router path and wildcards
+
+restify's new router backend
+[find-my-way](https://github.com/delvedor/find-my-way) has limited RegExp
+support.
+
+#### Guide to define paths
+
+To register a **parametric** path, use the *colon* before the parameter name.
+For **wildcard** use the *star*.
+*Remember that static routes are always inserted before parametric and wildcard.*
+
+```js
+// parametric
+server.get('GET', '/example/:userId', (req, res, next) => {}))
+server.get('GET', '/example/:userId/:secretToken', (req, res, next) => {}))
+
+// wildcard
+server.get('GET', '/example/*', (req, res, next) => {}))
+```
+
+Regular expression routes are supported as well, but pay attention, RegExp are
+very expensive in term of performance!
+
+```js
+// parametric with RegExp
+server.get('GET', '/example/:file(^\\d+).png', () => {}))
+```
+
+RegExp path chunks needs to be between parentheses.
+
+It's possible to define more than one parameter within the same couple of slash
+("/"). Such as:
+
+```js
+server.get('/example/near/:lat-:lng/radius/:r', (req, res, next) => {}))
+```
+
+*Remember in this case to use the dash ("-") as parameters separator.*
+
+Finally it's possible to have multiple parameters with RegExp.
+
+```js
+server.get('/example/at/:hour(^\\d{2})h:minute(^\\d{2})m', (req, res, next) => {
+  // req.params => { hour: 12, minute: 15 }
+}))
+```
+In this case as parameter separator it's possible to use whatever character is
+not matched by the regular expression.
+
+Having a route with multiple parameters may affect negatively the performance,
+so prefer single parameter approach whenever possible, especially on routes
+which are on the hot path of your application.
+
+Fore more info see: https://github.com/delvedor/find-my-way
+
+### Remove already deprecated `next.ifError`
+
+`next.ifError(err)` is not available anymore.
+
+### Disable DTrace probes by default
+
+DTrace probes comes with some performance impact that's fine for the sake of
+observability but you may don't use it at all.
+
+### Change in calling `next` multiple times
+
+Earlier `restify` automatically prevented calling the `next()` more than once.
+In the new version this behaviour is disabled by default, but you can activate
+it with the `onceNext` property.
+
+The behaviour of the `strictNext` option is unchanged.
+Which means `strictNext` enforces `onceNext` option.
+
+```js
+var server = restify.createServer({ onceNext: true })
+server.use(function (req, req, next) {
+    next();
+    next();
+});
+// -> fine
+
+var server = restify.createServer({ strictNext: true })
+server.use(function (req, req, next) {
+    next();
+    next();
+});
+// -> throws an Error
+```
+
+### Router versioning and content type
+
+`accept-version` and `accept` based conditional routing moved to the
+`conditionalHandler` plugin, see docs or example:
+
+```js
+var server = restify.createServer()
+
+server.use(restify.plugins.conditionalHandler({
+   contentType: 'application/json',
+   version: '1.0.0'
+   handler: function (req, res, next) {
+       next();
+   })
+});
+
+server.get('/hello/:name', restify.plugins.conditionalHandler([
+  {
+     version: '1.0.0',
+     handler: function(req, res, next) { res.send('1.x') }
+  },
+  {
+     version: ['1.5.0', '2.0.0'],
+     handler: function(req, res, next) { res.send('1.5.x, 2.x') }
+  },
+  {
+     version: '3.0.0',
+     contentType: ['text/html', 'text/html']
+     handler: function(req, res, next) { res.send('3.x, text') }
+  },
+  {
+     version: '3.0.0',
+     contentType: 'application/json'
+     handler: function(req, res, next) { res.send('3.x, json') }
+  }
+]);
+
+// 'accept-version': '^1.1.0' => 1.5.x, 2.x'
+// 'accept-version': '3.x', accept: 'application/json' => '3.x, json'
+```
--- a/docs/guides/6to7guide.md
+++ b/docs/guides/6to7guide.md
@@ -155,3 +155,14 @@ server.get('/hello/:name', restify.plugins.conditionalHandler([
 // 'accept-version': '^1.1.0' => 1.5.x, 2.x'
 // 'accept-version': '3.x', accept: 'application/json' => '3.x, json'
 ```
+
+### Metrics plugin latency
+
+In 7.x Metrics plugin's `latency` is calculated when the request is
+fully flushed. Earlier it was calculated when the last handler finished.
+
+To address the previous use-cases, new timings were added to the metrics plugin:
+
+ - `metrics.preLatency` pre handlers latency
+ - `metrics.useLatency` use handlers latency
+ - `metrics.routeLatency` route handlers latency
--- a/docs/guides/6to7guide.md
+++ b/docs/guides/6to7guide.md
@@ -156,6 +156,11 @@ server.get('/hello/:name', restify.plugins.conditionalHandler([
 // 'accept-version': '3.x', accept: 'application/json' => '3.x, json'
 ```
 
+### After event fires when both request is flushed and last handler is finished
+
+In 7.x `after` event fires after both request is flushed
+and last handler is finished.
+
 ### Metrics plugin latency
 
 In 7.x Metrics plugin's `latency` is calculated when the request is
--- a/lib/chain.js
+++ b/lib/chain.js
@@ -1,0 +1,171 @@
+'use strict';
+
+var assert = require('assert-plus');
+var once = require('once');
+
+module.exports = Chain;
+
+/**
+ * Create a new middleware chain
+ *
+ * @public
+ * @class Chain
+ * @param {Object} [options] - options
+ * @param {Boolean} [options.onceNext=false] - Prevents calling next multiple
+ *  times
+ * @param {Boolean} [options.strictNext=false] - Throws error when next() is
+ *  called more than once, enables onceNext option
+ * @example
+ * var chain = new Chain();
+ * chain.add(function (req, res, next) { next(); })
+ * // chain.add(function (req, res, next) { next(new Error('Foo')); })
+ * // chain.add(function (req, res, next) { next(false); })
+ *
+ * http.createServer((req, res) => {
+ *    chain.run(req, res, function done(err) {
+ *       res.end(err ? err.message : 'hello world');
+ *    });
+ * })
+ */
+function Chain(options) {
+    assert.optionalObject(options, 'options');
+    options = options || {};
+    assert.optionalBool(options.onceNext, 'options.onceNext');
+    assert.optionalBool(options.strictNext, 'options.strictNext');
+
+    this.onceNext = !!options.onceNext;
+    this.strictNext = !!options.strictNext;
+
+    // strictNext next enforces onceNext
+    if (this.strictNext) {
+        this.onceNext = true;
+    }
+
+    this._stack = [];
+    this._once = this.strictNext === false ? once : once.strict;
+}
+
+/**
+ * Public methods.
+ * @private
+ */
+
+/**
+ * Get handlers of a chain instance
+ *
+ * @memberof Chain
+ * @instance
+ * @returns {Function[]} handlers
+ */
+Chain.prototype.getHandlers = function getHandlers() {
+    return this._stack;
+};
+
+/**
+ * Utilize the given middleware `handler`
+ *
+ * @public
+ * @memberof Chain
+ * @instance
+ * @param {Function} handler - handler
+ * @returns {undefined} no return value
+ */
+Chain.prototype.add = function add(handler) {
+    // _name is assigned in the server and router
+    handler._name = handler._name || handler.name;
+
+    // add the middleware
+    this._stack.push(handler);
+};
+
+/**
+ * Returns the number of handlers
+ *
+ * @public
+ * @memberof Chain
+ * @instance
+ * @returns {Number} number of handlers in the stack
+ */
+Chain.prototype.count = function count() {
+    return this._stack.length;
+};
+
+/**
+ * Handle server requests, punting them down
+ * the middleware stack.
+ *
+ * @public
+ * @memberof Chain
+ * @instance
+ * @param {Request} req - request
+ * @param {Response} res - response
+ * @param {Function} done - final handler
+ * @returns {undefined} no return value
+ */
+Chain.prototype.run = function run(req, res, done) {
+    var self = this;
+    var index = 0;
+
+    function next(err) {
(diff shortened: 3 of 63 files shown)


hekike:
@DonutEspresso @retrohacker @yunong what do you think, should I continue? Even if I'm going to end up with necessary API change?

lib/middle.js
```diff
@@ -0,0 +1,258 @@
...
+/**
+ * Module variables.
+ * @private
+ */
+
+var env = process.env.NODE_ENV || 'development';
+var proto = {};
+
+/* istanbul ignore next */
+var defer =
```
retrohacker: Why are we checking for the existence of `setImmediate`?

hekike: Sry, I didn't make it clear that this is `connect`, I copy-pasted it, because I needed to extend it with the `next(false) logic. :(  Later we can try to open a PR or replace `next(false)` with something else.

hekike:
@retrohacker I added two more benchmarks: middleware and router-heavy
You can find the results in the description.

hekike:
Question: should `next(false)` only prevent to run the handlers in the same category or all of the handlers, which one:

1. `next(false)` in pre, stops the next `pre` handlers but runs all the `use` and router handlers
2. `next(false)` in pre, stops the next `pre` and `use` handlers but runs router
3. `next(false)` in pre, stops all the upcoming handlers

hekike:
Question: I don't find where the `createServer({ dtrace })` options it handled, does it still exist? I mean this option itself?

DonutEspresso:
Whoaaaaaaa this is awesome 🎉 🎉 🎉  

> Question: should next(false) only prevent to run the handlers in the same category or all of the handlers

The intention of false was to originally stop all subsequent middlewares in the event you had flushed a response already and considered this request complete. So I think 3 should be the desired behavior (if it isn't already).

hekike:
@DonutEspresso @yunong @retrohacker  I have an issue with the 'close' and 'aborted' test cases:
https://github.com/restify/node-restify/blob/master/test/server.test.js#L2648

No matter how I end/destroy/close the req/res/socket I always have both the `server.on('aborted')` and `server.on('close')` events, I cannot abort without `close` event, and I cannot close without `aborted` event. Is it possible to trigger `close` event without `aborted`? Does it make sense to separate the two connection states in `restify`? If I understand correctly we will always have `close` https://nodejs.org/api/http.html#http_event_aborted

yunong:
@hekike I think it may be the case that Node core actually fires both events? We could check the source code to be sure, but if that's the case, then it's okay that both events fire.

DonutEspresso:
Hmm great question - I was under impression a close wouldn't cause an abort but I can see both events being fired now. Short answer: I don't know! May have to dig into core to see if there is anything worth differentiating there.

hekike:
**close and aborted:** Thanks for the help! I've checked out Node core and seems to be that `emit('aborted')` is always immediately followed by an `emit('close')` in node core. *(I'm not 100% sure)* I also opened a question here: https://github.com/nodejs/help/issues/978

See: [_http_server.js#L417](https://github.com/nodejs/node/blob/3eb5778d1b1408d6223cdb3f5124a9b83f46f5eb/lib/_http_server.js#L417) and [http2/compat.js#L14](https://github.com/nodejs/node/blob/11a9f36cae0d1e6e11c9f40ec82d9f887aa0a911/lib/internal/http2/compat.js#L142)

in 7.x I unified server `close` and `aborted` events. restify throws only `RequestCloseError`, see: https://github.com/restify/node-restify/pull/1561/commits/0da821920f735a1b9b194d0b3ddcde360a0f77f3

lib/server.js
```diff
@@ -779,420 +700,250 @@ Server.prototype.toString = function toString() {
...
- * @param  {Response} res - response
- * @returns {undefined} no return value
+ * @function _onRequest
+ * @param    {Object}    req - the request object
+ * @param    {Object}    res - the response object
+ * @returns  {undefined} no return value
+ * @fires Request,Response#request
  */
-Server.prototype._routeAndRun = function _routeAndRun(req, res) {
+Server.prototype._onRequest = function _onRequest(req, res) {
```
DonutEspresso: So clean, so simple 💯

lib/server.js
```diff
@@ -1222,7 +980,122 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
...
+
+    // attach a listener for the response's 'redirect' event
+    res.on('redirect', function onRedirect(redirectLocation) {
+        self.emit('redirect', redirectLocation);
+    });
+};
+
+/**
+ * Wrapper method for emitting the after event. this is needed in scenarios
+ * where the async formatter has failed, and the ot assume that the
```
DonutEspresso: This comment looks a little outdated since we don't have async formatters anymore - don't want to bikeshed on this PR but could make a note to come back and clean up comments in another PR.

lib/server.js
```diff
@@ -1528,7 +1283,7 @@ function errEvtNameFromError(err) {
         // remap the name for router errors
         return 'VersionNotAllowed';
     } else {
-        return err.name.replace(/Error$/, '');
+        return err.name ? err.name.replace(/Error$/, '') : 'Error';
```
DonutEspresso: Under what scenarios does `err.name` not exist?

lib/server.js
```diff
@@ -779,420 +700,250 @@ Server.prototype.toString = function toString() {
...
+// 1. _onRequest (handle new request, setup request and triggers pre)
+//   2.1 _runPre
+//     2.2. _afterPre (runs after pre handlers are finisehd, triggers use)
+//       3.1 _runUse
+//         3.2. _afterUse (runs after use handlers are finised, triggers route)
+//          4.1 _runRoute
+//            4.2 _afterRoute (runs after last handler called next)
+//              4.3 _afterRoute (when, next('string') called)
+//              4.3 _routeErrorResponse (when error happaned in router)
+// 5. _finishReqResCycle (on response "finish" and "error" events)
```
DonutEspresso: This waterfall is great! The additional granularity is nice. Is this order different than the current behavior? I thought the order is as follows:

1. pre
2. router.find
3. use (use only applies to routes that were successfully routed)
4. done/done(err)

benchmark/benchmarks/router-heavy.js
```diff
@@ -0,0 +1,132 @@
+'use strict';
+
+var restify = process.argv.includes('version=head')
+    ? require('../../lib')
+    : require('restify');
+
+var server = restify.createServer();
+var path = '/whiskeys/scotch/islay/lagavulin/16-years/50';
```
DonutEspresso: 🍺

lib/server.js
hekike: You are right, I was dumb enough to work based on my `express` and `koa` experience instead of reading the docs carefully :/

> Allows you to add in handlers that run for all routes. Note that handlers added via use() will run only after the router has found a matching route. If no match is found, these handlers will never run. - restify docs

*(In express/koa everything is a middleware, the router as well, so they don't make difference between pre/after route, it depends on the order of the middleware "registration")*

Btw maybe it's misleading for others as well coming from the `express` world, shouldn't we keep `use` as an alias and introduce a new term that reflects it's behavior better? Like `afterRoute()` or `useRoute()`, these are not very good names, but something like these :)

lib/chain.js
```diff
@@ -0,0 +1,185 @@
...
+
+module.exports = Chain;
+
+/**
+ * Create a new middleware chain
+ *
+ * @public
+ * @class Chain
+ * @param {Object} [options] - options
+ * @param {Boolean} [options.onceNext=false] - Prevents calling next multiple
```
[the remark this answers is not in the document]
hekike: Yes, it's for backward compatibility. Making `strictNext` to the default causes performance degradation.

hekike:
@DonutEspresso I fixed the `use`, `route` handler running order issue. As a result of this, I also reverted all the related changes. Please check out the breaking changes list again, as it's smaller now.

I tried to improve the flow by using `vasync` and it's "competitors" but I always experienced significant performance degradation. I would skip it for now. Btw https://github.com/suguru03/neo-async was the fastest.

If I understand correctly then only the conditional handler need be improved to get your approval.

hekike:
@DonutEspresso with control flow: https://gist.github.com/hekike/26d486b2321c802513613fba86ca5b99

It's like 5-10% slower on synthetic tests, would you prefer this format?

hekike:
I updated this PR with the latest master, moved the current Radix tree based router lookup logic to a separate component with a more general interface. It will provide a way to use different router logic in the future.

Benchmarks on the updated PR with my new machine (pretty much the same) :p

cc @DonutEspresso @yunong 

```
pmarton@nflx node-restify (feat/new-router)*$ make benchmark 
up to date in 0.377s
? Do you want to compare HEAD with the stable release (6.3.4)? Yes
? How many connections do you need? 100
? How many pipelining do you need? 10
? How long does it take? 30
---- response-json ----
response-json throughput:
{
    "significant": "***",
    "equal": false,
    "wins": "head",
    "diff": "32.7%"
}

---- response-text ----
response-text throughput:
{
    "significant": "***",
    "equal": false,
    "wins": "head",
    "diff": "36.39%"
}

---- router-heavy ----
router-heavy throughput:
{
    "significant": "***",
    "equal": false,
    "wins": "head",
    "diff": "2702.78%"
}

---- middleware ----
middleware throughput:
{
    "significant": "***",
    "equal": false,
    "wins": "head",
    "diff": "43.48%"
}
```

docs/guides/6to7guide.md
```diff
@@ -0,0 +1,157 @@
...
+
+restify `7.x` comes with a completely new router and middleware logic that
+brings significant performance improvement to your application.
+From `v7.0.0` restify uses the Radix Tree based
+[find-my-way](https://github.com/delvedor/find-my-way) package as a router
+backend.
+
+## Breaking Changes
+
+### Server throws `RequestCloseError` instead of `RequestAbortedError`
```
DonutEspresso: I don't think the server actually "throws" right? Do we mean that the server just logs a `RequestCloseError`? We may also want to update RequestCloseError's default status code to 444.

docs/guides/6to7guide.md
```diff
@@ -0,0 +1,157 @@
...
+### Remove already deprecated `next.ifError`
+
+`next.ifError(err)` is not available anymore.
+
+### Disable DTrace probes by default
+
+DTrace probes comes with some performance impact that's fine for the sake of
+observability but you may don't use it at all.
+
+### Change in calling `next` multiple times
```
DonutEspresso: Is performance the motivation behind this new default behavior?

docs/guides/6to7guide.md
```diff
@@ -0,0 +1,157 @@
...
+    next();
+    next();
+});
+// -> throws an Error
+```
+
+### Router versioning and content type
+
+`accept-version` and `accept` based conditional routing moved to the
+`conditioalHandler` plugin, see docs or example:
```
DonutEspresso: s/conditioalHandler/conditionalHandler

lib/errorTypes.js
```diff
@@ -3,4 +3,4 @@
 var errors = require('restify-errors');
 
 errors.makeConstructor('RequestCloseError');
-errors.makeConstructor('RequestAbortedError');
+errors.makeConstructor('RouteMissingError');
```
DonutEspresso: Where is this error used?

lib/plugins/inflightRequestThrottle.js
```diff
@@ -65,8 +65,7 @@ function inflightRequestThrottle(opts) {
                 },
                 'maximum inflight requests exceeded, rejecting request'
             );
-            res.send(plugin._err);
-            return next(false);
+            return next(plugin._err);
```
DonutEspresso: I can't comment on the exact line, but I think this file here needs to create the error inline to get a proper stack trace (rather than creating it ahead of time).

lib/request.js
```diff
@@ -807,32 +809,47 @@ function patch(Request) {
...
+    /**
+     * Returns true when connection state is "close"
+     *
+     * @private
+     * @memberof Request
+     * @instance
+     * @function closed
+     * @returns {Boolean} is closed
+     */
+    Request.prototype.closed = function closed() {
```
DonutEspresso: 👍

lib/chain.js
```diff
@@ -0,0 +1,179 @@
...
+function Chain(options) {
+    assert.optionalObject(options, 'options');
+    options = options || {};
+    assert.optionalBool(options.onceNext, 'options.onceNext');
+    assert.optionalBool(options.strictNext, 'options.strictNext');
+
+    this.onceNext = !!options.onceNext;
+    this.strictNext = !!options.strictNext;
+
+    // strictNext next enfoces onceNext
```
DonutEspresso: s/enfoces/enforces

lib/chain.js
```diff
@@ -0,0 +1,179 @@
...
+ * @instance
+ * @param {Function|Chain} handler - handler or Chain instance
+ * @returns {undefined} no return value
+ */
+Chain.prototype.add = function add(handler) {
+    // wrap nested chain
+    if (typeof handler.run === 'function') {
+        var chain = handler;
+        handler = function nestedChainHandler(req, res, next) {
+            chain.run(req, res, next);
```
DonutEspresso: By chaining via a wrapper handler vs flattening the array, is there any impact on the result of `req.timers`?

lib/chain.js
```diff
@@ -0,0 +1,179 @@
...
+
+/**
+ * Returns the number of handlers
+ *
+ * @public
+ * @memberof Chain
+ * @instance
+ * @returns {Number} number of handlers in the stack
+ */
+Chain.prototype.count = function count() {
```
DonutEspresso: Potential bikeshed - why count() vs length()/.length property? (don't feel super strongly)

lib/chain.js
```diff
@@ -0,0 +1,179 @@
...
+/**
+ * Helper functions
+ * @private
+ */
+
+/**
+ * Invoke a handler.
+ *
+ * @private
+ * @param {Function} handler - rr function
```
DonutEspresso: `rr` function?

lib/router.js
```diff
@@ -155,534 +27,351 @@ function compileURL(options) {
...
-        DELETE: [],
-        GET: [],
-        HEAD: [],
-        OPTIONS: [],
-        PATCH: [],
-        POST: [],
-        PUT: []
-    };
+    // Internals
+    this._anonymusHandlerCounter = 0;
```
DonutEspresso: s/anonymus/anonymous

lib/router.js
```diff
@@ -155,534 +27,351 @@ function compileURL(options) {
...
+        spec: opts,
+        chain: chain
     };
-    routes.push(route);
 
-    if (!this.reverse[route.path.source]) {
-        this.reverse[route.path.source] = [];
-    }
+    handlers.forEach(function forEach(handler) {
+        // Assign name to anonymus functions
```
DonutEspresso: s/anonymus/anonymous

lib/router.js
```diff
@@ -155,534 +27,351 @@ function compileURL(options) {
...
+        handler._name =
```
DonutEspresso: Not sure I love that we're mutating user parameters - but wouldn't consider it blocking. Maybe worth adding a TODO for later (or filing issue).

lib/server.js
```diff
@@ -83,13 +79,17 @@ var PROXY_EVENTS = [
...
+ *  called more than once, enabled onceNext option
  * @param {Object} [options.httpsServerOptions] - Any options accepted by
  * [node-https Server](http://nodejs.org/api/https.html#https_https).
  * If provided the following restify server options will be ignored:
  * spdy, ca, certificate, key, passphrase, rejectUnauthorized, requestCert and
  * ciphers; however these can all be specified on httpsServerOptions.
- * @param {Boolean} [options.strictRouting=false] - If set, Restify
- * will treat "/foo" and "/foo/" as different paths.
+ * @param {Boolean} [options.noWriteContinue=false] - prevents
+ *  `res.writeContinue()` in `server.on('checkContinue')` when proxing
```
DonutEspresso: proxying?

lib/server.js
```diff
@@ -496,8 +499,9 @@ Server.prototype.use = function use() {
     var self = this;
     var handlers = Array.prototype.slice.call(arguments);
 
-    argumentsToChain(handlers).forEach(function forEach(h) {
-        self.chain.push(h);
+    argumentsToChain(handlers).forEach(function forEach(handler) {
+        handler._name = handler.name || 'use-' + self.useChain.count();
```
DonutEspresso: Isn't this already happening in router?

lib/server.js
```diff
@@ -466,8 +468,9 @@ Server.prototype.pre = function pre() {
     var self = this;
     var handlers = Array.prototype.slice.call(arguments);
 
-    argumentsToChain(handlers).forEach(function forEach(h) {
-        self.before.push(h);
+    argumentsToChain(handlers).forEach(function forEach(handler) {
+        handler._name = handler.name || 'pre-' + self.preChain.count();
```
DonutEspresso: Isn't this already happening in router?

lib/server.js
```diff
@@ -779,420 +733,306 @@ Server.prototype.toString = function toString() {
...
+// Lifecycle:
+//
+// 1. _onRequest (handle new request, setup request and triggers pre)
+//   2.1 _runPre
+//     2.2. _afterPre (runs after pre handlers are finisehd, triggers route)
+//       3.1 _runRoute
+//         3.2. _afterRoute (runs after route handlers are finised,
+//                          triggers use)
+//          4.1 _runUse
+//            4.2. _afterUse
```
DonutEspresso: I know we chatted about this before, are we changing the pre -> use -> route order? Looks like this is breaking from 6.x.

lib/server.js
```diff
@@ -779,420 +733,306 @@ Server.prototype.toString = function toString() {
...
-            id,
-            req.method,
-            req.href(),
-            req.headers
-        ];
-    });
-
-    req.timers = [];
+    // Preserve handler err for finish event
+    res.err = res.err || err;
```
DonutEspresso: I didn't see where this is initially set - what's the use case for storing it on the response directly?

lib/server.js
```diff
@@ -1223,7 +1070,122 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
...
+        req.removeListener('close', onClose);
+        res.removeListener('error', onError);
+        self._finishReqResCycle(req, res, res.err);
+    }
+    function onError(err) {
+        req.removeListener('close', onClose);
+        res.removeListener('finish', onFinish);
+        self._finishReqResCycle(req, res, err);
+    }
+    res.once('finish', onFinish);
```
DonutEspresso: Is this a breaking change from 6.x? IIRC in 6.x we consider a request finished once it has reached the end of the handler chain. Using this pattern, it won't fire until the request is completely done. 

I have to think about whether this is better/worse/different, but definitely has an impact on how we might view metrics. cc @yunong

lib/server.js
```diff
@@ -1223,7 +1070,122 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
...
+    var self = this;
+    var route = req.route; // can be undefined when 404 or error
+
+    // res.finished is set by node core's response object, when
+    // res.end() completes. if the request was closed by the client, then emit
+    // regardless of res status.
+
+    // after event has signature of function(req, res, route, err) {...}
+    if (!res.finished && !req.closed()) {
+        res.once('finish', function resFinished() {
```
DonutEspresso: Since `finishReqResCycle()` doesn't get triggered until finish event fires, is this scenario possible anymore?

DonutEspresso:
Great work! Love that we added benchmarks - thanks for digging in and getting those perf numbers. 💯   Left some comments - mostly minor, the two big things for me are:

1) 6.x order is pre->use->route, order here looks like pre->route->use. Did we agree to a breaking change on that?
2) In 6.x a request is considered "complete" (i.e., causing audit log and metrics plugin to fire) when reaching the end of the next chain. "Completion" is synonymous with the 'after' event, where audit log and metrics are triggered. With this PR, "completion" gets delayed until the request object's finish event is fired. Not sure if this is good/bad yet but it is definitely different and may have an impact on how we interpret metrics. Thoughts from the rest of the team? cc @yunong @retrohacker

docs/guides/6to7guide.md
retrohacker: Also have a preference for this to remain unchanged. Throwing by default helps us catch a whole class of bugs.

lib/chain.js
```diff
@@ -0,0 +1,185 @@
...
+        handle(err, req, res, next);
+        return;
+    } else if (!hasError && arity < 4) {
+        // request-handling middleware
+        handle(req, res, next);
+        return;
+    }
+
+    // continue
+    next(error, req, res);
```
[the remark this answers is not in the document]
retrohacker: :+1: on leaving style changes to their own PR and fixing there if they pass lint. Suspect this isn't the only place in the code base we do this.

docs/guides/6to7guide.md
hekike: Yes, performance is the motivation behind.
@retrohacker do you mean at Netflix? We can re-enable it with the option for NQ services.

lib/errorTypes.js
hekike: This error existed before but was created locally.
When a code calls `next('route-name')` but the route doesn't exist.
It's a programmatic error.

lib/chain.js
hekike: Me neither, I thought it's a good idea to differentiate it from the JavaScript Array length

lib/chain.js
hekike: I removed the nested chain support, it's not used anywhere and not documented neither.

lib/server.js
hekike: Hmm, I don't remember where I found this. I just documented it, could you help me with the correct behaviour?

lib/server.js
hekike: Router, pre, and use are registered separately with a different prefix.

lib/server.js
hekike: Router, pre, and use are registered separately with a different prefix.

lib/server.js
hekike: No, it's the same, the doc was wrong, fixed.

lib/server.js
hekike: It's used when the request succeeds but the response is an error:
https://github.com/restify/node-restify/blob/feat/new-router/lib/server.js#L1093

This is how it's ensured that error is always emitted in `after`.
It's not an elegant solution decorating the `res` object, but in the earlier version it was way to complex, I like that now it's based on the response events.

docs/guides/6to7guide.md
retrohacker: @hekike I just mean in general. We are a framework for building production ready REST services, I'm not sure this is in line with our goals as a project.

hekike:
We discussed the request "complete" topic in person with @DonutEspresso and we decided that the new behavior is acceptable and even better in some use cases like load shedding calculations where you really want to know when the request doesn't use your resources anymore. 

To have the correct timing about handler categories like `pre`, `use` and `router` execution durations a new timing object will be introduced in the next version of restify, related issue: https://github.com/restify/node-restify/issues/1603

hekike:
@DonutEspresso I added the new request timings and explained the change in the migration guide in a separate commit.

hekike:
We discussed in person that we need two latencies one when the request is flushed and one when all handlers are finished. The `after` event and inflight request counter will wait for both request flushing and handler finish.

lib/plugins/metrics.js
```diff
@@ -39,7 +39,7 @@ function createMetrics(opts, callback) {
             // REST verb
             method: req.method,
             // overall request latency
-            latency: hrTimeDurationInMs(req._time, process.hrtime()),
+            latency: hrTimeDurationInMs(req._time, req._timeFlushed),
```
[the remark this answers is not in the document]
DonutEspresso: Minus this last timer I think we are good to go!

hekike:
@DonutEspresso `totalLatency` added! https://github.com/restify/node-restify/pull/1561/commits/1acc1be51ce634eb571584581e6f2b960abe4a5b

Issue #1416, opened 2017-07-24
Default to `utf-8` charEncoding


retrohacker:
- [x] Used appropriate template for the issue type
- [x] Searched both open and closed issues for duplicates of this issue
- [x] Title adequately and _concisely_ reflects the feature or the bug

# Feature Request

## Use Case

To explicitly tell the client which encoding we are sending over the wire.

## Example API

When sending data over the wire, we should explicitly define the content encoding in the `Content-Type` header. By default the string that gets converted to a buffer by Node.js is encoded as utf-8. This seems to be the best candidate for a default.

## Are you willing and able to implement this?

Yes

PR #1564, opened 2017-11-14
feat(response): set content-type charset default to utf-8 #1416

andersnylund:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #1416 
* Issue #
* Issue #

# Changes

> What does this PR do?

Makes a Response's content-type to default to charset=utf-8, if charSet() not specifically called

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -491,6 +491,8 @@ function patch(Response) {
 
         if (self._charSet) {
             type = type + '; charset=' + self._charSet;
+        } else {
+            type = type + '; charset=utf-8';
         }
 
         // Update header to the derived content type for our formatter
--- a/test/response.test.js
+++ b/test/response.test.js
@@ -514,7 +514,10 @@ test('should fail to set header due to missing formatter', function(t) {
     CLIENT.get(join(LOCALHOST, '/11'), function(err, _, res) {
         t.ifError(err);
         t.equal(res.statusCode, 200);
-        t.equal(res.headers['content-type'], 'application/octet-stream');
+        t.equal(
+            res.headers['content-type'],
+            'application/octet-stream; charset=utf-8'
+        );
         t.end();
     });
 });
@@ -617,6 +620,33 @@ test('GH-1429: setting code with res.status not respected', function(t) {
 
     CLIENT.get(join(LOCALHOST, '/404'), function(err, _, res) {
         t.equal(res.statusCode, 404);
+        t.end();
+    });
+});
+
+test('should default Response content-type to charset=utf-8', function(t) {
+    SERVER.get('/200', function(req, res, next) {
+        res.status(200);
+        res.send(null);
+    });
+
+    CLIENT.get(join(LOCALHOST, '/200'), function(err, _, res) {
+        t.equal(res.headers['content-type'], 'application/json; charset=utf-8');
+        t.end();
+    });
+});
+
+test('should not overwrite setting response content-type charset', function(t) {
+    SERVER.get('/200', function(req, res, next) {
+        res.charSet('ISO-8859-1');
+        res.send(null);
+    });
+
+    CLIENT.get(join(LOCALHOST, '/200'), function(err, _, res) {
+        t.equal(
+            res.headers['content-type'],
+            'application/json; charset=ISO-8859-1'
+        );
         t.end();
     });
 });


retrohacker:
Also request @lrowe

lib/response.js
```diff
@@ -491,6 +491,8 @@ function patch(Response) {
 
         if (self._charSet) {
             type = type + '; charset=' + self._charSet;
+        } else {
+            type = type + '; charset=utf-8';
```
lrowe: Choosing a default content type for a formatter should probably be the responsibility of the formatter, since binary content type should not have a charset.

Existing formatters are currently also tied to a particular charset since they set the `Content-Length` based on `Buffer.byteLength(data)` (Buffer.byteLength without a charset defaults to utf-8.) Example: https://github.com/restify/node-restify/blob/master/lib/formatters/json.js

(Setting Content-Length in the formatter should no longer be needed in recent NodeJS versions (>1.5.0), see: nodejs/node#1062.)

Perhaps it would make more sense to set the Content-Type header only after calling the formatter, providing the opportunity for the formatter to set a format appropriate default charset.

```js
        const results = formatter(self.req, self, body);
        
        if (self._charSet) {
            type = type + '; charset=' + self._charSet;
        }

        // Update header to the derived content type for our formatter
        self.setHeader('Content-Type', type);

        // Finally, invoke the formatter and flush the request with it's results
        return flush(self, results);
```

test/response.test.js
```diff
@@ -514,7 +514,10 @@ test('should fail to set header due to missing formatter', function(t) {
     CLIENT.get(join(LOCALHOST, '/11'), function(err, _, res) {
         t.ifError(err);
         t.equal(res.statusCode, 200);
-        t.equal(res.headers['content-type'], 'application/octet-stream');
+        t.equal(
+            res.headers['content-type'],
+            'application/octet-stream; charset=utf-8'
+        );
```
lrowe: Binary content types should not include a charset as it only makes sense for textual content types.

lib/response.js
DonutEspresso: I'm confused -  I thought the content type indirectly specifies the formatter, not the other way around?

test/response.test.js
andersnylund: That is true. Didn't think of that. Having a list of different mime-types that should default to utf-8 is not convenient either keeping in mind all combinations. How about a blacklist of mime-types where not to format with utf-8?

lib/response.js
lrowe: @DonutEspresso while the content type picks the formatter the formatter should probably pick the charset if it is unspecified.

test/response.test.js
lrowe: The list of binary types is also unbounded. I think it makes most sense to make the formatter to be responsible for picking a default / forcing a particular charset.

lrowe: The Python WebOb library uses an include list for setting a default charset:

>     * The ``default_charset`` is used as the default character set to return on
>       the ``Content-Type`` header, if the ``Content-Type`` allows for a
>       ``charset`` parameter. Currently the only ``Content-Type``'s that allow
>       for a ``charset`` are defined to be: ``text/*``, ``application/xml``, and
>       ``*/*+xml``. Any other ``Content-Type``'s will not have a ``charset``
>       added.

https://github.com/Pylons/webob/blob/master/src/webob/response.py#L142

We'd probably need to add at least "application/javascript", "application/json", and "*/*+json" to that list.

Maintaining such a list does seem very inconvenient. Attaching the default to the formatter somehow seems more extensible.

retrohacker: Would this help? https://www.npmjs.com/package/mime-types

```js
mime.charset('text/markdown') // 'UTF-8'
```

lrowe:
Binary formatters will likely return a Buffer so perhaps we could do:
```
        const results = formatter(self.req, self, body);
        
        if (self._charSet) {
            type = type + '; charset=' + self._charSet;
        } else if (typeof results === "string") {
            type = type + '; charset=utf-8';
        }

        // Update header to the derived content type for our formatter
        self.setHeader('Content-Type', type);

        // Finally, invoke the formatter and flush the request with it's results
        return flush(self, results);
```

Issue #1567, opened 2017-11-17
uncaughtException on route error when response have already been sent


beckend:
we have middlewares and routes which send response normally, however server listeners such as `MethodNotAllowed ` will trigger the route error handler and send response regardless if payload has been sent or not and causes an uncaughtException exception needlessly.

Issue #1557, opened 2017-11-04
Flaky test in static plugin with Node v8


hekike:
# Bug Report

```sh
static resource plugin static does not open a file stream and next() is properly called when the client disconnects immediately after sending a request:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

## Restify Version

HEAD

## Node.js Version

First appeared with `v8.9.0`

## Repro case

TravisCI job

## Cause

Don't know.

## Are you willing and able to fix this?

Yes, I need help.

PR #1568, opened 2017-11-17
server _routeErrorResponse  avoiding uncaughtException by only sending response when not sent

beckend:
middlewares and routes which send response normally, however server listeners such as MethodNotAllowed will trigger the route error handler and send response regardless if payload has been sent or not and causes an uncaughtException needlessly.

Closes:

* Issue #1567


hekike:
Thanks for the contribution!

I applied a very similar fix on the 7.x branch: https://github.com/restify/node-restify/pull/1561 with https://github.com/restify/node-restify/pull/1561/commits/e31564172a82b886613961e40da9d3fb88d30256 and I think it's also related to the https://github.com/restify/node-restify/issues/1557

@beckend could you please add a test to cover this case?

beckend:
@hekike test proven

test/server.test.js
```diff
@@ -224,6 +224,33 @@ test('rm route and clear cached route', function(t) {
     });
 });
 
+test('_routeErrorResponse does not cause uncaughtException when called when header has already been sent', function(
+    t
+) {
+    SERVER.on('MethodNotAllowed', function(req, res, error, next) {
+        res.json(405, { status: 'MethodNotAllowed' });
+        var nextHasErrored;
```
hekike: You can do this:

```js
try {
    next();
} catch (err) {
    t.fail('next() should not throw error when header has already been sent');
}
t.end();
```

hekike:
@beckend thanks, a small comment, otherwise looks great!

hekike:
@beckend thanks for your contribution! Landed as `v6.3.4`.

Issue #779, opened 2015-03-17
set-cookie headers should not include comma separated values


yunong:
According to http://tools.ietf.org/html/rfc6265#section-3
```
   Origin servers SHOULD NOT fold multiple Set-Cookie header fields into
   a single header field.  The usual mechanism for folding HTTP headers
   fields (i.e., as defined in [RFC2616]) might change the semantics of
   the Set-Cookie header field because the %x2C (",") character is used
   by Set-Cookie in a way that conflicts with such folding.
```

set-cookie headers should not include comma separated values -- however, with PR #693 this is enabled comma separated values for all headers, including set-cookie, which breaks browser parsing of set-cookie headers.

yunong:
fixed

lrowe:
The HTTP RFC only disallows the merging of multiple Set-Cookie headers into a single header. Multiple Set-Cookie headers are common and necessary.

Testing back to the latest Node 4.x LTS release, `resposne.setHeader("Set-Cookie", ["a=1", "b=2"])` will render a multiple Set-Cookie values correctly:
```
$ curl "http://localhost:3000/" -v
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 3000 (#0)
> GET / HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/7.43.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Set-Cookie: a=1
< Set-Cookie: b=2
< Date: Tue, 21 Nov 2017 20:19:28 GMT
< Connection: keep-alive
< Content-Length: 21
< 
* Connection #0 to host localhost left intact
```

PR #1570, opened 2017-11-21
fix: Allow multiple unmerged set-cookie headers.

lrowe:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #779 - set-cookie headers should not include comma separated values

# Changes

Fixes the fix for #779 to allow multiple set-cookie headers. The RFC only disallows these headers from being merged into a single header, it does not disallow multiple headers.


DonutEspresso:
Thanks for the PR! Oddly, this is the exact opposite of #779, but appears to have the desired behavior. Wonder if something in Node core changed where it knows about how to set concatted headers for cookies, rather than universally transforming multiple values into a comma delimited string. 

Maybe @yunong can remember how the original issue was manifesting?

lrowe:
It looks like @yunong was wanting to avoid set-cookie headers being merged into a single header (as this is not valid per RFC.) Maybe this is something Restify did at one point? It looks like Node has supported Arrays for multiple header lines back to Sep 2010 https://github.com/nodejs/node-v0.x-archive/commit/6560ab9073f5b23c467dd0662c4cd236685564ed yet I see recent bugs like https://github.com/nodejs/node/issues/3591 reporting that header merging is happening.

I've run this test script back to v0.10.48 and see an appropriate response with two separate Set-Cookie headers.
```js
"use strict";

var http = require('http');
var port = 3000;

function requestHandler(request, response) {
  console.log(request.url);
  response.setHeader("Set-Cookie", ["a=1", "b=2"]);
  response.end('Hello Node.js Server!');
}

var server = http.createServer(requestHandler);

server.listen(port, function (err) {
  if (err) {
    return console.log('something bad happened', err);
  }

  console.log('server is listening on ' + port);
})
```

DonutEspresso:
LGTM

PR #1579, opened 2017-12-01
Fix #1565: Return 444 status code for closed and aborted requests

### LINKED ISSUE #1565: Close/aborted requests should log as http 444

# Bug Report

When a request is closed by the client or aborted, restify doesn't change the default status code which is 200. This trickles through the audit log as well as the metrics plugin. Technically a response was never written, but if you are looking at aggregate metrics this can be quite misleading. 

## Restify Version

All

## Node.js Version

All

## Expected behaviour

I would expect an http 444 for `req.on('close')`. This seems applicable for `req.on('abort')` as well, but open to other status codes.

## Actual behaviour

Restify leaves the default status code of 200. 

## Repro case

```
const log = require('bunyan').createLogger({ name: 'foo '});
const server = restify.createServer({
    name: 'foo'
});

server.on('after', restify.plugins.auditLogger({
    server: server, // this ensures `audit` event is called
    event: 'after',
    log: log.child({
        component: 'audit'
    })
}));

    setTimeout(function() {
        return next();
    }, 3000);
});

server.listen(3000);
```

Start the process, then curl and ctrl-c the request. Wait for timeout to expire. See 200 in the audit logs. 

## Cause

These event listeners should probably set the status code as appropriate:
https://github.com/restify/node-restify/blob/master/lib/server.js#L1012

benjaminRomano:
@DonutEspresso 
## Issues

Closes:
* https://github.com/restify/node-restify/issues/1565

# Changes
Sets the response status code to 444 for aborted or closed requests.

The client will never see this response. This is solely for plugins like audit-logger and metrics to record an accurate `statusCode` instead of what was set using `res.send`


retrohacker:
Thanks for this @benjaminRomano 

IIRC we just did something similar at another layer in our stack @DonutEspresso?

benjaminRomano:
:+1: I agree checking connectionState explicitly is safer. I've updated the PR 

@DonutEspresso

benjaminRomano:
@retrohacker I believe this is ready to be merged

Issue #1343, opened 2017-05-10
Support custom fields for audit logger


retrohacker:
Migrated from https://github.com/restify/plugins/issues/74

sfaizanh:
Hi @retrohacker https://github.com/restify/node-restify/compare/master...sfaizanh:patch-1#diff-a64f77eab03fa30693debd295ef35673R234 should i create PR?

PR #1590, opened 2018-01-07
Custom fields from request in audit logger

sfaizanh:
Currently there is not way to log custom fields in auditLogger, This PR support custom logging for auditlogger
```
restify.plugins.auditLogger({
    log: bunyan.createLogger({
    name: 'audit',
    streams: [
        {
            level: 'info',
            type: 'raw'
        }
    ]
    }),
    event: 'after',
    customFields: ['user']
})
```
Pre-Submission Checklist
- [x] Ran the linter and tests via make prepush

**Closes:**
Issue #1343


sfaizanh:
Node stable build on Travis failed (but all build are passed). Looks like a random failure with npm. Closing (will reopen to retrigger Travis build).

retrohacker:
If you agree with the changes, feel free to merge @hekike!

lib/plugins/audit.js
```diff
@@ -50,6 +51,7 @@ function getResponseHeaders(res) {
  * the audit log object programmatically
  * @param {boolean} [opts.printLog=true] - Whether to print the log
  * via the logger.
+ * @param {Array<String>} [opts.customFields=[]] - Custom fields for logger
  * @returns {Function} Handler
  * @fires audit when an audit log has been generated
  * @example
```
hekike: could you please provide a JSDoc example?

hekike:
@sfaizanh @retrohacker small comment about docs.

lib/plugins/audit.js
sfaizanh: Should i change `Array<String>` to `string[]`?

hekike: Yes, please. Also you can add a new example to JSDoc like:
```
* @example
* <caption>Some text</caption>
* Some JS Code

**Comment by sfaizanh** [comment]
Hi @hekike, any plan to merge?

File: lib/plugins/audit.js [hunk 2]
```diff
@@ -6,6 +6,7 @@ var assert = require('assert-plus');
 var bunyan = require('bunyan');
 var HttpError = require('restify-errors').HttpError;
 var VError = require('verror');
+var lodash = require('lodash');
```
hekike: could you change this line to `var _ = require('lodash');`?
We call lodash `_` in other files in restidf.

hekike:
@sfaizanh yes, sry for the delay. One last comment.

sfaizanh:
I haven't changed https://github.com/sfaizanh/node-restify/blob/6f6ae16c2f013f6e35b3b50147d99fcbdd176f54/test/plugins/audit.test.js#L89 but `lodash` as variable is used, should i change it?

DonutEspresso:
This seems overlapping with the existing `context` option:
https://github.com/restify/node-restify/blob/master/test/plugins/audit.test.js#L625

The `context` option allows specifying fields as well as ability to format. Do we need both of these options?

retrohacker:
I wasn't aware of the `context` key, though it is documented: http://restify.com/docs/plugins-api/#auditlogger

Does this do what you were needing it to @sfaizanh?  Sorry that we got so far with this PR before this was brought up.

hekike:
I'd prefer to move forward with this PR either way.
@sfaizanh did you have time to check out the `context` key? Would it solve your use-case?

sfaizanh:
Ok I’ll test this

PR #1593, opened 2018-01-30
Use `Buffer.isBuffer` instead of `util.isBuffer`.

NatalieWolfe:
## Pre-Submission Checklist

## Issues

No issues opened for this.

# Changes

Replaced `util.isBuffer` call with `Buffer.isBuffer`.

[`util.isBuffer`](https://nodejs.org/api/util.html#util_util_isbuffer_object) was introduced in Node 0.11 and deprecated in Node 4. [`Buffer.isBuffer`](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_isbuffer_obj) was introduced in Node 0.1 and has not been deprecated. Since Restify supports back to 0.10 according to the package.json `engines` field, `util.isBuffer` cannot be used.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -5,7 +5,6 @@
 var http = require('http');
 var sprintf = require('util').format;
 var url = require('url');
-var util = require('util');
 
 var assert = require('assert-plus');
 var mime = require('mime');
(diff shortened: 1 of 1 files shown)


retrohacker:
Awesome! Thank you for this!

Looks like the tests are failing lint, `util` is no longer used in that file.

NatalieWolfe:
@retrohacker Yup, forgot to commit that correction after running the tests! Original change was done via GitHub's web UI which doesn't run commit hooks.

NatalieWolfe:
@retrohacker The tests are passing now.

Issue #1411, opened 2017-07-13
jsonBodyParser fails with invalid JSON when the first invalid character is %


Temetz:
# Bug Report
Sending invalid JSON with % as the first invalid character to bodyParser will crash restify.

For example the following payload will always fail:
```
{"answer": "firefoxurl:test|"%20-new-window%20javascript:alert(\'Cross%2520Browser%2520Scripting!\');""}
```
With other invalid characters the jsonBodyParser works as expeceted.

## Restify Version
v 5.0.0

## Node.js Version
v 6.10.3

## Expected behaviour

Should return:
```
HTTP/1.1 400 Bad Request
{"code":"InvalidContent","message":"Invalid JSON: Unexpected token % in JSON at position 29"}
```
## Actual behaviour

```
/home/temetz/work-project/backend/node_modules/restify-errors/node_modules/extsprintf/lib/extsprintf.js:72
                        throw (new Error('too few args to sprintf'));
                        ^

Error: too few args to sprintf
    at jsSprintf (/home/temetz/work-project/backend/node_modules/restify-errors/node_modules/extsprintf/lib/extsprintf

.js:72:11)
    at parseConstructorArguments (/home/temetz/work-project/backend/node_modules/restify-errors/node_modules/verror/li
b/verror.js:120:26)
    at WError (/home/temetz/work-project/backend/node_modules/restify-errors/node_modules/verror/lib/verror.js:414:11)
    at HttpError (/home/temetz/work-project/backend/node_modules/restify-errors/lib/baseClasses/HttpError.js:35:12)
    at RestError (/home/temetz/work-project/backend/node_modules/restify-errors/lib/baseClasses/RestError.js:27:15)
    at new acc.(anonymous function) (/home/temetz/work-project/backend/node_modules/restify-errors/lib/restErrors.js:5
9:19)
    at parseJson (/home/temetz/work-project/backend/node_modules/restify/lib/plugins/jsonBodyParser.js:40:25)
    at Server.parseBody (/home/temetz/work-project/backend/node_modules/restify/lib/plugins/bodyParser.js:96:13)
    at next (/home/temetz/work-project/backend/node_modules/restify/lib/server.js:1028:30)
    at f (/home/temetz/work-project/backend/node_modules/once/once.js:25:25)
    at IncomingMessage.done (/home/temetz/work-project/backend/node_modules/restify/lib/plugins/bodyReader.js:135:13)
    at IncomingMessage.g (events.js:292:16)
[...]

PR #1597, opened 2018-02-05
fix(jsonBodyParser): fix percent sign causing server fail (#1411)

wojtkowiak:
This is a fix for https://github.com/restify/node-restify/issues/1411 which seems to be still happening.


DonutEspresso:
Thanks @wojtkowiak, looks like there are multiple places for this to trigger. Thank you for the PR _with_ accompany tests!

wojtkowiak:
Happy to help, thanks for fast merge! Greets!

PR #1598, opened 2018-02-05
plugins.md - metrics requires an opt structure

mtmail:
Documentation improvement.

The call example for plugins.metrics didn't include the required first parameter (which itself requires a `server` key to be set). Compare to https://github.com/restify/node-restify/blob/master/test/plugins/metrics.test.js

Closes no issue(s).

### DIFF
--- a/docs/_api/plugins.md
+++ b/docs/_api/plugins.md
@@ -1052,7 +1052,7 @@ about that request.
 **Examples**
 
 ```javascript
-server.on('after', plugins.metrics(function onMetrics(err, metrics) {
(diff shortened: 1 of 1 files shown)


hekike:
@mtmail good catch, thanks for your contribution!
The `_api/*.md` files are auto-generated from the JSDoc comments in the code. Could you change your PR to pick the example from the JSDoc comment? 
You can find it here: https://github.com/restify/node-restify/blob/master/lib/plugins/metrics.js#L23

You can re-generate the docs output .md files with `make docs-build`.

mtmail:
Sure. Replaced by https://github.com/restify/node-restify/pull/1599

PR #1599, opened 2018-02-06
Update metrics plugin example

mtmail:
Documentation improvement.

The call example for plugins.metrics didn't include the required first parameter (which itself requires a server key to be set). Compare to https://github.com/restify/node-restify/blob/master/test/plugins/metrics.test.js

Replaces https://github.com/restify/node-restify/pull/1598. Closes no issue(s).


retrohacker:
Thank you for this @mtmail :heart:

PR #1611, opened 2018-02-24
docs: Mark options object as optional in createServer

peterblazejewicz:
> What does this PR do?

This change corrects the JSDoc annotation and removes errors from
linters and checkers like built-in into VSCode when using the
createMethod without an option object.

Thanks!


hekike:
@peterblazejewicz thank you for the contribution!

PR #1652, opened 2018-05-01
feat: Added export for Chain

paulbakker:
Added an export for Chain so that it can be used in libraries such as enroute. 
Specifically, this will be used in enroute to add a development mode that support hot reloading of routes. To be able to do this the chain logic needs to be available in enroute, hence the extra export in restify.

Although the code change itself is trivial, it does imply that Chain becomes part of the API.


hekike:
We agreed in person that we won’t expose `Chain` as a public API. 
If we need to do it we will create a `compose([handler1, handler2, ...])`utility function which will create a chain in the background and return that `chain.run` function.

PR #1660, opened 2018-05-15
added exported function to compose a handler chain

paulbakker:
This compose function makes it possible to create a handler chain, without the need to export Chain publicly. This will be used to implement hot-reloading in the enroute project.
Discussed the idea in person with @hekike.


lib/chainComposer.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
```
hekike: Could you move this under a `utils` and export it like `module.exports.utils = require('./utils');`.
Would you mind adding a unit test?

hekike:
@DonutEspresso what do you think? should we export it under a namespace like `utils`, `helpers` etc? What would be the best naming?

DonutEspresso:
Either is fine - I don't feel too strongly about it. Let's go with `helpers.compose`? Since the name "Chain" has no meaning outside of restify, I think keeping the exported name generic is probably good. We can keep the file name `composeChain`.

hekike:
@paulbakker okay thanks, one small thing: please just call it `helpers.compose`, see Alex's reasoning.

paulbakker:
@hekike renamed to `helpers.compose` and squashed commits.

Issue #1348, opened 2017-05-15
Server constructor mutates global state


retrohacker:
- [ X ] Used appropriate template for the issue type
- [ X ] Searched both open and closed issues for duplicates of this issue
- [ X ] Title adequately and _concisely_ reflects the feature or the bug

# Bug Report

## Restify Version

5.x

## Node.js Version

All

## Expected behaviour

Constructor should only effect the object being instantiated

## Actual behaviour

Constructor effects all future server objects

## Cause

https://github.com/restify/node-restify/blob/5.x/lib/server.js#L306-L311

Since `PROXY_EVENTS` are shared between every instance of a restify server, and `upgrade` is added but never removed, all future instances of `Server` will have the `upgrade` event proxied.

## Are you willing and able to fix this?

Yes. It seems the  fix should be to make `PROXY_EVENTS` local.

DonutEspresso:
Nice catch, code certainly reads like this is the case.

DonutEspresso:
Fixed by #1661

PR #1661, opened 2018-05-16
fix: proxy events into instance var and add test script

akoserwal:
## Pre-Submission Checklist

- [ x] Opened an issue discussing these changes before opening the PR
- [ x] Ran the linter and tests via `make prepush`
- [ x] Included comprehensive and convincing tests for changes

## Issues
#1348 
Closes:

* Issue #1348
* Issue #
* Issue #

Since PROXY_EVENTS are shared between every instance of a restify server, and upgrade is added but never removed, all future instances of Server will have the upgrade event proxied.

# Changes

> What does this PR do?
update PROXY_EVENTS to an instance variable.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -187,7 +187,7 @@ function Server(options) {
 
     this.router.on('mount', this.emit.bind(this, 'mount'));
 
-    if (!options.handleUpgrades && this.proxyEvents.indexOf('upgrade') === -1) {
+    if (!options.handleUpgrades) {
         this.proxyEvents.push('upgrade');
     }
     this.proxyEvents.forEach(function forEach(e) {
--- a/lib/server.js
+++ b/lib/server.js
@@ -34,14 +34,6 @@ patchRequest(http.IncomingMessage);
 ///--- Globals
 
 var sprintf = util.format;
-var PROXY_EVENTS = [
-    'clientError',
-    'close',
-    'connection',
-    'error',
-    'listening',
-    'secureConnection'
-];
 
 ///--- API
 
@@ -138,6 +130,14 @@ function Server(options) {
     var fmt = mergeFormatters(options.formatters);
     this.acceptable = fmt.acceptable;
     this.formatters = fmt.formatters;
+    this.proxyEvents = [
+        'clientError',
+        'close',
+        'connection',
+        'error',
+        'listening',
+        'secureConnection'
+    ];
 
     if (options.spdy) {
         this.spdy = true;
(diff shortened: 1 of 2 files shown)


akoserwal:
@DonutEspresso, @hekike : I created a new branch. Instead of rebase my previous branch. I accidentally clicked over update branch on the github uI. Which will add non-essential commits. I will take care from next time.

DonutEspresso:
Thanks @akoserwal! Apologies for the trouble. I left some minor comments - let me know if they make sense!

test/server.test.js
```diff
@@ -2430,3 +2430,19 @@ test('should emit error with multiple next calls with strictNext', function(t) {
         });
     });
 });
+
+test('should have proxy event handlers as instance', function(t) {
+    var server = restify.createServer({
+        handleUpgrades: false
+    });
+    t.equal(server.proxyEvents.length, 7);
```
DonutEspresso: I think I expect handleUpgrades: true to be 7, and 6 for false (may be related to the && indexOf check above)

lib/server.js
```diff
@@ -187,10 +187,10 @@ function Server(options) {
 
     this.router.on('mount', this.emit.bind(this, 'mount'));
 
-    if (!options.handleUpgrades && PROXY_EVENTS.indexOf('upgrade') === -1) {
-        PROXY_EVENTS.push('upgrade');
+    if (!options.handleUpgrades && this.proxyEvents.indexOf('upgrade') === -1) {
```
DonutEspresso: Now that proxyEvents is specific to the instance of the server being created, I don't think we even need the indexOf check right?

test/server.test.js
akoserwal: As per the doc

options.handleUpgrades Boolean Hook the upgrade event from the node HTTP server, pushing Connection: Upgrade requests through the regular request handling chain. (optional, default false) (http://restify.com/docs/server-api/). 

We are updating proxyEvents array when 
`options.handleUpgrades:false`
size will be 7 and in-case of true. we will skip it.

lib/server.js
akoserwal: @DonutEspresso: right, I can remove it.

test/server.test.js
DonutEspresso: Ah, I can't read. Must have been a long day. :D

DonutEspresso:
@akoserwal thank you for the contribution! 🎉

akoserwal:
@DonutEspresso : Thank you :)

Issue #1639, opened 2018-04-04
v7 middleware functions no longer have server context (undocumented breaking change)


davidmarkclements:
- [x] Used appropriate template for the issue type
- [x] Searched both open and closed issues for duplicates of this issue
- [x] Title adequately and _concisely_ reflects the feature or the bug

# Bug Report

## Restify Version

v7.1.0
v7.0.0

## Node.js Version

v8.11.1

### If the breaking change is not deliberate:
The `this` value of a middleware function should be the server instance 

### If the breaking change is deliberate:
The changelog and migration guide (http://restify.com/docs/6to7/) should document the change

## Actual behaviour

The `this` value is `undefined` (or global in non-strict context)

## Repro case

```js
'use strict'

const restify = require('restify')
const server = restify.createServer()

server.use(function (req, res, next) {
  console.assert(this === server, 'function context is server instance')
  next()
})

server.get('/', function (req, res, next) {
  res.send('hello')
  next()
})

server.listen(3000)
```

## Cause

`handler` isn't called with server instance  in the `call` function in `chain.js` (`Chain` would have to be instantiated with a server instance passed as an option)
https://github.com/restify/node-restify/blob/master/lib/chain.js#L160
https://github.com/restify/node-restify/blob/master/lib/chain.js#L164

Or else `self` should be bound to `handler` in the `use` method:
https://github.com/restify/node-restify/blob/master/lib/server.js#L506

## Are you willing and able to fix this?

Yes

davidmarkclements:
tangentially related to https://github.com/restify/node-restify/issues/1595

hekike:
Thanks for reporting it!
Was it an expected and documented behavior before `v7.x`? cc @DonutEspresso @retrohacker

DonutEspresso:
Handlers were bound to the server in older versions, though in retrospect I don't think its ever been "documented" as part of the public API. It appears to have been an unofficial part of the contract, however, as it's come up many times in different issues and pull requests (and acknowledged by us as well). 

I think we have two choices here:
1) rebind the server context
2) document breaking change for 7.x

I don't have a strong preference here, but if there no penalty for binding (or maybe it's an option?), let's go for it.

leodutra:
Time for benchmarking .call() and .apply()

I remember .bind() is more costful than these.

Em sex, 6 de abr de 2018 21:56, Alex Liu <notifications@github.com>
escreveu:

DonutEspresso:
[...]

PR #1668, opened 2018-06-01
Fix: Bind server to middleware context

retrohacker:
Fixes #GH-1639

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -470,8 +470,11 @@ Server.prototype.pre = function pre() {
     var handlers = Array.prototype.slice.call(arguments);
 
     argumentsToChain(handlers).forEach(function forEach(handler) {
-        handler._name = handler.name || 'pre-' + self.preChain.count();
-        self.preChain.add(handler);
+        // GH-1639 Bind the server to the this context for the handler, keeping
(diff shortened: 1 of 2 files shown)


retrohacker:
cc:// @davidmarkclements

DonutEspresso:
Awesome, thanks for tackling this! Might want to run prettier to make sure styling is passing. 

I thought @hekike had a benchmark at some point, we could use that as a quick comparison of bind vs call/apply. With call/apply we wouldn't have to redefine the function name, but not sure how feasible that would be given that the router executes the chains now. Assuming a user land provided router is where we want to head I don't think that may not work either unless we decided to close over every handler.

retrohacker:
I tried using `call/apply` but ran into the problem of trying to get the `server` variable down into the chain executor. IIRC I had to pass it into 2 different constructors and a function in order to get it into scope. I could just attach it to the handler as a value (i.e. `handler.server = server`), its hacky but should work if we want to take that path. Then it would be `handler.apply(handler.server ...)`.

jrnail23:
is there a possibility of getting this PR merged?  I'm currently upgrading an app that depends on `this` in route handler chains being the server instance, and it's now `undefined` since installing v7.1.1.

PR #1669, opened 2018-06-02
Abstracted regex checker to a util function

ifiok:
## Issues

Closes:

* Edits required by @mattbishop from PR #1663 

# Changes

> The Regexp that validated the content type is used in two different places. Make it a const and share


lib/plugins/utils/isExtendedJson.js
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @public
+ * @function  validateExtendedJsonContentType
+ * @param   {String} contentType - the content type string to validate
+ * @returns {Boolean}       validity flag
+ */
+function validateExtendedJsonContentType(contentType) {
+    var type = contentType.toLowerCase();
+    // map any +json to application/json
+    var jsonPatternMatcher = new RegExp('^application/[a-zA-Z.]+\\+json');
```
mattbishop: Interesting to see this RegExp being rebuilt on every call to the function. I built a little test that ran this function with the RegExp declared outside the function and found it to be about 460% faster. Is there a reason to leave it declared in the function rather than as a module var?

mattbishop: Here is the test script if interested: ```
console.time('RegExp outside');
var jsonPatternMatcher = new RegExp('^application/[a-zA-Z.]+\\+json');
for (var i = 0; i < 100000; i++ ) {
  jsonPatternMatcher.test('application/json');
}
console.timeEnd('RegExp outside');

console.time('RegExp inside');
for (var i = 0; i < 100000; i++ ) {
  var jsonPatternMatcher = new RegExp('^application/[a-zA-Z.]+\\+json');
  jsonPatternMatcher.test('application/json');
}
console.timeEnd('RegExp inside');
```

PR #1670, opened 2018-06-04
chore: lift json regex to top of file

DonutEspresso:
Supersedes #1669


DonutEspresso:
@hekike do you know why `make fix-style` doesn't seem to be fixing the issues that are failing the build?  cc @retrohacker maybe that's why your PR was failing too.

lib/plugins/jsonBodyParser.js
```diff
@@ -7,6 +7,9 @@ var errors = require('restify-errors');
 
 var bodyReader = require('./bodyReader');
 
+///---local globals
+var jsonPatternMatcher = new RegExp('^application/[a-zA-Z.]+\\+json');
```
retrohacker: It looks like these two lines are the same, perhaps we should break it out into it's own file to avoid them falling out of sync?

I.E. `reqiure('jsonPatternMatcher')` and `module.exports = new RegExp...`?

I can see myself reviewing a PR in the future that modifies this line and not realizing that it exists multiple places in the codebase.

lib/plugins/bodyParser.js
```diff
@@ -192,6 +187,14 @@ function bodyParser(options) {
                 break;
         }
 
+        // if still no match, do more expensive regex matches. map any +json to
+        // application/json.
+        if (!parser) {
```
retrohacker: Nit/preference: the above seemed more obvious to me. This nested `if` required me to think a lot more about what was happening. I think it is handling both the explicit match and the regex as separate cases. Would it make sense to remove the `application/json` case above and assume that if `jsonPatternMatcher` matches here that `parser` should always be set to `parseJson`? This is similar to what the previous code did by coercing type into `application/json` if the regex matched.

DonutEspresso: Good question! The primary motivator for the PR was to avoid running/creating the regex when it wasn't necessary (since we were lifting the regex too to also avoid creating a new one for every request). The string comparisons are cheaper and likely to cover the 90% use case, so I wanted to reserve running the regex matching only when those cases failed. I expanded the comment with this context.

lib/plugins/jsonBodyParser.js
DonutEspresso: Done!

retrohacker:
https://github.com/restify/node-restify/pull/1670/commits/265955b1c5911a62c1dd11a8ffda61eb5b75cad1 appears to revert #1671

DonutEspresso:
Uh, that's weird, https://github.com/restify/node-restify/pull/1670/commits/265955b1c5911a62c1dd11a8ffda61eb5b75cad1 is created by just running `make fix-style`. 

I ran `make prepush` and all seems well. What happened earlier?

retrohacker:
This appears to be failing on CI as well. Maybe whack your local `node_modules` directory and do a fresh install to get the latest eslint?

retrohacker:
Not sure why your local version and Travis are out of sync using the same eslint config... Possibly different versions of eslint installed?

DonutEspresso:
Foiled by package-lock.json *shakes fist*.  Weird that semver compatible versions of prettier would generate such different output. We should start deleting package-lock/yarn.lock in `make clean` or otherwise..

DonutEspresso:
Cool, CI looks good now!

retrohacker:
Looks good to me if you want to merge.

PR #1671, opened 2018-06-05
style: fix failing check-style

retrohacker:
It looks like master's `make prepush` fails with a style issue. This PR fixes that and unblocks #1668

### DIFF
--- a/test/plugins/jsonBodyParser.test.js
+++ b/test/plugins/jsonBodyParser.test.js
@@ -257,9 +257,12 @@ describe('JSON body parser', function() {
             '\r\n' +
             '{"foo":"bar"}';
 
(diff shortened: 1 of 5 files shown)


retrohacker:
Not sure how we managed to merge something that broke the master branch? Or did an eslint update change styles out from under us? Either way, this should fix things :smile:

hekike:
@retrohacker we have seen this earlier with prettier that why I removed carrots from package.json from prettier related packages.

retrohacker:
Failing Node.js v10 tests, unrelated to this PR, merging.

retrohacker:
Actually looks like this might stomp on #1672 if you want to just fix the style issues with that PR?

PR #1672, opened 2018-06-05
fix(server): address Node v10.x req close event firing order

hekike:
Req close always fires in Node 10 and fires in a process.nextTick
This PR handles the close event in restify in a way that we only return an error if the response is not flushed yet.

Related changes in Node:

- https://github.com/nodejs/node/pull/20611
- https://github.com/nodejs/node/pull/20941

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -1155,6 +1155,8 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
         var err = new errors.RequestCloseError();
         err.statusCode = 444;
 
+        // For backward compatibility we only set connection state to closed
+        // for RequestCloseError
         req._connectionState = 'close';
 
         // Set status code and _sent for audit
--- a/lib/server.js
+++ b/lib/server.js
@@ -1091,7 +1091,7 @@ Server.prototype._onHandlerError = function _onHandlerError(err, req, res) {
         return;
     }
 
-    // Handlers doesn't continue when error happen
+    // Handlers don't continue when error happen
     res._handlersFinished = true;
 
     // Preserve handler err for finish event
@@ -1178,7 +1178,7 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
         req._timeUseEnd = req._timeUseEnd || processHrTime;
         req._timeRouteEnd = req._timeRouteEnd || processHrTime;
 
-        // In Node < 10 close didn't fire always
+        // In Node < 10 "close" event dont fire always
         // https://github.com/nodejs/node/pull/20611
         self._finishReqResCycle(req, res);
     }
@@ -1187,7 +1187,7 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
     function onResClose() {
         res._flushed = true;
 
-        // Finish may already set flushed timer
+        // Finish may already set the req._timeFlushed
         req._timeFlushed = req._timeFlushed || process.hrtime();
 
         self._finishReqResCycle(req, res, res.err);
--- a/lib/server.js
+++ b/lib/server.js
@@ -1139,24 +1139,18 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
     }
 
     // Request lifecycle events
-    // attach a listener for 'close' events, this will let us set
-    // a flag so that we can stop processing the request if the client closes
+    // attach a listener for 'aborted' events, this will let us set
+    // a flag so that we can stop processing the request if the client aborts
     // the connection (or we lose the connection).
     // we consider a closed request as flushed from metrics point of view
-    function onReqClose() {
-        // Do not handle close if response is already sent (may not flushed yet)
-        // In Node >= 10.4.0 res's close event is always emitted
-        // Introduced in: https://github.com/nodejs/node/pull/20611
-        if (res._sent) {
-            return;
-        }
-
-        // Request was aborted or closed. Override the status code
+    function onReqAborted() {
+        // Request was aborted, override the status code
         var err = new errors.RequestCloseError();
         err.statusCode = 444;
 
-        // For backward compatibility we only set connection state to closed
-        // for RequestCloseError
+        // For backward compatibility we only set connection state to "close"
+        // for RequestCloseError, also aborted is always immediatly followed
+        // by a "close" event
         req._connectionState = 'close';
 
         // Set status code and _sent for audit
@@ -1181,15 +1175,18 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
 
         self._finishReqResCycle(req, res);
     }
+
+    // Emitted when the response has been sent.
     function onResClose() {
         res._flushed = true;
-        req._timeFlushed = process.hrtime();
+        // Finish may alread set flushed timer
+        req._timeFlushed = req._timeFlushed || process.hrtime();
 
         self._finishReqResCycle(req, res, res.err);
     }
 
     // Request events
-    req.once('close', onReqClose);
+    req.once('aborted', onReqAborted);
 
     // Response events
     res.once('finish', onResFinish);
--- a/lib/server.js
+++ b/lib/server.js
@@ -1151,18 +1151,15 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
 
         // For backward compatibility we only set connection state to "close"
         // for RequestCloseError, also aborted is always immediatly followed
-        // by a "close" event so we don't change connection state here
+        // by a "close" event.
+        // We don't set _connectionState to "close" in the happy path
+        req._connectionState = 'close';
 
         // Set status code and _sent for audit
         // as req is already closed connection
         res.statusCode = err.statusCode;
         res._sent = true;
         res.err = err;
-    }
-
-    // Indicates that the underlying connection was closed
-    function onReqClose() {
-        req._connectionState = 'close';
     }
 
     // Response lifecycle events
@@ -1183,7 +1180,6 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
         self._finishReqResCycle(req, res);
     }
 
-    // Emitted when the response has been sent.
     function onResClose() {
         res._flushed = true;
 
@@ -1194,7 +1190,6 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
     }
 
     // Request events
-    req.once('close', onReqClose);
     req.once('aborted', onReqAborted);
 
     // Response events
--- a/lib/server.js
+++ b/lib/server.js
@@ -1148,8 +1148,8 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
 
         req._connectionState = 'close';
 
-        // Do no handle close if request is already flushed
-        // In Node >= 10 res's close event is always emitted
+        // Do not handle close if response is already flushed
+        // In Node >= 10.4.0 res's close event is always emitted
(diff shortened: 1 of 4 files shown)


lib/server.js
```diff
@@ -1143,19 +1143,37 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
...
     function onClose() {
-        res._flushed = true;
-        req._timeFlushed = process.hrtime();
-
         res.removeListener('finish', onFinish);
         res.removeListener('error', onError);
+
         req._connectionState = 'close';
 
+        // Do no handle close if request is already flushed
```
misterdjules: Typo "Do not"?

misterdjules: Also, is it "if response is already flushed" here, or do we mean "request" in the sense of "the whole request/response"? I think it'd help to distinguish between request and response here for clarity, and in this case it seems we're checking if the _response_ is flushed.

lib/server.js
```diff
@@ -1143,19 +1143,37 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
...
+        // In Node >= 10 res's close event is always emitted
```
misterdjules: Is is Node > 10.4.0 at least? It seems that change hasn't been made it to any release yet.

lib/server.js
```diff
@@ -1143,19 +1143,37 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
...
+            res,
+            res.err || new errors.RequestCloseError()
+        );
     }
-    req.once('close', onClose);
+
+    // Handle res's close on to the next cycle
+    // Required in Node >= 10:
+    // introduced in: https://github.com/nodejs/node/pull/20941
+    req.once('close', function onReqClose() {
```
misterdjules: Is it that because _res_' `'close'` event can now be emitted using `process.nextTick` we want to handle _req_'s `'close'` event on the actual next tick of the libuv event loop so that we process req's `'close'` event _after_ `res`'s `'close'` event?

It seems this ordering could break if .e.g node started emitting the res' `'close'` event using `setImmediate`, which seems a bit brittle.

test/server.test.js
```diff
@@ -1924,8 +1924,12 @@ test("should emit 'after' on successful request", function(t) {
     });
 
     SERVER.get('/foobar', function(req, res, next) {
-        res.send('hello world');
-        next();
+        setTimeout(function() {
+            res.send('hello world');
+            setTimeout(function() {
+                next();
+            }, 500);
```
misterdjules: Why do we need any `setTimeout` call here?

hekike: We generate some "work" in the server to be sure that events are firing in the right order. In my previous fix tests passed with positive false.

misterdjules: What do you mean by "some work"? It seems it's not CPU-bound "work", so is it a way to make events on request/response objects to happen in a specific order or within a certain timeframe? If so, should we document what the test is actually testing?

And why not keeping the original test? Should be add a new test instead of replacing the existing one?

test/plugins/metrics.test.js
```diff
@@ -63,7 +62,11 @@ describe('request metrics plugin', function() {
                     assert.isAtLeast(metrics.latency, 150);
                     assert.isAtLeast(metrics.totalLatency, 150);
                     assert.equal(metrics.path, '/foo');
-                    assert.equal(metrics.connectionState, undefined);
+                    // close in Node >= 10
+                    assert.include(
+                        [undefined, 'close'],
```
misterdjules: What does that `'[undefined, 'close']` mean?

hekike: Any of them. Undefined is the unknown connection state basically.

hekike:
Excellent idea, implemented!

lib/server.js
```diff
@@ -1170,22 +1173,24 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
...
-        // Do not trigger false
         self._finishReqResCycle(req, res);
     }
-    function onError(err) {
+
+    // Emitted when the response has been sent.
+    function onResClose() {
         res._flushed = true;
-        req._timeFlushed = process.hrtime();
+        // Finish may alread set flushed timer
```
misterdjules: nit: typo "already".

lib/server.js
```diff
@@ -1170,22 +1173,24 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
...
 
-        req.removeListener('close', onClose);
-        res.removeListener('error', onError);
-
-        // Do not trigger false
         self._finishReqResCycle(req, res);
     }
-    function onError(err) {
+
+    // Emitted when the response has been sent.
```
DonutEspresso: Is this still true? Also, I think I know the answer, but is it possible for close to get fired but `_flushed` is false?

lib/server.js
```diff
@@ -1170,22 +1173,24 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
         req._timeUseEnd = req._timeUseEnd || processHrTime;
         req._timeRouteEnd = req._timeRouteEnd || processHrTime;
 
-        req.removeListener('close', onClose);
-        res.removeListener('error', onError);
-
-        // Do not trigger false
         self._finishReqResCycle(req, res);
```
misterdjules: Why do we need to call `_finishReqResCycle` here?

lib/server.js
```diff
@@ -1170,22 +1173,24 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
...
+    function onResClose() {
         res._flushed = true;
-        req._timeFlushed = process.hrtime();
+        // Finish may alread set flushed timer
+        req._timeFlushed = req._timeFlushed || process.hrtime();
 
-        req.removeListener('close', onClose);
-        res.removeListener('finish', onFinish);
-        self._finishReqResCycle(req, res, err);
+        self._finishReqResCycle(req, res, res.err);
```
misterdjules: Why do we need to call `_finishReqResCycle` here?

test/plugins/metrics.test.js
misterdjules: What are the cases when the connection state would be `'close'`? Shouldn't it be `'close'` only when the request was aborted? In this case it doesn't seem like the request would be aborted.

test/plugins/metrics.test.js
```diff
@@ -103,6 +106,7 @@ describe('request metrics plugin', function() {
     it('should return metrics with pre error', function(done) {
         SERVER.on('uncaughtException', function(req, res, route, err) {
             assert.ok(err);
+            res.send(err);
```
misterdjules: Is there any exception thrown in this test? It doesn't seem so, it seems there's only an error being passed to a handler's `next` callback? If so, why does the server emit `'uncaughtException'`?

Also, why do we need to `res.send()` now when we wouldn't need to do it before?

test/plugins/metrics.test.js
```diff
@@ -139,6 +143,7 @@ describe('request metrics plugin', function() {
     it('should return metrics with use error', function(done) {
         SERVER.on('uncaughtException', function(req, res, route, err) {
             assert.ok(err);
+            res.send(err);
```
misterdjules: Same question regarding `'uncaughtException'` and the need for `res.send()` than above.

lib/server.js
```diff
@@ -1091,7 +1091,8 @@ Server.prototype._onHandlerError = function _onHandlerError(err, req, res) {
         return;
     }
 
-    res._handlersFinished = true;
+    // Handlers doesn't continue when error happen
+    self._onHandlerStop(req, res);
```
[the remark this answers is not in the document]
hekike: reverted

test/plugins/metrics.test.js
```diff
@@ -271,7 +283,11 @@ describe('request metrics plugin', function() {
                     assert.isNumber(metrics.latency, 200);
                     assert.equal(metrics.path, '/foo');
                     assert.equal(metrics.method, 'GET');
-                    assert.equal(metrics.connectionState, undefined);
+                    // close in Node >= 10
+                    assert.include(
+                        [undefined, 'close'],
```
misterdjules: Why is there two possible values here? Shouldn't we always expect `'close'`?

test/plugins/metrics.test.js
```diff
@@ -260,6 +271,7 @@ describe('request metrics plugin', function() {
                 {
                     server: SERVER
                 },
+                // TODO: test timeouts if any of the following asserts fails
```
misterdjules: Should we file a corresponding issue?

lib/server.js
hekike: To tigger after (after is triggered after both handlers are finished and req is flushed)

misterdjules:
@hekike Added a bunch more comments, but I think this is going in the right direction, thank you very much again for considering my feedback!

test/plugins/metrics.test.js
hekike: > f you listen to this event, you must send a response to the client. This behavior is different from the standard error events. If you do not listen to this event, restify’s default behavior is to call res.send() with the error that was thrown. - docs

So it should have been there

test/plugins/metrics.test.js
hekike: see comment above

test/plugins/metrics.test.js
```diff
@@ -63,7 +62,11 @@ describe('request metrics plugin', function() {
                     assert.isAtLeast(metrics.latency, 150);
                     assert.isAtLeast(metrics.totalLatency, 150);
                     assert.equal(metrics.path, '/foo');
-                    assert.equal(metrics.connectionState, undefined);
+                    // close doesn't always fire in Node < 10
+                    assert.include(
+                        [undefined, 'close'],
```
misterdjules: The problem with using `assert.include` here is that if we run this test with Node.js >= 10.x and we expect the connection state to be `'close'`, the test would pass even if the connection state was `undefined`. Would it be too much work to differentiate between node >= 10.x and < 10.x here and have two different asserts, something like:

```
if (process.version >= 10.0) {
  assert.equal(metrics.connectionState, 'close');
} else {
  assert.equal(metrics.connectionState, undefined);
}
```
?

Same for all other connection state assertions in this file.

test/plugins/metrics.test.js
```diff
@@ -63,6 +63,7 @@ describe('request metrics plugin', function() {
                     assert.isAtLeast(metrics.latency, 150);
                     assert.isAtLeast(metrics.totalLatency, 150);
                     assert.equal(metrics.path, '/foo');
+                    // close doesn't always fire in Node < 10
```
misterdjules: Is that comment still relevant?

test/plugins/metrics.test.js
```diff
@@ -271,6 +281,7 @@ describe('request metrics plugin', function() {
                     assert.isNumber(metrics.latency, 200);
                     assert.equal(metrics.path, '/foo');
                     assert.equal(metrics.method, 'GET');
+                    // close doesn't always fire in Node < 10
```
misterdjules: Is that comment still relevant?

test/server.test.js
misterdjules: Remaining questions above still relevant:

1. Should we document _why_ we want to control the timing of those operations in this test?
2. Should we also keep the original test and add a new test with those changes?

test/plugins/metrics.test.js
hekike: yes

hekike:
Thanks for re-reviewing, restify just get better ;)

PR #1673, opened 2018-06-06
chore: new makefile. fix hanging tests

DonutEspresso:
Opinionated clean up of Makefile. Add code coverage task via nyc. Fix hanging tests.


Makefile
```diff
@@ -1,79 +1,130 @@
 #
-# Copyright (c) 2012, Joyent, Inc. All rights reserved.
+# Directories
```
retrohacker: Hooray for flattening all the makefiles!

Makefile
```diff
@@ -1,79 +1,130 @@
...
-	$(NPM) install
 
-.PHONY: cover
-cover: $(NODECOVER)
-	@rm -fr ./.coverage_data
-	$(NODECOVER) run $(NODEUNIT) ./test/*.js
-	$(NODECOVER) report html
+.PHONY: help
+help:
+	@perl -nle'print $& if m{^[a-zA-Z_-]+:.*?## .*$$}' $(MAKEFILE_LIST) \
```
retrohacker: Where does `MAKEFILE_LIST` come from?

Makefile
```diff
@@ -1,79 +1,130 @@
...
 .PHONY: benchmark
-benchmark:
+benchmark: $(NODE_MODULES)
 	@(cd ./benchmark && $(NPM) i && $(NODE) index.js)
 
-include ./tools/mk/Makefile.deps
-include ./tools/mk/Makefile.targ
+
+.PHONY: prepush
+prepush: $(NODE_MODULES) lint style test ## Git pre-push hook task. Run before committing and pushing.
```
retrohacker: Should this be `test` or `coverage`?

Makefile
DonutEspresso: It's a special variable that contains the names of the makefiles loaded:
https://www.gnu.org/software/make/manual/make.html#Special-Variables

Makefile
DonutEspresso: I have it as test, since the coverage target is specified as a post build hook in .travis.yml. We could run it locally too though - don't feel strongly about it.

DonutEspresso:
It seems yarn uncovers a problem that was going unnoticed:

```sh
$ yarn
yarn install v1.7.0
info No lockfile found.
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
[3/5] 🚚  Fetching packages...
error shallow-clone@3.0.0: The engine "node" is incompatible with this module. Expected version ">=8".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
$ npm ls shallow-clone
restify@7.2.1 /Users/aliu/Sandbox/restify
└─┬ autocannon@2.4.1
  └── shallow-clone@3.0.0
```

It looks like npm ignores the engine check - it doesn't even appear to produce a warning of any sort. Yarn on the other hand [chooses to fail loudly](https://github.com/yarnpkg/yarn/issues/1102), which seems to be the correct behavior. There's nothing in shallow-clone that seems to require node8:
https://github.com/jonschlinkert/shallow-clone/blob/master/index.js

Do we have an alternative for autocannon?

DonutEspresso:
[Relevant npm issue](https://github.com/npm/npm/issues/17778).

hekike:
Not really :/
Autocannon is the main module of the benchmark suite, but it's a dev dependency and not used by CI.
Maybe we can move it as a dependency of `benchmark/` directory, we already have a `package.json` there.

hekike:
Just be sure `make benchmark` installs it.

Issue #1666, opened 2018-05-28
Unclear documentation about Server config


Junkern:
The docs on http://restify.com/docs/server-api/ are not consistent. 

There is one section for `createServer` and one section for `Server`. Both are describe to `create a server`, but their documentation states different options. E.g. for `Server` there is a `noWriteContinue` which is not present for `createServer`.

Also `options.httpsServerOptions`
> Any options accepted by node-https Server. If provided the following restify server options will be ignored: spdy, ca, certificate, key, passphrase, rejectUnauthorized, requestCert and ciphers; however these can all be specified on httpsServerOptions. 

So can I also pass these httpsServerOptions directly to the restify server? If so, that should be documented as `options`

Moreover, according to the docs you still can pass an regexp to e.g. `.get()`. `Server~methodOpts` is listed as accepting `Regexp`.

pkyeck:
also found this:

> Versioned Routes
[...] You can default the versions on routes by passing in a version field at server creation time. [...]

But this is not mentioned in the docs - neither here (http://restify.com/docs/server-api/#createserver) nor here (http://restify.com/docs/server-api/#server).

Is this still possible? Or was this removed at some point?

DonutEspresso:
@Junkern thanks for filing this - `createServer` and `Server` should be identical for all intents and purposes. We started doing automatic API generation at some point so this could be a result of new auto generated stuff clashing with existing API docs. cc @hekike 

@pkyeck it seems the versioned routing also seems to be missing from the verb APIs, which is where they live now: http://restify.com/docs/6to7/

We definitely need to add those back in..

hekike:
I created a PR to fix this: https://github.com/restify/node-restify/pull/1678

hekike:
Closed by https://github.com/restify/node-restify/pull/1678

PR #1678, opened 2018-06-10
docs(index): sync createServer and Server constructor docs

hekike:
- Sync `createServer` and `Server` constructor docs
- update seever.md route versioning part

# Closes

https://github.com/restify/node-restify/issues/1666


DonutEspresso:
Thanks this looks great! I wonder if at some point we should remove the options from the Server class at some point in the future, since folks never new up a server directly?

hekike:
You are right, `Server` is not exported, we should hide it from docs. Unfortunately `@hideconstructor` JSDoc decorator doesn't work in `Document.js` yet.

I filed an issue: https://github.com/bitovi/documentjs/issues/285

hekike:
https://github.com/restify/node-restify/issues/1680

Issue #1644, opened 2018-04-10
Documentation regarding regex paths is out of date


thisandagain:
# Bug Report

## Restify Version
`7.1.0`

## Node.js Version
`8.11.1`

## Expected behaviour
Restify should recognize regex literals when defining a path as per documentation:

> You can also pass in a RegExp object and access the capture group with req.params (which will not be interpreted in any way):
```js
server.get(/^\/([a-zA-Z0-9_\.~-]+)\/(.*)/, function(req, res, next) {
  console.log(req.params[0]);
  console.log(req.params[1]);
  res.send(200);
  return next();
});
```

## Actual behaviour
```
[Redacted]/node_modules/assert-plus/assert.js:22
    throw new assert.AssertionError({
    ^

AssertionError [ERR_ASSERTION]: path (string) is required
    at RouterRegistryRadix.add ([Redacted]/node_modules/restify/lib/routerRegistryRadix.js:39:12)
    at Router.mount ([Redacted]/node_modules/restify/lib/router.js:167:20)
    at Server.serverMethod [as get] ([Redacted]/node_modules/restify/lib/server.js:1476:33)
    at Object.<anonymous> ([Redacted]/test.js:4:8)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Function.Module.runMain (module.js:693:10)
```

## Repro case
```js
const restify = require('restify');

const server = restify.createServer();
server.get(/^\/([a-zA-Z0-9_\.~-]+)\/(.*)/, function(req, res, next) {
  console.log(req.params[0]);
  console.log(req.params[1]);
  res.send(200);
  return next();
});
server.listen(8000);
```

## Cause

I believe this may be the intended behavior as per changes in [this commit](https://github.com/restify/node-restify/commit/82832771826321480e5e524db258668f62b689c2#diff-504fd89f344be1e9c4b583f3ebd7428a) and documented [here](https://github.com/restify/node-restify/blob/master/docs/guides/6to7guide.md#different-regexp-usage-in-router-path-and-wildcards).

## Are you willing and able to fix this?

Yes, but I have the following questions:
- Was breaking the example case intentional?
- If so: given what the example in the existing documentation is trying to show does it seem reasonable to replace it with this example from the "6 to 7" migration guide?

```js
[...]

PR #1679, opened 2018-06-10
fix(server): remove regexp check from method factory

hekike:
Closes

https://github.com/restify/node-restify/issues/1644


lib/server.js
```diff
@@ -1484,8 +1486,24 @@ function errEvtNameFromError(err) {
...
+     * @example
+     * server.get('/example/near/:lat-:lng/radius/:r', (req, res, next) => {});
+     * server.get({ path: '/foo' }, (req, res, next) => {});
+     * server.get({ url: '/foo' }, (req, res, next) => {});
+     */
     return function serverMethod(opts) {
-        if (opts instanceof RegExp || typeof opts === 'string') {
+        // RegExp deprecation message
+        if (opts instanceof RegExp) {
+            new TypeError(REGEXP_PATH_DEPRECATED_TEXT);
```
DonutEspresso: Is it deprecated or completely broken? Did you mean to throw this error?

hekike: It's completely removed in 7x.

DonutEspresso: Cool, then shouldn't we throw here then?

hekike: oops, that was the plan :D

hekike:
@DonutEspresso I addressed your comments.

Issue #1601, opened 2018-02-07
Support altering audit log serializers


ekristen:
# Feature Request

## Use Case

I want to be able to use custom serializers to sanitize sensitive fields. For example, using a hmac 256 hash of a parameter in a query string, or sanitizing the authorization header using a hmac256 hash.

## Are you willing and able to implement this?

I'm willing to implement. I think the implementation is pretty straight forward, allow a custom serializer to be passed in for the req and res objects and basically if they are functions call them last before returning the custom version that already is present.

retrohacker:
Hey @ekristen!

Thank you for taking the time to open this feature request, and for volunteering to implement this in the codebase :heart:

I believe what you are trying to accomplish is already possible through Bunyan serializers.
https://www.npmjs.com/package/bunyan#serializers

Example:

```js
var logger =  bunyan.createLogger({
  name: 'audit',
  stream: process.stdout
});
logger.addSerializers({
  req: function (req) {
    return { foo: req.foo, bar: hmac(req.bar) };
  }
});
server.on('after', restify.plugins.auditLogger({
  log: logger,
  server: server,
  logMetrics : logBuffer,
  printLog : true
}));
```

ekristen:
I'll try your pseudo code out but I'm fairly positive that will not work, I'd be happy to be wrong.

Please see https://github.com/restify/node-restify/blob/master/lib/plugins/audit.js#L210-L238

The audit logger creates a child and then specifically sets req and res serializers, but does allow for a custom error serializer.

retrohacker:
Oh interesting, yeah I'm not sure if they are chainable or if the child's serializer displaces the parent's. Interested to hear what you find!

ekristen:
AFAICT we'll have to modify the code to allow for a custom serializers to be passed to the auditLogger as it has a hardcoded default one.

OrthoDex:
Hi! Can this issue be readdressed? I have the same use case.

retrohacker:
I believe this is being addressed in #1746

PR #1681, opened 2018-06-12
feat(audit): Support overriding default serializers

mattgoucher:
## Pre-Submission Checklist

- [x] Opened an issue discussing these changes before opening the PR **(existing issue)**

## Issues

Closes:

* Issue #1601

> Similar to @ekristen, I have a use-case to override the existing req/res serializers that are passed by the Restify Audit plugin. As discussed in the issue #1601, when a child logger is derrived from the passed in logger, the serializers of the parent are overwritten by the child's.

# Changes

> This PR gives consumers the ability to pass their own serializers for req, res and err objects.

### DIFF
--- a/lib/plugins/audit.js
+++ b/lib/plugins/audit.js
@@ -237,7 +237,7 @@ function auditLogger(opts) {
      * @private
      * @function auditResponseSerializer
      * @param {http.ServerResponse} res - outgoing server response
-     * @returns {object} Serialized request
+     * @returns {object} Serialized response
      */
     function auditResponseSerializer(res) {
         if (!res) {
--- a/lib/plugins/audit.js
+++ b/lib/plugins/audit.js
@@ -196,10 +196,81 @@ function auditLogger(opts) {
     if (typeof printLog === 'undefined') {
         printLog = true;
     }
-    var errSerializer = bunyan.stdSerializers.err;
 
-    if (opts.log.serializers && opts.log.serializers.err) {
-        errSerializer = opts.log.serializers.err;
+    /**
+     * Serialize the HTTP Request into a plain object, for logging
+     *
+     * @private
+     * @function auditRequestSerializer
+     * @param {http.ClientRequest} req - Incoming request
+     * @returns {object} Serialized request
+     */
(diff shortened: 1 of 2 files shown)


lib/plugins/audit.js
```diff
@@ -196,68 +196,90 @@ function auditLogger(opts) {
...
+    }
+
     var errSerializer = bunyan.stdSerializers.err;
+    var reqSerializer = auditRequestSerializer;
+    var resSerializer = auditResponseSerializer;
 
-    if (opts.log.serializers && opts.log.serializers.err) {
-        errSerializer = opts.log.serializers.err;
+    // Allow override of serializers
+    if (opts.log.serializers) {
```
DonutEspresso: I know this is how we were doing it before, but I wonder if we can be more explicit here, i.e., pass in opts.serializers as a top level option, rather than implicitly depending on any serializers on the existing logger. What do you think @hekike? If we went that route we might have to keep the old L201 line in there though for backwards compatibility.

DonutEspresso:
Left a minor comment about how these are passed in, but everything else looks reasonable.

lib/plugins/audit.js
mattgoucher: If my understanding is correct, you're thinking that we would use the entire serializers object instead of picking off three (req, res, err) and using those. This would allow consumers the flexibility of owning the serializers without having to keep adding conditional logic in this code.

I prefer this approach. However, to be fully backward compatible, we'll probably end up with the same conditional logic to ensure that the Restify default serializers still run if they're not explicitly passed by consumers.

Happy to make the change in either direction.

DonutEspresso: Basically, I am thinking of an explicit serializers option:

```js
SERVER.once('after', restify.plugins.auditLogger({
    serializers: {...}
});
```

Then within the audit logger we'd pull serializers from a top level `opts.serializers` instead of from `opts.log.serializers`.  I believe the fallback logic you have for picking between a passed in serializer and the default one is still applicable. This just makes serializers a bit more explicit since we're creating a new child logger here anyway.

PR #1683, opened 2018-06-19
breaking: restify-errors@6.x

DonutEspresso:
The primary change in restify-errors@6.x is that all errors generated by restify and plugins (BadRequest, InternalServer, GatewayTimeout, etc) now derive from VError base classes now. Folks interact with these errors via the event listeners, like the server's `after` event, or the `uncaughtException` event. 

While the API provided by the Error instances [is not breaking](https://github.com/restify/errors#migration-from-5x-to-6x), any class or prototype interrogation will definitely fail because the base classes have changed.

### DIFF
--- a/lib/errorTypes.js
+++ b/lib/errorTypes.js
@@ -2,5 +2,7 @@
 
 var errors = require('restify-errors');
 
-errors.makeConstructor('RequestCloseError');
-errors.makeConstructor('RouteMissingError');
+module.exports = {
+    RequestCloseError: errors.makeConstructor('RequestCloseError'),
+    RouteMissingError: errors.makeConstructor('RouteMissingError')
+};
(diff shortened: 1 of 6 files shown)


misterdjules:
> While the API provided by the Error instances is not breaking, any class or prototype interrogation will definitely fail because the base classes have changed.

Can it be considered not breaking then? Also, this PR is prefixed with `breaking:`, so I'm a bit confused :)

DonutEspresso:
I lean towards not breaking as all the consumable APIs aren't changing - but some folks are more conservative than others. Do we need to put up a vote? :)

misterdjules:
@DonutEspresso What is the cost of considering this breaking? Is it that we'd need to go throigh a major release process (which I imagine is more involved than a non-major release?) and update the dependencies of a significant number of consumers?

misterdjules:
@DonutEspresso Am I blocking this to land? I don't want to get in the way, so should I go ahead and just review the code changes without considering the semverness of the changes?

PR #1687, opened 2018-07-17
Document change from req.route to req.route.spec

karthikiyengar:
This PR documents a breaking change in newer Restify versions. 

Many packages and middleware depend on `req.route`, and this has been changed to `req.route.spec` since v7.


ValYouW:
Too bad this PR has not been merged, just had some failovers because of this change, having it documented could have helped...

Issue #549, opened 2014-03-05
NotAuthorized when serving static files.


trevorstarick:
I'm trying to serve a static index.html file and I think my code is correct but when I go to `http://localhost/static` I get:
 ```
{
  code: "NotAuthorized",
  message: "/static"
}
``` 

Is it an issue with the current changes that have been made to the static file serving JS or is it an issue with my code?

Route:
```
app.get(/\/static\/?.*/, restify.serveStatic({ 
    directory: "./",
    default: "index.html"
}));
```

phanimahesh:
I can confirm that it is impossible to not hit a NotAuthorizedError when trying to serve static files. Explanation same as above.

playeren:
I got this to work (inspired by @zepheiryan ) by changing this [line](https://github.com/restify/node-restify/blob/master/lib/plugins/static.js#L115) to `var file = './' + path.join(opts.directory,`.

node-restify v0.12.7 (via npm)
Debian 8
Node v0.12.7

CocaCola183:
I got the same issue, too.
I fixed it by changing `directory: ''.` to  `directory: __dirname,`.
node-restify ^4.0.2
Node 4.0.0

yerol:
I was having this exact issue with no auth checks in my code. 
Switching from `directory: ''` to `directory: __dirname` fixed the problem for me too.

On Windows 10
Node 5.8.0
Restify 4.0.4

PR #1692, opened 2018-08-03
fix NotAuthorized error when serving static files while directory var…

unsol:
## Pre-Submission Checklist

## Issues

Closes:

* Issue 549

# Changes

This is to resolve issue #549, so I did not create a new issue for this pull request.

The serveStatic function sets the appendRequestPath option to true by default. This works the majority of the time, but if the directory option is set to "./", then the regex test ("re.test(file.replace(/\\/g, '/'))" on line 194 -- https://github.com/restify/node-restify/blob/e663699c8c5383eb6affad16ac2eb14fcd66a858/lib/plugins/static.js#L194) fails, which results in NotAuthorizedError. This test fails, as it is trying to compare a slash, against a full path. By moving the "p", and "re" variable declarations to the serve function, after the file path is joined with opts.directory, we are then able to normalize the path, run that output through escapeRE, and test the expected expression.

The "p", and "re" variables are not used anywhere else in this file -- only within the serve function.

I tested this solution with the following routes:

        server.get("/firsttest", restify.plugins.serveStatic({
            directory: "./app",
            file: "firsttest.html",
            appendRequestPath: false,
        }));

        server.get("/secondtest", restify.plugins.serveStatic({
            directory: "./app",
            file: "secondtest().html",
            appendRequestPath: false,
        }));

        server.get("/docs/*", restify.plugins.serveStatic({
            directory: "./",
            default: "index.html",
            appendRequestPath: true,
        }));

My folder structure for the tests was something along the lines of:
  /
    docs/
      v1/
        index.html
      v2/
        index.html
    app/
      firsttest.html
      secondtest().html
      (contains source code)
    dist/
      contains typescript build output

I had the first/second test routes set to serve files from the ./app directory, as they just didn't want to serve from there (kept getting a not found event). I'm assuming that has something to do with the wild card route taking precedence for some reason.

### DIFF
--- a/lib/plugins/static.js
+++ b/lib/plugins/static.js
@@ -89,9 +89,6 @@ function serveStatic(options) {
     assert.optionalString(opts.file, 'options.file');
     assert.bool(opts.appendRequestPath, 'options.appendRequestPath');
 
-    var p = path.normalize(opts.directory).replace(/\\/g, '/');
-    var re = new RegExp('^' + escapeRE(p) + '/?.*');
-
     function serveFileFromStats(file, err, stats, isGzip, req, res, next) {
         if (typeof req.closed === 'function' && req.closed()) {
             next(false);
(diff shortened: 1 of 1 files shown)


unsol:
Bad stale bot! Bad!

Has anyone had a chance to test this out? I've been using this patch in production since August, and have had no issues.

unsol:
Bumping it up again.

Still no issues with daily use.

unsol:
Been running on 7.4.0. Doesn't look like 7.5.0 will have any effect on the serveStatic changes.

PR #1695, opened 2018-08-10
Upgrade router

amitguptagwl:
## Pre-Submission Checklist

## Issues

Closes:

* Issue # 1694

As I had a discussion on linkedin, I'm raising this PR to upgrade router. 

# Changes

Functionality wise there is no change. This PR will just improve the router speed. And will enable restify to add new features of upgraded router.

## Impact
There is no customer impact as old and new both routers has the same API structure. However, there is one small change in defining regex for path parameters. If there are 2 consecutive path parameters then they should not be separated by '-' now.

I would be happy to know the result of performance tests.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -184,19 +184,6 @@ Router.prototype.unmount = function unmount(name) {
 };
 
 /**
- * toString() serialization.
- *
- * @public
- * @memberof Router
- * @instance
- * @function toString
- * @returns  {String} stringified router
- */
-/* Router.prototype.toString = function toString() {
-    return this._registry.toString();
-}; */
-
-/**
  * count()
  *
  * @public
--- a/lib/router.js
+++ b/lib/router.js
@@ -192,8 +192,21 @@ Router.prototype.unmount = function unmount(name) {
  * @function toString
  * @returns  {String} stringified router
  */
-Router.prototype.toString = function toString() {
+/* Router.prototype.toString = function toString() {
     return this._registry.toString();
+}; */
+
+/**
+ * count()
+ *
+ * @public
+ * @memberof Router
+ * @instance
+ * @function count
+ * @returns  {Number} count of registered routes
+ */
+Router.prototype.count = function count() {
+    return this._registry.count();
 };
 
 /**
--- a/lib/routerRegistryRadix.js
+++ b/lib/routerRegistryRadix.js
@@ -99,7 +99,7 @@ RouterRegistryRadix.prototype.lookup = function lookup(method, pathname) {
     var fmwRoute = this._anumargak.find(method, pathname);
 
     // Not found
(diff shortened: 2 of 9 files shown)


amitguptagwl:
Can you please guide me about failure tests?

amitguptagwl:
I've added some more features in Anumargak. But it'll not impact this PR. I would be happy to know your feedback on this PR, and if any necessary change has to be taken.

amitguptagwl:
Any update or feedback?

amitguptagwl:
[Here](https://github.com/node-muneem/anumargak/blob/master/benchmark/perf.js) is the script that I've used for benchmarking.

amitguptagwl:
Here is the list of features supported by Anumargak but not included in this PR;

1. Request parameters; Anumargak adds _path.url, _path.params, _queryStr, _hashStr to the request object.
2. You can register multiple routes to a series of handlers.
3.  You can register events: request, found, not found, and end.

amitguptagwl:
I'm fine to update the PR if it is required. And I'm fine to close it if is denied by the team.

lib/router.js
```diff
@@ -192,8 +192,21 @@ Router.prototype.unmount = function unmount(name) {
  * @function toString
  * @returns  {String} stringified router
  */
-Router.prototype.toString = function toString() {
+/* Router.prototype.toString = function toString() {
     return this._registry.toString();
+}; */
```
Goodluckhf: why is this commented?

lib/routerRegistryRadix.js
```diff
@@ -133,8 +133,21 @@ RouterRegistryRadix.prototype.get = function get() {
  * @function toString
  * @returns  {String} stringified RouterRegistryRadix
  */
-RouterRegistryRadix.prototype.toString = function toString() {
+/* RouterRegistryRadix.prototype.toString = function toString() {
     return this._findMyWay.prettyPrint();
+}; */
```
Goodluckhf: and this one too
Someone could use this serialization

Goodluckhf: For example some cli tools which provide scaffolding. Usually with such cli you can get registered routes in terminal.

But anyway it's better to remove code not to comment it

amitguptagwl: For display, yes we can add a function to the Anumargak. But for the time being, I've removed the comments.

amitguptagwl:
I've recently written an article comparing the latest version of both routers. And I believe, we don't need to migrate. Because we can't achieve the actual speed of a router in a framework. A framework has its own processing/limitations. Hence I'm closing this PR.

https://medium.com/@amitgupta.gwl/a-quick-guide-to-select-a-router-for-a-node-js-application-62c682e40afa

Issue #1709, opened 2018-10-26
[Restify 7] handleUncaughtExceptions incorrectly handles other types of errors


mridgway:
- [x] Used appropriate template for the issue type
- [x] Searched both open and closed issues for duplicates of this issue
- [x] Title adequately and _concisely_ reflects the feature or the bug

# Bug Report

## Restify Version

Restify 7.x

## Node.js Version

Node 8.11.3

## Expected behaviour

Turning on `handleUncaughtExceptions` should trigger `uncaughtException` handlers only for thrown exceptions that have not been caught.

## Actual behaviour

Other exceptions like `next(new Error())` or calling a method (like `HEAD`) that is not handled by the server trigger the `uncaughtExceptions` handlers.

## Repro case

```js
var restify = require('./');

var server = restify.createServer({
    handleUncaughtExceptions: true
});

server.on('uncaughtException', function exceptionHandler(req, res, route, err) {
    res.send(500, 'foo');
});

// Hitting this route incorrectly triggers the uncaughtException handler
server.get('/', function rootGet(req, res, next) {
    next(new Error('Test')); 
});

// Hitting this route correctly triggers the uncaughtException handler
server.get('/throw', function throwGet(req, res, next) {
    throw new Error('Thrown Error');
});

server.listen(3000);

// Hitting `HEAD /` incorrectly triggers the uncaughtException handler with `MethodNotAllowed` error
```

## Cause

The logic for handling errors has been unified between uncaught exceptions and other types of route handling errors, but information on whether an error was uncaught is not retained.

## Are you willing and able to fix this?

Yes, I have a branch ready at https://github.com/restify/node-restify/compare/master...mridgway:fixUncaughtExceptionTrigger

mridgway:
Fixed in https://github.com/restify/node-restify/pull/1710

PR #1710, opened 2018-10-26
Fix invalid triggers of the uncaughtException handler

mridgway:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #1709 

# Changes

The previous (prior to 7.x) functionality for uncaught exceptions only triggered the `uncaughtException` handler if it was truly an uncaught thrown error. Right now in Restify 7, if you turn on `handleUncaughtExceptions` and add a listener for the event, all errors from `next(err)` and internal errors like `MethodNotAllowed` will trigger the `uncaughtException` handler.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -89,8 +89,7 @@ function createServer(options) {
             e
         ) {
             if (
-                (opts.handleUncaughtExceptions &&
-                    this.listeners('uncaughtException').length > 1) ||
+                this.listeners('uncaughtException').length > 1 ||
                 res.headersSent
             ) {
                 return false;
--- a/lib/index.js
+++ b/lib/index.js
@@ -89,7 +89,8 @@ function createServer(options) {
             e
         ) {
             if (
-                this.listeners('uncaughtException').length > 1 ||
+                (opts.handleUncaughtExceptions &&
+                    this.listeners('uncaughtException').length > 1) ||
                 res.headersSent
             ) {
                 return false;
--- a/lib/server.js
+++ b/lib/server.js
@@ -816,7 +816,7 @@ Server.prototype._onRequest = function _onRequest(req, res) {
         handlerDomain.add(req);
         handlerDomain.add(res);
         handlerDomain.on('error', function onError(err) {
-            self._onHandlerError(err, req, res);
+            self._onHandlerError(err, req, res, true);
         });
         handlerDomain.run(function run() {
             self._runPre(req, res);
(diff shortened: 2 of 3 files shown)


lib/index.js
```diff
@@ -89,7 +89,8 @@ function createServer(options) {
             e
         ) {
             if (
-                this.listeners('uncaughtException').length > 1 ||
+                (opts.handleUncaughtExceptions &&
```
DonutEspresso: Isn't this already gated (`opts.handleUncaughtExceptions`) on L84?

test/server.test.js
```diff
@@ -674,6 +674,42 @@ test('GH-77 uncaughtException (default behavior)', function(t) {
     });
 });
 
+// eslint-disable-next-line
+test('handleUncaughtExceptions should not call handler for internal errors', function(t) {
+    SERVER.get('/', function(req, res, next) {
```
DonutEspresso: I'm a little bit confused about this test setup. I think we trying to verify restify/router errors (e.g., 404) don't trigger uncaught exceptions? If yes, should we just remove the get route and have the client do a get against a non-existent route?

test/server.test.js
```diff
@@ -674,6 +674,42 @@ test('GH-77 uncaughtException (default behavior)', function(t) {
...
+
+    CLIENT.head('/', function(err, _, res) {
+        t.ok(err);
+        t.equal(res.statusCode, 405);
+        t.end();
+    });
+});
+
+// eslint-disable-next-line
+test('handleUncaughtExceptions should not call handler for next(new Error())', function(t) {
```
DonutEspresso: :+1:

test/server.test.js
mridgway: There seems to be a behavior difference if you have no routes (which 404s) and having only GET routes and trying HEAD. I agree that this makes the test confusing, but I think if the behavior difference should be fixed, it should be fixed separately.

lib/index.js
mridgway: Yeah, it is indeed. I'll remove this check.

test/server.test.js
DonutEspresso: Gotcha - thanks for context! I expect this is related to CORS use cases - do you mind adding all of this as a comment to the test? Otherwise, it's a bit confusing to register the empty `/` that we actually don't expect to be run or triggered. Maybe we could do an `assert.fail('should not run')` or something inside the GET handler for clarity as well?

mridgway: Makes sense. I added a comment and the assertion that you mentioned.

lib/server.js
```diff
@@ -816,7 +816,7 @@ Server.prototype._onRequest = function _onRequest(req, res) {
         handlerDomain.add(req);
         handlerDomain.add(res);
         handlerDomain.on('error', function onError(err) {
-            self._onHandlerError(err, req, res);
+            self._onHandlerError(err, req, res, true);
```
misterdjules: I think this would still make errors _emitted_ on objects bound to the domain (in this case `req` and `res`) trigger an `uncaughtException` event. It seems that would also be a problem?

If we consider that is indeed not what the behavior of `handleUncaughtExceptions` should be, we should set the fourth parameter of `_onHandlerError` to `true` only if `err.domainThrown === true`.

@mridgway @DonutEspresso Thoughts?

mridgway: I believe that behavior is consistent with Restify versions <7. I'm not sure what the original intention was though.

I agree that it would make more sense to filter to only uncaught exceptions.

misterdjules: > I believe that behavior is consistent with Restify versions <7.

I just verified that looking at the 6.x branch of this repository and this is correct. Thus I'm fine with keeping that behavior, at least for now.

PR #1711, opened 2018-10-30
breaking: update travis targets

DonutEspresso:


DonutEspresso:
We're in a weird place right now - 10 isn't LTS yet (but almost!) so branches seem to be pointing to all the versions around it. Hardcoded all the relevant versions for now.

hekike:
10 is LTS now ;)

sean3z:
@DonutEspresso @hekike Any objections moving forward with this one? I suspect not but, want to confirm as this would imply "officially" dropping support for Node 4  🙂 and we'd presumably also want to update the engine target in [package.json](https://github.com/restify/node-restify/blob/master/package.json#L92-L94)

DonutEspresso:
No concerns here - though I think technically dropping 4 should be a breaking change. I'll rename the PR to reflect that.

rajatkumar:
This is addressed as part of restify v8 release. Closing this PR.

PR #1713, opened 2018-10-31
feat: support retrieving all req context

DonutEspresso:
# Changes

This adds `req.getAll()` to return the entirety of the request context data bucket. e.g.,

```
req.set('foo', 'bar');
req.set('baz', 'qux');
req.getAll(); // => { foo: bar, baz: qux }
```

Would `req.get()` with no params be preferable? `getAll()` seems more explicit, but the empty arg method seems like an established access pattern in other JS libraries. 

This all removes the private `req._getAllDebug()` method in favor of the public `getAll()` method.


test/plugins/plugins.test.js
```diff
@@ -120,137 +120,6 @@ describe('all other plugins', function() {
         });
     });
 
-    describe('context', function() {
```
DonutEspresso: We were spreading plugin tests across multiple files - just moved these into the existing context plugin test file.

lib/plugins/pre/context.js
```diff
@@ -48,8 +48,8 @@ function ctx() {
             return data[key];
         };
 
-        // private method which returns the entire context object
-        req._getAllContext = function _getAllContext() {
+        // returnthe entire context object
```
hekike: /s/returnthe/return the

PR #1715, opened 2018-11-02
Relax test

### LINKED ISSUE #1691: metrics.test.js seems to be flaky
- [x] Used appropriate template for the issue type
- [x] Searched both open and closed issues for duplicates of this issue
- [x] Title adequately and _concisely_ reflects the feature or the bug

# Bug Report

## Restify Version

v7.2.1, from commit e663699c8c5383eb6affad16ac2eb14fcd66a858 in master branch.

## Node.js Version

Various Node.js versions, including 4.x, 6.x up to 10.x.

## Expected behaviour

Tests in `metrics.test.js` should consistently pass.

## Actual behaviour

Tests in `metrics.test.js` sometimes fail with the following error:

```
/home/travis/build/restify/node-restify/node_modules/chai/lib/chai/interface/assert.js:2895
      throw(val);
      ^
AssertionError: expected 49 to be at least 50
    at /home/travis/build/restify/node-restify/test/plugins/metrics.test.js:206:28
    at Server.metrics (/home/travis/build/restify/node-restify/lib/plugins/metrics.js:92:16)
    at emitMany (events.js:108:13)
    at Server.emit (events.js:182:7)
    at Server._finishReqResCycle (/home/travis/build/restify/node-restify/lib/server.js:1241:14)
    at Server._afterRoute (/home/travis/build/restify/node-restify/lib/server.js:979:10)
    at Immediate.afterRouter (/home/travis/build/restify/node-restify/lib/server.js:949:18)
    at Immediate.immediate._onImmediate (timers.js:458:18)
    at processImmediate [as _immediateCallback] (timers.js:396:17)
make: *** [test] Error 7
```

See https://travis-ci.org/restify/node-restify/jobs/411481436 for an example.

## Repro case

The `metrics.test.js` tests suite fails sporadically in the Travis environment. It is not clear to me yet whether this is reproducible in another environment. The nature of the environment in which the tests suite runs is relevant here, as the presence or absence of a virtualization layer may have an impact on the timing facilities of the Node.js runtime, and may have an impact on the behavior of this tests suite.

## Cause

From a quick look at it it seems it could be due to either (this list is not exhaustive):

1) an issue with node timers that _sometimes_ fire 1ms earlier than specified.
2) an issue with the `process.hrtime` implementation that sometimes round up to the nearest milisecond.

Yes. Some thoughts on how to fix this:

1. Investigate timers/`process.hrtime`'s behavior and fix any erroneous behavior that might be identified.
2. Use something like `assert.closeTo` instead of `assert.isAtLeast` to relax the test so that it doesn't fail when the actual value is very close but less than the expected value (suggested by @DonutEspresso ).

michaelnisi:
Seeing values between 49 and 56, relaxing the test to resolve #1691 seems reasonable.


test/plugins/metrics.test.js
```diff
@@ -203,11 +203,11 @@ describe('request metrics plugin', function() {
                     // However the timeout value is 200,
                     // it's calculated by the client,
                     // but setTimeout is happening on the server, tolerate 10ms
-                    assert.isAtLeast(metrics.preLatency, 50);
+                    assert.closeTo(metrics.preLatency, 50, 10);
```
misterdjules: Looking a bit further into this issue, it seems it could be due to https://github.com/nodejs/node/issues/10154. If so, it means that the value of `metrics.preLatency` should always be at least 49ms in this test.

Thus I would suggest changing this to:

```
assert.isAtLeast(metrics.preLatency, 49);
```

And adding a comment why we use the value 49 instead of 50 (and adding a link to that GH issue in that comment).

Moreover, it seems that this specific Node.js core issue has been fixed in Node.js 11.x (but it hasn't been back ported to previous versions), so it would be interesting to check whether we're indeed not seeing this behavior for those tests when running with node 11.x.

michaelnisi:
Well done, looks like your first intuition was correct, locally with Node.js v11.1.0 the tests pass unmodified. I’ve adjusted the test and added a comment.

michaelnisi:
I don’t know why Travis is complaining about unrelated style issues. @misterdjules Would you mind checking this?

PR #1721, opened 2018-11-19
feat: add support for non-strict formatters

misterdjules:
## Pre-Submission Checklist

## Issues

I did not create an issue prior to submitting this PR, as I thought looking at the rationale for this change without looking at the actual code change could be confusing and less productive. Please let me know if you'd still like me to open an issue.

# Changes

> What does this PR do?

This PR makes `res.send` **not** override any content-type that was previously set on the response _even when no formatter can be found for that content-type_.

The goal of this PR is to improve developers' experience. Silently overriding the content-type of the response when no formatter is present seems implicit and surprising. I think we should aim for behavior that is less surprising.

One consequence of this PR is that it decouples the content-type from formatters as long as the content-type is set explicitly (e.g. using `res.setHeader('content-type', ...)` or `res.contentType = ...`). In those cases, it effectively makes formatters optional.

If no content-type is set explicitly on the response, the request would still result in an error if  content-type negotiation failed to match the set of accepted content types from the client's request's `accept` header and the set of formatters available to the server.

Another approach I considered was to make `res.send` return or throw an error in this case and having users who run into that issue use `res.sendRaw`. However it seems that while that behavior would be even more explicit, the current approach seems more flexible and natural for at least some of our partners.

For instance, the approach implemented in this PR is what some developers at Netflix seem to expect from a HTTP framework. Using `res.sendRaw` feels very unnatural to them and would require that they write linting rules to catch when their developers use `res.send` instead of `res.sendRaw`. This change would allow them to get rid of the no-op formatter they wrote to work around this issue while still using `res.send`.

@restify/core I'd like to read your thoughts on this.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -437,10 +437,18 @@ server.get('/websocket/attach', function upgradeRoute(req, res, next) {
 
 ## Content Negotiation
 
-If you're using `res.send()` restify will automatically select the content-type
-to respond with, by finding the first registered `formatter` defined.  Note in
-the examples above we've not defined any formatters, so we've been leveraging
-the fact that restify ships with `application/json`, `text/plain` and
+If you're using `res.send()` restify will determine the content-type to respond
+with by:
+
+1. negotiating the content-type by matching available formatters with the
+   request's `accept` header
+1. otherwise, using `application/json` if the body is an object that is not a
+   Buffer instance
+1. otherwise, using the value of `res.contentType` if present
+1. otherwise, using the value of the `Content-Type` response header if set
+
+Note in the examples above we've not defined any formatters, so we've been
+leveraging the fact that restify ships with `application/json`, `text/plain` and
 `application/octet-stream` formatters. You can add additional formatters to
 restify by passing in a hash of content-type -> parser at server creation time:
 
@@ -460,9 +468,14 @@ var server = restify.createServer({
 });
 ```
 
-If a content-type can't be negotiated, then restify will default to using the
-`application/octet-stream` formatter. For example, attempting to send a
-content-type that does not have a defined formatter:
+If a content-type can't be determined, then restify will respond with an error.
+
+If a content-type can be negotiated, but that content-type doesn't have a
+corresponding formatter, restify will not format the response with any
+formatter.
+
+For example, attempting to send a content-type that does not have a defined
+formatter:
 
 ```js
 server.get('/foo', function(req, res, next) {
@@ -472,12 +485,13 @@ server.get('/foo', function(req, res, next) {
 });
 ```
 
-Will result in a response with a content-type of `application/octet-stream`:
+Will result in a response with a content-type of `text/css` even though no
+`'text/css'` formatter is present:
 
 ```sh
 $ curl -i localhost:3000/
 HTTP/1.1 200 OK
-Content-Type: application/octet-stream
+Content-Type: text/css
 Content-Length: 2
 Date: Thu, 02 Jun 2016 06:50:54 GMT
 Connection: keep-alive
--- a/docs/index.md
+++ b/docs/index.md
@@ -435,17 +435,31 @@ server.get('/websocket/attach', function upgradeRoute(req, res, next) {
 });
 ```
 
-## Content Negotiation
+## Responses' Content Negotiation And Formatting
 
 If you're using `res.send()` restify will determine the content-type to respond
-with by:
+with by, from highest priority to lowest priority:
 
-1. negotiating the content-type by matching available formatters with the
-   request's `accept` header
+1. using the value of `res.contentType` if present
+1. otherwise, using the value of the `Content-Type` response header if set
 1. otherwise, using `application/json` if the body is an object that is not a
    Buffer instance
-1. otherwise, using the value of `res.contentType` if present
-1. otherwise, using the value of the `Content-Type` response header if set
+1. otherwise, negotiating the content-type by matching available formatters with
+   the request's `accept` header
+
+If a content-type can't be determined, then restify will respond with an error.
+
+If a content-type can be negotiated, restify then determines what formatter to
+use to format the response's content.
+
+If no formatter matching the content-type can be found, restify will by default
+override the response's content-type to `'application/octet-stream'` and then
+error if no formatter is found for that content-type.
+
+This default behavior can be changed by passing `optionalFormatters: true`
+(default is false) when creating the restify server instance. In that case, if
+no formatter is found for the negotiated content-type, the response is flushed
+without applying any formatter.
 
 Note in the examples above we've not defined any formatters, so we've been
 leveraging the fact that restify ships with `application/json`, `text/plain` and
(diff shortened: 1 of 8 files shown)


cprussin:
For those who are not able to access the internal Netflix repo linked above, the no-op formatter looks like this:

```js
const formatter = (req, res, body) => body;

...

restify.createServer({
    formatters: {
        'text/html': formatter,
        'text/css': formatter,
        ...
    }
});
```

It's not a great ergonomic, because if a developer forgets to add a line for each content-type they want to send, and don't remember to use `sendRaw`, then the content-type will be overwritten by restify.  This can be hard for folks to debug since they typically don't suspect the server to be changing the content type they set, and remembering to use `sendRaw` is unnatural, especially with developers who are used to other frameworks like express.

micahr:
I'm a 👍 for this change. Definitely don't want to magically pick a content-type without developer input. An additional option would be to allow a developer to specify a fallback or default content-type at initialization time that would then be used if there wasn't anything else specified on the request.

lib/response.js
```diff
@@ -464,33 +466,20 @@ function patch(Response) {
...
             type = type + '; charset=' + self._charSet;
         }
 
-        // Update header to the derived content type for our formatter
+        // Update Content-Type header to the one originally set or to the type
+        // inferred from the most relevant formatter found.
         self.setHeader('Content-Type', type);
 
+        if (!formatter) {
+            return flush(self, body);
```
DonutEspresso: I was wondering if we should assert on this unformatted body to ensure it is either string/buffer, but I realized we call `flush()` in many places so it may be outside of the scope of this PR. I would expect res.write to throw in that scenario, but it may be better if we can catch it earlier and provide a better error message.

DonutEspresso:
Apologies, one more q - does this effectively remove the need for `sendRaw`?

docs/index.md
```diff
@@ -437,10 +437,18 @@ server.get('/websocket/attach', function upgradeRoute(req, res, next) {
...
 ## Content Negotiation
 
-If you're using `res.send()` restify will automatically select the content-type
-to respond with, by finding the first registered `formatter` defined.  Note in
-the examples above we've not defined any formatters, so we've been leveraging
-the fact that restify ships with `application/json`, `text/plain` and
+If you're using `res.send()` restify will determine the content-type to respond
+with by:
+
+1. negotiating the content-type by matching available formatters with the
```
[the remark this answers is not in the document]
misterdjules: The way I ordered it is by order of priority, which is the reverse of the order in which the code runs those checks. Does that make sense?

If so would you still prefer to order it the same way the code performs those checks or should I keep it the way it is and mention that it's ordered by priority?

lib/response.js
misterdjules: That's a good point. We're already [asserting on that in the `sendRaw` case](https://github.com/restify/node-restify/blob/9a59b0f8ce6b4c75b3904aff00d11e60b40a6846/lib/response.js#L422-L428), so we might as well do the same as you suggest in this case.

misterdjules:
@DonutEspresso 

> Apologies, one more q - does this effectively remove the need for sendRaw?

I don't think it does. With this change `res.sendRaw`'s use case would still be about:

1. not applying any formatter at all to the response, whether it's optional or not
1. not performing any content-type negotiation (not inferring the response's content-type based on the request's accept header if content-type is not set)

misterdjules:
@DonutEspresso I just pushed a second commit that changes the logic in `res.__send` a little bit so that we assert on the type of the unformatted body both when:

1. `res.__send` is called through `res.sendRaw`
2. `res.__send` is called through `res.send` but not formatter is found

Let me know if that looks better to you!

misterdjules:
@cprussin Would it be worth it for you to test in a branch of your application that this PR addresses your use cases before we merge it?

docs/index.md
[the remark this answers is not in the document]
DonutEspresso: Ah, I see now, I was expecting it the other way around. Maybe a note to say "ascending priority" for clarity, then?

lib/response.js
```diff
@@ -417,71 +417,72 @@ function patch(Response) {
...
+                // results
+                return flush(self, formatter(self.req, self, body));
+            }
         }
 
-        // Finally, invoke the formatter and flush the request with it's results
-        return flush(self, formatter(self.req, self, body));
+        // if no formatting, assert that the value to be written is a string or
+        // a buffer, then send it.
+        assert.ok(
```
DonutEspresso: I just realized that the formatters don't necessarily enforce a proper return value, so could we benefit from moving this assertion into `flush()`?

misterdjules: 👍

misterdjules:
Another approach that I haven't mentioned so far would be to "rename" `res.sendRaw` to `res.send` and rename `res.send` to `res.sendFormatted`, or `res.send(code, body, headers, {formatted: true})`, and have the "formatted version" of `res.send` be more strict when formatters are missing.

@restify/core Any preference between the approach currently implemented in this PR (making formatters optional and their usage implicit) and the one I just mentioned (making using formatters more explicit)?

I don't have a strong preference either way, but I thought I'd make sure we present various approaches before committing, as those changes are all breaking.

misterdjules:
Update on current status: add a discussion with @DonutEspresso and @mridgway and we decided to make this change non-breaking. This will:

* Avoid the need for cutting a new major release of restify (by definition)
* Allow consumers to opt into this behavior rather than force them to transition.

If/when we're confident this new behavior is suitable for all consumers, we can propose to enable it by default as a breaking change.

misterdjules:
@restify/core This is ready for another (hopefully final) round of review. Thanks for your patience and my apologies for the unnecessary churn on that one 🙏 

@hekike Since this is no longer a breaking change, we don't need a migration guide anymore.

lib/response.js
```diff
@@ -346,6 +348,11 @@ function patch(Response) {
             };
         }
 
+        assert.ok(
```
DonutEspresso: 👌

test/formatter-optional.test.js
```diff
@@ -0,0 +1,77 @@
...
+        process.exit(1);
+    }
+});
+
+test('should send 200 on formatter missing but optional', function(t) {
+    // When server is passed "optionalFormatters: true" at creation time,
+    // res.send still sends a successful response even when a formatter is not
+    // set up for a specific content-type.
+    SERVER.get('/11', function handle(req, res, next) {
+        console.log('got request');
```
DonutEspresso: Should this be removed? One in the subsequent test too.

misterdjules:
@DonutEspresso 

> Would it be less confusing if the option name was strictFormatters, which defaults to true and enforces the current RFC behavior?

`strictFormatters` definitely works too, I went back and forth between those two when putting those changes together.

I went for `optionalFormatters` because by default it is `false`, and one has to opt into it by setting the value to `true`, which I found less confusing at the time. I like the name `strictFormatters` better though, so I'm totally OK with making that change.

test/formatter-optional.test.js
misterdjules: Ah yes good catch :)

misterdjules:
@DonutEspresso I believe I addressed your latest comments, do you mind taking another look? Thanks!

DonutEspresso:
:tada:

PR #1722, opened 2018-11-21
Update response.send() to check HTTP status code length

Adimvicky:
I updated line 300 which checks that the HTTP status code passed into the Response.send() function is of type "number" to also check that the length of the status code is equal to three since all HTTP status codes are three digits. This allows for the Response.send() function to send ordinary numbers of any length ( other than three ) as response if need be, instead of throwing.

## Pre-Submission Checklist

## Issues

Closes:

* Issue #
* Issue #
* Issue #

# Changes

> What does this PR do?

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -297,7 +297,7 @@ function patch(Response) {
         var self = this;
         var sendArgs;
 
-        if (typeof code === 'number') {
+        if (typeof code === 'number' && String(code).length === 3) {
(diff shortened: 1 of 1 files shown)


kolbma:
I don't think this would be a good feature like it is...

1. It adds computation overhead to each and every request
2. You can simply use something like https://www.npmjs.com/package/statuses and check it in your own code

Issue #713, opened 2014-11-27
How to disable built-in formatters?


bsander:
We have created our own formatters for various response types (json, xml, html). I want to exclusively use these custom formatters and not allow the use of the built-in restify formatters which are still allowing requests with an accept header of `appliction/octet-stream` or `text/plain` for example to get through. How can I remove these built-in formatters from my restify instance?

micahr:
if you pass a set of formatters in to your server:

```
restify.createServer({
    formatters: { 
        'application/javascript; q=0.1': require('./your-jsonp'),
        'application/json; q=0.4': require('./your-json'),
        'text/plain; q=0.3': require('./your-text'),
        'application/octet-stream; q=0.2': require('./your-binary')
    }
});
````

You should be able to set the q values lower than the built ins, which will pick yours over the built ins.

**Comment by bsander** [comment]
@micahr but what about actually *disabling* these response types? I don't have my own to replace them with, I just want restify to return a 406 like it does with any other unsupported type.

**Comment by micahr** [comment]
I believe if you pass in a formatters property when creating the server that doesn't have a specific mime-type, that will accomplish what you are looking for. For Example, this would only allow `application/javascript` as a mime-type.

```javascript
restify.createServer({
    formatters: { 
        'application/javascript; q=0.1': require('./your-jsonp')
    }
});
```

bsander:
Unfortunately, that is not the case (at least in 2.8.5).

PR #1728, opened 2018-12-21
feat: add support for disabling built-in formatters

misterdjules:
## Pre-Submission Checklist

## Issues

Possibly related: #713 (even if already closed)

# Changes

> What does this PR do?

This PR allows users of Restify to disable built-in formatters. There are several use cases for this, one being the ability to completely replace built-in formatters, which is described in https://github.com/restify/node-restify/issues/713.

Another one is for Restify users who set `strictFormatters: false` when creating the server and send already formatted content using `res.send`. In this case, they don't want any formatter (e.g. built-in formatters) to format the content sent in any way.

An alternative to this use case might be to use `res.sendRaw`, however I believe there are still subtle behaviors that would be lost when moving to `.sendRaw` and that would be needed for those users (like setting the proper charset for the response's content-type, etc.).


misterdjules:
Thanks @cprussin and @rajatkumar for the review!

We'll let @cprussin test this further by making his application depend on the `add-use-builtin-formatters-flag` branch of this repo, and once he reports whether he found any issue with this, we'll merge into master.

misterdjules:
@cprussin made his application depend on a version of restify with the changes in this PR and identified cases where responses would not have a `content-type` header set manually, and because built-in formatters are disabled, [content-type negotiation that happens in `response.send`](https://github.com/restify/node-restify/blob/de1833a44084e5f231de289421518ec646b86f60/lib/response.js#L448) would fail to identify a `content-type` for the response, which would [result in setting the response's status to 406](https://github.com/restify/node-restify/blob/de1833a44084e5f231de289421518ec646b86f60/lib/response.js#L450-L461).

This is not exactly the desired behavior because:

1. `content-type` not being a required header (see section 7.2.1 of [RFC 2616](https://www.ietf.org/rfc/rfc2616.txt)), failing the response when it is not set seems a bit outside of restify's scope. Instead, the application should be able to determine whether a response without a content-type set is acceptable

1. if the application determines that responses without a `content-type` header set should fail, it seems that those failures should be handled as programmer errors by throwing an exception, as they represent bugs in the application

Thinking more about this, it also feels like disabling the built-in formatters was a way for us to opt-out of the response's formatting subsystem without impacting the content negotiation mechanism, when those two are actually dependent on each other.

As a result, @mridgway, @cprussin and myself agreed to **abandon this PR** and explore another path forward that consists of:

1. introducing a new `resSendFormatsResponse` server flag that is true by default and can be set to false so that `res.send` is effectively equivalent to calling `res.sendRaw`

1. introducing a new `requiredResponseHeaders` server option that is an array (empty by default) of required header names. If one of those headers is not set on any response sent by that server, an exception is thrown.

@restify/core Thoughts on the proposal above?

misterdjules:
Closing (at least for now) as per [the above comment](https://github.com/restify/node-restify/pull/1728#issuecomment-460477205) and in favor of #1747 and #1748.

Issue #1726, opened 2018-12-07
static.test.js tests suite is flaky


misterdjules:
- [x] Used appropriate template for the issue type
- [x] Searched both open and closed issues for duplicates of this issue
- [x] Title adequately and _concisely_ reflects the feature or the bug

# Bug Report

## Restify Version

7.3.0

## Node.js Version

8.12.0

## Expected behaviour

`static.test.js` should pass all the time.

## Actual behaviour

`static.test.js` fails sometimes when running that test locally or when it's run by TravisCI.

## Repro case

```
$ (latest_exit_code=0; while [ $latest_exit_code -eq 0 ]; do ./node_modules/.bin/mocha test/plugins/static.test.js; latest_exit_code=$?; done)
```

The sub-shell will exit after a few dozens or hundreds runs and error with:

```
 static resource plugin
    ✓ static serves static files
    ✓ static serves static files in nested folders
    ✓ static serves static files in with a root regex
    ✓ static serves static files with a root, !greedy, regex
    ✓ static serves default file
    ✓ restify-GH-379 static serves file with parentheses in path
    ✓ restify-GH-719 serve a specific static file
    ✓ static serves static file with appendRequestPath = false
    ✓ static serves default file with appendRequestPath = false
    ✓ restify serve a specific static file with appendRequestPath = false
    ✓ static responds 404 for missing file
    ✓ GH-1382 static responds 404 for missing file with percent-codes
    ✓ static does not leak the file stream and next() is properly called when the client disconnects before receiving a reply
    1) static does not open a file stream and next() is properly called when the client disconnects immediately after sending a request
    ✓ static responds 404 for missing file
    ✓ GH-1382 static responds 404 for missing file with percent-codes


  15 passing (132ms)
  1 failing

  1) static resource plugin static does not open a file stream and next() is properly called when the client disconnects immediately after sending a request:
     Uncaught AssertionError: Unspecified AssertionError
      at Object.fs.createReadStream (test/plugins/static.test.js:344:17)
      at serveFileFromStats (lib/plugins/static.js:113:26)
      at fileStat (lib/plugins/static.js:159:17)
      at FSReqWrap.oncomplete (fs.js:153:5)
```

## Cause

I don't know the cause.

## Are you willing and able to fix this?

Yes.

misterdjules:
[...]

PR #1729, opened 2018-12-21
fix: fix flaky static.test.js

misterdjules:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #1726

The `static.test.js` tests suite is flakey because it incorrectly relies on a specific order of events that is not the only one possible.

# Changes

> What does this PR do?

This PR makes the test actually wait for the request to be marked as closed before running the `static` middleware. I believe this was the original intention when writing this test.

Without that additional check, it is possible for the server to receive and handle the request _before_ it actually handles the event that closes the connection.

@koute I'd like to get your thoughts on this since you're the original author of that test.

### DIFF
--- a/test/plugins/static.test.js
+++ b/test/plugins/static.test.js
@@ -365,14 +365,16 @@ describe('static resource plugin', function() {
                     }
 
                     // closed before serve
-                    if (req.closed) {
+                    if (req.closed()) {
                         doServe();
                     } else {
                         req.on('close', doServe);
                     }
                 });
                 SERVER.on('after', function(req, res, route, afterErr) {
-                    assert(afterErr.name, 'RequestCloseError');
+                    if (afterErr) {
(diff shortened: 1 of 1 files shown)


koute:
> @koute I'd like to get your thoughts on this since you're the original author of that test.

I'd love to help you out, however I wrote this so long ago that I don't really remember much anymore, unfortunately. (:

I don't remember this test being flakey when I wrote it, but it may or may not have depended on some internal implementation details regarding the event ordering, I'm not sure at this point.

misterdjules:
@koute No problem, thank you for sharing that!

misterdjules:
As [mentioned in the relevant issue](https://github.com/restify/node-restify/issues/1726#issuecomment-450699340), closing this because for some reason I cannot reproduce that issue anymore.

PR #1730, opened 2018-12-22
fix: more flaky metrics.test.js fixes

misterdjules:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #1691

# Changes

> What does this PR do?

Fix more occurrences of the problem described at #1691. Previous fixes did not cover the whole tests suite. I expect all it to be fully covered now.

### DIFF
--- a/test/plugins/metrics.test.js
+++ b/test/plugins/metrics.test.js
@@ -15,6 +15,13 @@ var SERVER;
 var CLIENT;
 var PORT;
 
+function adjustExpectedLatency(expectedLatency, nbTimers) {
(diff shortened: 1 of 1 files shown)


misterdjules:
Tests failure is not a regression due to this PR and is fixed by https://github.com/restify/node-restify/pull/1729, thus merging this asap. Thanks for the review @DonutEspresso!

misterdjules:
Ah, that commit message should probably have used the `test:` prefix, not `fix:`. It shouldn't have any impact though, since I assume the next release will be at least a patch release and thus would bump the patch version number anyway.

Issue #1731, opened 2019-01-01
static.test.js fails with node >= v11.4.x


misterdjules:
- [x] Used appropriate template for the issue type
- [x] Searched both open and closed issues for duplicates of this issue
- [x] Title adequately and _concisely_ reflects the feature or the bug

# Bug Report

## Restify Version

7.3.0

## Node.js Version

\>= v11.4.0

## Expected behaviour

`static.test.js` should pass.

## Actual behaviour

```
$ nvm use 11.4.0
Now using node v11.4.0 (npm v6.4.1)
$ ./node_modules/.bin/mocha test/plugins/static.test.js 


  static resource plugin
    ✓ static serves static files
    ✓ static serves static files in nested folders
    ✓ static serves static files in with a root regex
    ✓ static serves static files with a root, !greedy, regex
    ✓ static serves default file
    ✓ restify-GH-379 static serves file with parentheses in path
    ✓ restify-GH-719 serve a specific static file
    ✓ static serves static file with appendRequestPath = false
    ✓ static serves default file with appendRequestPath = false
    ✓ restify serve a specific static file with appendRequestPath = false
    ✓ static responds 404 for missing file
    ✓ GH-1382 static responds 404 for missing file with percent-codes
    ✓ static does not leak the file stream and next() is properly called when the client disconnects before receiving a reply
    1) static does not open a file stream and next() is properly called when the client disconnects immediately after sending a request
    ✓ static responds 404 for missing file
    ✓ GH-1382 static responds 404 for missing file with percent-codes


  15 passing (134ms)
  1 failing

  1) static resource plugin static does not open a file stream and next() is properly called when the client disconnects immediately after sending a request:
     Uncaught TypeError: Cannot read property 'name' of undefined
      at Server.<anonymous> (test/plugins/static.test.js:367:37)
      at Server.EventEmitter.emit (domain.js:441:20)
      at Server._finishReqResCycle (lib/server.js:1259:14)
      at ServerResponse.onResClose (lib/server.js:1207:14)
      at ServerResponse.EventEmitter.emit (domain.js:441:20)
      at Socket.onServerResponseClose (_http_server.js:169:44)
      at Socket.EventEmitter.emit (domain.js:441:20)
      at TCP._handle.close (net.js:613:12)



$ 
```

## Repro case

See above.

## Cause

[...]

PR #1732, opened 2019-01-01
fix: emit after event with proper error param for node versions >= 11.4.0

misterdjules:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #1731

Restify currently relies on specific timing of the response's 'close' event to emit the 'after' event with the correct error parameter in case the request is aborted by the client.

# Changes

> What does this PR do?

This PR makes Restify _not_ overwrite any error previously set as `res.err` so that it is passed as the first parameter when emitting the `'after'` event once the req/res lifecycle is finished.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -1258,8 +1258,14 @@ Server.prototype._finishReqResCycle = function _finishReqResCycle(
         // after event has signature of function(req, res, route, err) {...}
         self.emit('after', req, res, route, err || res.err);
     } else {
-        // preserve error for actual finish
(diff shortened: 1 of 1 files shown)


misterdjules:
Note that this is _potentially_ a breaking change, as there are potentially one or more situations where an `'after'` event would have been emitted without an error that would now result in an `'after'` event being emitted with an error.

I'm not sure this is worth the overhead though, so for now I'm inclined to treat this as a bug fix and keep this as a patch version bump.

PR #1733, opened 2019-01-02
fix(plugins): do not include user-input in UnsupportedMediaTypeError

hekike:
User input in `VError` based errors (`restify-errors`) can cause the process to crash as `VError` uses the `sprintf` library internally which expects additional arguments for formatting symbols liek `%s`.

For example, a `content-type` `%s` would crash restify without this change.

### DIFF
--- a/lib/plugins/bodyReader.js
+++ b/lib/plugins/bodyReader.js
@@ -95,7 +95,7 @@ function bodyReader(options) {
         var hash;
         var md5;
 
-        var unsupportedCompression;
+        var unsupportedContentEncoding;
 
         if ((md5 = req.headers['content-md5'])) {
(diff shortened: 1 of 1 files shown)


hekike:
LTS fails but @misterdjules is working on a fix, unrelated to this pr.

misterdjules:
Nitpicking and thinking out loud: should that be a minor version bump since we're adding an info field to the error object (which one could argue is technically an addition to the plug-in's API)?

hekike:
Minor sounds good to me, I think the reasoning is valid.
I changed the commit message two reflect your points, thanks!

hekike:
released as `restify@7.4.0`

PR #1740, opened 2019-01-18
feat(req): add restifyDone event

hekike:
Emit a `restifyDone` event on the `req` object when the request is "fully served" (see docs in this PR).

### DIFF
--- a/docs/api/request-events.md
+++ b/docs/api/request-events.md
@@ -7,7 +7,7 @@ restify considers a request to be fully serviced when either:
 2) An error was returned to `next()`, and the corresponding error events have
    been fired for that error type
 
-The signature is for the `restifyDone` event is as follows:
+The signature for the `restifyDone` event is as follows:
 
 ```js
 function(route, error) { }
--- a/docs/api/request-events.md
+++ b/docs/api/request-events.md
@@ -1,0 +1,21 @@
+### restifyDone
+
+After request has been fully serviced, an `restifyDone` event is fired.
+restify considers a request to be fully serviced when either:
+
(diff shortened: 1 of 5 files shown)


docs/api/request-events.md
```diff
@@ -0,0 +1,21 @@
+### restifyDone
+
+After request has been fully serviced, an `restifyDone` event is fired.
+restify considers a request to be fully serviced when either:
+
+1) The handler chain for a route has been fully completed
+2) An error was returned to `next()`, and the corresponding error events have
+   been fired for that error type
+
+The signature is for the `restifyDone` event is as follows:
```
DonutEspresso: s/is/''

docs/api/request-events.md
```diff
@@ -0,0 +1,21 @@
...
+
+```js
+function(route, error) { }
```
DonutEspresso: Any reason not to match the API of server after?

test/request.test.js
```diff
@@ -223,3 +223,24 @@ test('should provide date when request started', function(t) {
         t.end();
     });
 });
+
+// restifyDone is emitted at the same time when server's after event is emitted,
+// you can find more comprehensive testing for `after` lives in server tests.
+test('should emit restifyDone event when request is fully served', function(t) {
```
DonutEspresso: Should we add a test for `next(err)` too?

docs/api/request-events.md
hekike: I just thought it would look strange that we pass req in a req event...

PR #1742, opened 2019-01-24
Fix dev

retrohacker:
I've broken this into two commits to make things easier to manage in our git history in case we need to revert something.

#### `fix(dev): remove nsp since the project merged with npm`

The equivalent of `nsp` now runs on every `npm install`.

#### `fix(dev): pin to exact versions of linting tools and fix lint errors`

I've been seeing non-deterministic behavior for our linting rules, and suspect it has a lot to do with different versions of eslint and prettier getting pulled in under different circumstances. This commit pins to the exact version of these tools to keep our linting rules stable. It also fixes the linting errors that these versions complain about.


Makefile
```diff
@@ -63,10 +62,6 @@ test: $(NODEUNIT)
 	$(NODEUNIT) test/*.test.js
 	$(MOCHA) test/plugins/*.test.js
 
-.PHONY: nsp
```
rajatkumar: I think we should do `npm audit` instead of removing this check altogether.

retrohacker: `npm audit` runs on every `npm install`.

DonutEspresso:
Conventional commit prefix?

retrohacker:
The commits have the prefix, I'm planning on a rebase to master :-)

PR #1743, opened 2019-01-24
Dynamic maxBodySize for bodyReader plugin

retrohacker:
Instead of caching the value for `opts.maxBodySize` in memory, we read it off the configuration object whenever we need it. This allows the caller to update the value at runtime.

### DIFF
--- a/lib/plugins/bodyReader.js
+++ b/lib/plugins/bodyReader.js
@@ -66,8 +66,6 @@ function bodyReader(options) {
     var opts = options || {};
     assert.object(opts, 'opts');
 
-    var maxBodySize = opts.maxBodySize || 0;
-
     function readBody(req, res, originalNext) {
         var next = once(originalNext);
 
@@ -118,8 +116,8 @@ function bodyReader(options) {
                 return;
             }
 
-            if (maxBodySize && bytesReceived > maxBodySize) {
-                var msg = 'Request body size exceeds ' + maxBodySize;
+            if (opts.maxBodySize && bytesReceived > opts.maxBodySize) {
+                var msg = 'Request body size exceeds ' + opts.maxBodySize;
                 var err;
 
                 // Between Node 0.12 and 4 http status code messages changed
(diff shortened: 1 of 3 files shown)


retrohacker:
This includes commits from #1742 for making linting consistent between CI/CD and local development.

That PR should be reviewed and merged before this one.

test/plugins/jsonBodyParser.test.js
```diff
@@ -413,6 +413,44 @@ describe('JSON body parser', function() {
...
+            path: '/',
+            method: 'POST',
+            agent: false,
+            headers: {
+                accept: 'application/json',
+                'content-type': 'application/x-www-form-urlencoded',
+                'transfer-encoding': 'chunked'
+            }
+        };
+        var client = http.request(opts, function(res) {
```
DonutEspresso: Any reason not to use the local global restify-client `CLIENT`?

retrohacker:
> One thought - the "dynamic by reference" is implicit and not immediately obvious.

Yeah, this bothers me a bit too. Especially if we adopt this as a pattern it just takes one place in the codebase to dereference the value and we have a nasty bug. For now, I personally feel comfortable relying on this pattern given:

1) Documentation (maybe we should add additional comments where we read the values?)
2) Test that ensures the value remains dynamic moving forward.

> You also can't opt out of it.

```js
// Don't accept payloads larger than 5MB
var bodyParserOpts = { maxBodySize: 5000000 }
// Don't allow this value to be changed at runtime
Object.freeze(bodyParserOpts)
// Add bodyParser to the middleware chain
server.use(bodyParser(bodyParserOpts))
```

test/plugins/jsonBodyParser.test.js
retrohacker: Heh, I just copied another test :stuck_out_tongue: I'll give the provided client a go.

lib/plugins/bodyReader.js
```diff
@@ -66,8 +66,6 @@ function bodyReader(options) {
     var opts = options || {};
     assert.object(opts, 'opts');
 
-    var maxBodySize = opts.maxBodySize || 0;
```
hekike: I guess we could still apply the default value.

lib/plugins/bodyReader.js
```diff
@@ -118,8 +116,8 @@ function bodyReader(options) {
                 return;
             }
 
-            if (maxBodySize && bytesReceived > maxBodySize) {
```
hekike: How about checking that it's a Number here?

retrohacker:
Closing this in favor of using a dynamic `bodyReader` at runtime.

PR #1746, opened 2019-01-30
feat(audit): Add the ability to specify a custom audit log serializer…

daviande:
… (for err, req and res)

## Pre-Submission Checklist

- [x ] Ran the linter and tests via `make prepush`
- [x ] Included comprehensive and convincing tests for changes

## Issues

Closes:

* Issue #
* Issue #
* Issue #

# Changes

> What does this PR do?

Adds the ability to specify a custom audit log serializer for err, req and res


lib/plugins/audit.js
```diff
@@ -198,67 +201,80 @@ function auditLogger(opts) {
...
-                    timers: timers,
-                    connectionState:
-                        req.connectionState && req.connectionState()
-                };
-            },
-            res: function auditResponseSerializer(res) {
-                if (!res) {
-                    return false;
-                }
+    var reqSerializer = function auditRequestSerializer(req) {
```
DonutEspresso: Now that you've broken out the default serializers, it looks like these could probably be lifted to the top as constants/defaults (or maybe even a `DEFAULT_SERIALIZERS` obj). Is that possible, or do we need the closure still?

lib/plugins/audit.js
```diff
@@ -198,67 +201,80 @@ function auditLogger(opts) {
...
+
+        return {
+            statusCode: res.statusCode,
+            headers: getResponseHeaders(res),
+            trailer: res._trailer || false,
+            body: body
+        };
+    };
+
+    if (opts.serializers && opts.serializers.res) {
```
DonutEspresso: Style nit: might we be able to collapse these three if checks into a single assignment? Something like:

```js
var serializers = Object.assign({}, DEFAULT_SERIALIZERS, opts.serializers);
```

lib/plugins/audit.js
daviande: I think we still need the closure because of the " body: opts.body === true ? req.body : undefined,".  We could do something like create a function that returns the DEFAULT_SERIALIZERS and pass it the opts/opts.body param.  It might just be enough to define a DEFAULTS_SERIALIZERS within the clousure, though.  What do you think is more readable?

lib/plugins/audit.js
daviande: This is a good idea.

daviande: Made the change, what do you think now @DonutEspresso?

PR #1747, opened 2019-02-05
feat: add support for disableResponseFormatting flag

misterdjules:
## Pre-Submission Checklist

# Changes

> What does this PR do?

This PR implements the first part of the proposal at https://github.com/restify/node-restify/pull/1728#issuecomment-460477205. It implements a new `Server` flag named `disableResponseFormatting` that basically makes `res.send` behave like `res.sendRaw`.

The goal is to provide users of restify with a familiar response API (`res.send`) while bypassing the formatting/content-type negotiation mechanism.

The `strictFormatters` flag added recently was a first attempt at solving the same use case, but it assumed that all responses would have a `content-type` header which is not always the case. 

Moreover, when a `content-type` header is not present, instead of sending a 406 response and handle it as an operational error, we want to handle it as a programmer error, but only if the programmer opts into this behavior separately. As a result, a follow-up PR will add support for another `requiredResponseHeaders` option that can be passed to the Server constructor.


misterdjules:
Closed as part of cleaning up my list of inflight PRs, and because I don't think this work is relevant anymore.

PR #1750, opened 2019-02-19
breaking: restify drops Node v4.x and v6.x

rajatkumar:
## Pre-Submission Checklist

## Issues

Closes: None

* Issue #
* Issue #
* Issue #

# Changes

> What does this PR do?
This PR is to upgrade `find-my-way` to  `^2.0.0` which drops support for Node 4.x. This means going forth restify will also drop support for Node 4.x. The changes included in this PR also upgrades `lodash@4.17.11` (4.17.10 had a vulnerability) and marks target engine as Node 6.x or above. Some of the test files were changed by `make fix-style`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-   - '6'
    - '8'
    - '10'
    - "lts/*"                # Active LTS release
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 sudo: false
 language: node_js
 node_js:
-   - "4"                    # Maintenance LTS release
+   - '6'
+   - '8'
+   - '10'
    - "lts/*"                # Active LTS release
    - "node"                 # Latest stable release
 env:
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "report-latency": "./bin/report-latency"
   },
   "engines": {
-    "node": ">=4.0"
+    "node": ">=6.0"
   },
   "dependencies": {
     "assert-plus": "^1.0.0",
@@ -98,10 +98,10 @@
     "csv": "^1.1.1",
     "escape-regexp-component": "^1.0.2",
     "ewma": "^2.0.1",
-    "find-my-way": "^1.13.0",
+    "find-my-way": "^2.0.0",
     "formidable": "^1.2.1",
     "http-signature": "^1.2.0",
-    "lodash": "^4.17.10",
+    "lodash": "^4.17.11",
     "lru-cache": "^4.1.3",
     "mime": "^1.5.0",
     "negotiator": "^0.6.1",
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "report-latency": "./bin/report-latency"
   },
   "engines": {
-    "node": ">=6.0"
+    "node": ">=8.0"
   },
   "dependencies": {
     "assert-plus": "^1.0.0",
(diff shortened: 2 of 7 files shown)


.travis.yml
```diff
@@ -1,7 +1,9 @@
 sudo: false
 language: node_js
 node_js:
-   - "4"                    # Maintenance LTS release
+   - '6'
+   - '8'
+   - '10'
```
misterdjules: Isn't 10 covered by `lts/*` if `lts/*` means "active LTS release"? Or is it that we want to hard-code 10 because we want to keep tests running when 10 becomes not active?

.travis.yml
```diff
@@ -1,7 +1,9 @@
 sudo: false
 language: node_js
 node_js:
-   - "4"                    # Maintenance LTS release
+   - '6'
```
misterdjules: I think we discussed dropping support for node 6.x .too as it drops from maintenance in April 2019. Thoughts on that?

.travis.yml
rajatkumar: The idea is to ensure Node10 always gets tested even when it is not LTS.

.travis.yml
rajatkumar: Looking at https://nodesource.com/node-by-numbers seems like Node 6 usage did not really dip so much. I believe we will still have systems using Node6 and may be it is worth supporting for some more time. Thoughts?

.travis.yml
DonutEspresso: I would vote to either pin the versions or use lts/* always - I think we historically pinned since dropping support for LTS meant (usually) bumping major version, so someone always had to do something to make it happen (vs CI just automatically dropping tests for that version).

rajatkumar: We are using `lts/*` which means it will cover active LTS but not maintenance LTS. For maintenance LTS you need to specify the specific versions. (Atleast that is what my understanding is)

.travis.yml
misterdjules: On one hand, this major bump can be a good opportunity for Restify 8.x not to be burdened by the need to support node 6.x when it's not maintained starting in April 2019.

On the other hand, those systems running node 6.x would need to be supported and we may need to maintain two Restify release lines (7.x and 8.x) if we go down that route.

I'm fine either way.

I think @retrohacker and @hekike were the ones who suggested to drop 6.x as well. Do you mind sharing your thoughts on dropping support for node 6.x in this major bump?

rajatkumar: On chatting with @hekike I found that restify codebase seems conservative in order to support older Node versions. The additional burden is that we have had PRs using new language specs and we had to turn them down plus this adds restrictions and prevents us from updating a few npm modules. I think we should go ahead and drop Node6 support too.

misterdjules: I don't have any objection as long as we accept that we might need to maintain Restify 7.x for some time in the future.

package.json
```diff
@@ -90,18 +90,18 @@
...
+    "node": ">=8.0"
   },
   "dependencies": {
     "assert-plus": "^1.0.0",
     "bunyan": "^1.8.12",
     "csv": "^1.1.1",
     "escape-regexp-component": "^1.0.2",
     "ewma": "^2.0.1",
-    "find-my-way": "^1.13.0",
+    "find-my-way": "^2.0.0",
```
delvedor: Hello! `find-my-way` maintainer here. `v2.0.0` had a security issue, we have already fixed it in `v2.0.1`. I would recommend to use it as default version since `v2.0.0` is deprecated and tools such as [snyk](https://snyk.io) will flag the project as vulnerable :)

rajatkumar: @delvedor thanks a ton for letting us know! I will bump it up to use `v2.0.1`

Issue #1617, opened 2018-02-27
Support serve-static from express


retrohacker:
- [x] Used appropriate template for the issue type
- [x] Searched both open and closed issues for duplicates of this issue
- [x] Title adequately and _concisely_ reflects the feature or the bug

# Feature Request

## Use Case

It would be super cool if we could just use the [serve-static](https://ghub.io/serve-static) plugin from the express ecosystem instead of continuing to invest in maintaining a parallel (and broken) internal serveStatic plugin.

> Why do you want this?

## Example API

```js
var restify = require('restify');
var serveStatic = require('serve-static');

var app = restify.createServer();
var serve = serveStatic('./static');

app.get('.*', function(req, res, next) {
    res.addListener('finish', next);
    serve(req, res, function (err) {
        res.removeListener('finish', next);
        next(err);
     });
});

app.listen(8080);
```

It also looks like we need to support the `req.originalUrl` pattern similar to express: https://github.com/expressjs/express/blob/c0136d8b48dd3526c58b2ad8666fb4b12b55116c/lib/router/index.js#L172

> This should include code snippets and documentation for the proposed feature

## Are you willing and able to implement this?

> "Yes" or, if "no", what can current contributors do to help you create a PR?

Yup!

kokarn:
This looks very promising!

Would this allow us to have say a static folder with some files served that way and the rest of the routes served by Restify itself?

My use case is that a pretty common scenario is to have a bunch of static files i'd like to serve, robots.txt, analytics auth and so on and the easiest way would be to have a static folder with that, and all the other routes just configured as per usual.

PR #1753, opened 2019-02-23
feat(plugin): plugin to serve static files

rajatkumar:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #1617 
* Issue #
* Issue #

# Changes
Added the `serveStaticFiles` plugin and set of tests. Also updates the docs to reflect this new plugin and its API.

> What does this PR do?
Adds another plugin to serve static files. This plugin uses [send](https://github.com/pillarjs/send) under hood, which is the same library that expressjs relies on. This should resolve most of the other issues developers have seen with restify's static plugin.

I decided to keep the existing plugin along with this one so that we are not breaking any existing usage of the `static` plugin in the ecosystem.


lib/plugins/staticFiles.js
```diff
@@ -0,0 +1,131 @@
...
+        // Lets handle the various events being emitted by send module
+
+        // stream has ended, must call `next()`
+        stream.on('end', function handleEnd() {
+            return next();
+        });
+
+        // when `send` encounters any `error`, we have the opportunity
+        // to handle the errors here
+        stream.on('error', function handleError(err) {
```
hekike: how about:

```
        stream.on('error', function handleError(err) {
            // When file does not exist
            if (err.statusCode === 404) {
                return next(new ResourceNotFoundError(requestedFullPath));
            } 
            // or action is forbidden (like requesting a directory)
            return next(new NotAuthorizedError(requestedFullPath));
        });
```

hekike: will it be always `NotAuthorizedError `?

rajatkumar: I wanted to put down explicit cases and hence the if/else structure. I foresee we might have other cases(which we find later on as people start using this plugin) where we want to send explicit errors.

If it is fine, I would like to keep the current structure.

lib/plugins/staticFiles.js
```diff
@@ -0,0 +1,131 @@
+'use strict';
+var assert = require('assert-plus');
```
DonutEspresso: If this plugin is for new versions only, maybe we can use let/const here?

lib/plugins/staticFiles.js
```diff
@@ -0,0 +1,131 @@
...
+var NotAuthorizedError = errors.NotAuthorizedError;
+var ResourceNotFoundError = errors.ResourceNotFoundError;
+
+/**
+ * Serves static files, with API similar to expressjs
+ *
+ * @public
+ * @function serveStaticFiles
+ * @param    {String} directory - the directory to serve files from
+ * @param    {Object} opts - an options object, which is optional
```
DonutEspresso: Is it worth calling this out as an options object meant explicitly for `send`? I think we do something similar with other plugins like mime.

rajatkumar: The options object is not explicitly for `send`, I added another functionality (similar to express) that allows us to set custom headers by passing a function under `options.setHeaders`. eg.
```
setHeaders: function setCustomHeaders(response, requestedPath, stat) {
              response.setHeader('restify-plugin-x', 'awesome');
          }
 })
```

Also, we explicitly override `options.root` before passing to `send`. So the other options might be similar to what `send` needs but they are not explicitly for `send`.

lib/plugins/staticFiles.js
rajatkumar: I did not want to tackle the eslint rule that disables usage of es6 constructs in code as part of this PR. (Refer: https://github.com/restify/node-restify/blob/master/.eslintrc.js#L12). Let's keep that for a future PR where we might convert vars -> consts/let in the codebase.

DonutEspresso: 👍

lib/plugins/staticFiles.js
DonutEspresso: Cool, makes sense. This might be paranoia but if it's not too expensive, you could think about making a new copy of opts when we call `send` but with only send specific options. It's nice to be able to separate internal opts from send opts. At the minimum - worth calling out in the jsdocs what's used by us vs used by `send`.

PR #1755, opened 2019-03-04
fix(dev): upgrading modules including restify-errors

rajatkumar:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #
* Issue #
* Issue #

# Changes

> What does this PR do?
This PR upgrades the dependencies to use the latest versions. This also cleans up a couple of tests, and includes minor changes to and ensures `restify` uses the latest `restify-error` library.
Major libraries updated:
- mocha
- lru-cache
- restify-error
- mime
- vasync
- spdy


lib/server.js
```diff
@@ -1164,7 +1165,8 @@ Server.prototype._setupRequest = function _setupRequest(req, res) {
     // we consider a closed request as flushed from metrics point of view
     function onReqAborted() {
         // Request was aborted, override the status code
-        var err = new errors.RequestCloseError();
+        //var err = new errors.RequestCloseError();
```
misterdjules: Why do we want to keep this comment?

test/server.test.js
```diff
@@ -10,7 +10,7 @@ var http = require('http');
 var stream = require('stream');
 
 var errors = require('restify-errors');
-var filed = require('filed');
+// var filed = require('filed');
```
misterdjules: Why do we want to keep this commented out (as well as the test below)?

lib/errorTypes.js
```diff
@@ -2,5 +2,8 @@
 
 var errors = require('restify-errors');
 
-errors.makeConstructor('RequestCloseError');
-errors.makeConstructor('RouteMissingError');
+// This allows Restify to work with restify-errors v6+
```
misterdjules: Just out of curiosity: what is the change in restify v6+ that requires this?

lib/plugins/accept.js
```diff
@@ -38,7 +38,7 @@ function acceptParser(accepts) {
             return a;
         })
         .map(function map(a) {
-            return a.indexOf('/') === -1 ? mime.lookup(a) : a;
+            return a.indexOf('/') === -1 ? mime.getType(a) : a;
```
misterdjules: What is the difference between `mime.lookup` and `mime.getType`?

lib/errorTypes.js
rajatkumar: In v5 of `restify-errors` the function call to `makeConstructor` added the constructor to `restify-errors` exports. This changed in v6 and the recommendation was that the consumer should keep a reference to the new error constructor.

lib/plugins/accept.js
rajatkumar: `lookup` was renamed to `getType` in the newer version of `mime`.

test/server.test.js
rajatkumar: I will get rid of it, this is not needed.

lib/server.js
rajatkumar: I will get rid of it, this is not needed.

PR #1757, opened 2019-03-13
fix: properly handle non-errors thrown in domains

DonutEspresso:
## Pre-Submission Checklist

## Issues

Restify doesn't properly handle non-errors being thrown inside the domain. The two tests augmented in this PR are currently passing unintentionally. When you dump the response sent back from restify, they actually show restify blowing up while attempting to handle the domain, and not the original non-error value thrown in the domain: `TypeError: Cannot read property \'statusCode\' of undefined` and `AssertionError [ERR_ASSERTION]: first argument to VError, SError, or WError constructor must be a string, object, or Error (string) is required'`. 

I think this behavior is unintentional. @misterdjules I'm not sure if this is related to your previous domain findings, but it appears the domain is able to re-catch the secondary exception thrown by restify code attempting to handle the original user land exception. 

# Changes

This PR fixes the restify code and does best effort at trying to proxy the non-error values thrown by the domain back to the client.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -1303,25 +1303,24 @@ Server.prototype._routeErrorResponse = function _routeErrorResponse(
         return;
     }
 
-    self._emitErrorEvents(req, res, null, err, function emitError(emitErr) {
+    self._emitErrorEvents(req, res, null, err, function emitError() {
         // Prevent double handling
         if (res._sent) {
             return;
         }
 
         // only automatically send errors that are known (e.g., restify-errors)
-        if (err && _.isNumber(err.statusCode)) {
+        if (err instanceof Error && _.isNumber(err.statusCode)) {
             res.send(err);
             return;
         }
 
-        var finalErr = emitErr || err;
-        if (!(finalErr instanceof Error)) {
-            // if user land domain threw a string or number or any other
-            // non-error object, handle it here as best we can.
-            finalErr = (finalErr && finalErr.toString()) || '';
-        }
-        res.send(new errors.InternalError(finalErr));
+        // if the thrown exception is not really an Error object, e.g.,
+        //    "throw 'foo';"
+        // try to do best effort here to pass on that value by casting it to a
+        // string. This should work even for falsy values like 0, false, null,
+        // or undefined.
+        res.send(new errors.InternalError(String(err)));
     });
 };
 
--- a/lib/server.js
+++ b/lib/server.js
@@ -1309,13 +1309,19 @@ Server.prototype._routeErrorResponse = function _routeErrorResponse(
             return;
         }
 
-        // Expose only knonw errors
-        if (_.isNumber(err.statusCode)) {
+        // only automatically send errors that are known (e.g., restify-errors)
+        if (err && _.isNumber(err.statusCode)) {
             res.send(err);
             return;
         }
 
-        res.send(new errors.InternalError(emitErr || err));
+        var finalErr = emitErr || err;
+        if (!(finalErr instanceof Error)) {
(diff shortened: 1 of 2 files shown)


test/server.test.js
```diff
@@ -2505,9 +2505,10 @@ test('uncaughtException should handle thrown undefined literal', function(t) {
         }
     );
 
-    CLIENT.get('/foo', function(err, _, res) {
+    CLIENT.get('/foo', function(err, _, res, data) {
         t.ok(err);
         t.equal(res.statusCode, 500);
+        t.equal(data.message, '');
```
DonutEspresso: I guess we could technically send `'undefined'` over the wire, but don't feel strongly.

hekike: well if you do `throw undefined`, you have a different problem as well

lib/server.js
```diff
@@ -1309,13 +1309,19 @@ Server.prototype._routeErrorResponse = function _routeErrorResponse(
             return;
         }
 
-        // Expose only knonw errors
-        if (_.isNumber(err.statusCode)) {
+        // only automatically send errors that are known (e.g., restify-errors)
+        if (err && _.isNumber(err.statusCode)) {
```
retrohacker: This will evaluate to true if the user does `throw true;`

Not sure we are wanting to protect against folks throwing `Boolean`s.

retrohacker: This will be followed for `throw res;`, I suspect that is okay?

lib/server.js
```diff
@@ -1309,13 +1309,19 @@ Server.prototype._routeErrorResponse = function _routeErrorResponse(
...
             res.send(err);
             return;
         }
 
-        res.send(new errors.InternalError(emitErr || err));
+        var finalErr = emitErr || err;
+        if (!(finalErr instanceof Error)) {
+            // if user land domain threw a string or number or any other
+            // non-error object, handle it here as best we can.
+            finalErr = (finalErr && finalErr.toString()) || '';
```
retrohacker: How does this behave if `finalErr` is `false` or `0`? Do we expect both of those to return `''`

misterdjules:
@DonutEspresso 

>  I'm not sure if this is related to your previous domain findings, but it appears the domain is able to re-catch the secondary exception thrown by restify code attempting to handle the original user land exception.

I think you're correct that it's related to some of my recent findings re: domains.

What may be happening here is that the same domain is pushed more than once on the stack. This can happen with domains attached to HTTP requests because generally the domain associated to that requests is also associated to e.g. the socket used for that request. Since every time an event is emitted on those objects their associated domain is entered, the domain is entered when any event is emitted on the socket and subsequently when any event is emitted on the request.

FWIW, I submitted a PR to node to fix that behavior at https://github.com/nodejs/node/pull/26211. I doubt it'll be merged anytime soon.

DonutEspresso:
@retrohacker thanks for the feedback, I redid this with some more robust logic that now reflects the thrown values.

lib/server.js
```diff
@@ -1303,19 +1303,24 @@ Server.prototype._routeErrorResponse = function _routeErrorResponse(
         return;
     }
 
-    self._emitErrorEvents(req, res, null, err, function emitError(emitErr) {
+    self._emitErrorEvents(req, res, null, err, function emitError() {
```
retrohacker: What happened to `emitError`?

DonutEspresso: I looked at `_emitErrorEvents`, and `emitErr is actually the same object as `err`,  just passed back. It seemed confusing as heck, so I removed it outright. In retrospect, I should probably remove calling this callback w/ err to begin with.

Issue #1765, opened 2019-03-26
Restify inflight request counter leaks when using domains and throwing an error after connection is aborted


cprussin:
- [x] Used appropriate template for the issue type
- [x] Searched both open and closed issues for duplicates of this issue
- [x] Title adequately and _concisely_ reflects the feature or the bug

# Bug Report

## Restify Version

Proven reproducible on 8.2.0 and 7.3.0

## Node.js Version

Proven reproducible on v8.15.0 and v10.15.1

## Expected behaviour

Restify inflight counter should not leak

## Actual behaviour

Restify inflight counter leaks

## Repro case

```js
const restify = require('restify');
const http = require('http');
const { InternalServerError } = require('restify-errors');

const server = restify.createServer({ handleUncaughtExceptions: true });

server.pre((req, res, next) => {
    console.log(`Inflight requests: ${server.inflightRequests()}`);
    next();
});

server.get('/', (req, res, next) => {
    setTimeout(() => { throw new Error('foo'); }, 20);
});

server.on('uncaughtException', (req, res) => res.send('foo'));

server.listen(8080, () => {
    console.log('Server now up on port 8080');
});

setInterval(() => {
    const req = http.get("http://localhost:8080/").on('error', () => {});
    setTimeout(() => req.abort(), 10);
}, 20);
```

## Cause

Something in Restify's domain handling causes this function to never get called with `res._handlersFinished === true`: https://github.com/restify/node-restify/blob/f363b36b0e5176c3f134b345c057ec6199d70b83/lib/server.js#L1347

I'm still working out the precise bug here but I should be able to trace more precisely what the bug is from there and update this ticket with more details.

## Are you willing and able to fix this?

Yes (but extra eyes are appreciated)

cprussin:
FYI @mridgway @misterdjules @jdarren

PR #1766, opened 2019-03-26
feat: provide callback to uncaughtException handler

cprussin:
## Pre-Submission Checklist

## Issues

Closes:

* Issue https://github.com/restify/node-restify/issues/1765

The 'after' event is not called when a request has an uncaught exception (using `handleUncaughtExceptions`) if the request has been aborted before the exception.

# Changes

> What does this PR do?

This PR is a proposal for a mechanism to fix the issue by adding a continuation function to the `uncaughtException` handler.  If this continuation function isn't called, then restify can never know that a request has completed all work and call the 'after' handler (because `aborted` is triggered on the request object before work has completed, so neither `finish` nor `close` ever get triggered on the response and thus we have no other signal that work has completed).  This is actually consistent with how middlewares work--if `next` isn't called then the accounting will leak--so I believe the tradeoff is acceptable.

This pattern also allows for a fix without making a breaking API change.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -1412,6 +1412,16 @@ Server.prototype._routeErrorResponse = function _routeErrorResponse(
             req.route,
             err,
             function uncaughtExceptionCompleted() {
+                // We provide a callback to listeners of the 'uncaughtException'
+                // event and we call _finishReqResCycle when that callback is
+                // called so that, in case the actual request/response lifecycle
+                // was completed _before_ the error was thrown or emitted, and
+                // thus _before_ route handlers were marked as "finished", we
+                // can still mark the req/res lifecycle as complete.
+                // This edge case can occur when e.g. a client aborts a request
+                // and the route handler that handles that request throws an
+                // uncaught exception _after_ the request was aborted and the
+                // response was closed.
                 self._finishReqResCycle(req, res, err);
             }
         );
--- a/lib/server.js
+++ b/lib/server.js
@@ -1405,7 +1405,16 @@ Server.prototype._routeErrorResponse = function _routeErrorResponse(
         self.handleUncaughtExceptions &&
         self.listenerCount('uncaughtException') > 1
     ) {
-        self.emit('uncaughtException', req, res, req.route, err);
+        self.emit(
+            'uncaughtException',
+            req,
+            res,
+            req.route,
+            err,
+            function uncaughtExceptionCompleted() {
+                self._finishReqResCycle(req, res, err);
+            }
+        );
         return;
     }
 
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -2051,31 +2051,48 @@ test(
     }
 );
 
+// This test reproduces https://github.com/restify/node-restify/issues/1765. It
+// specifically tests the edge case of an exception being thrown from a route
+// handler _after_ the response is considered to be "flushed" (for instance when
+// the request is aborted before a response is sent and an exception is thrown).
 // eslint-disable-next-line max-len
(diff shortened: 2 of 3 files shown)


misterdjules:
@cprussin I think this approach looks good to me. Two things I'd like to see as part of this PR and that are currently missing:

1. New tests that make sure that inflight requests count is correctly tracked in the uncaught exception use case. Adding tests to https://github.com/restify/node-restify/blob/master/test/server.test.js#L2054 may be a good place to add those tests.
2. Updating documentation in `.md` files to describe this new callback.

test/server.test.js
```diff
@@ -2051,6 +2051,34 @@ test(
...
+test("should emit 'after' on client closed request with an uncaughtException", function(t) {
+    SERVER.on('after', function(req, res, route, err) {
+        t.ok(err);
+        t.equal(req.connectionState(), 'close');
+        t.equal(res.statusCode, 444);
+        t.equal(err.name, 'Error');
+        t.end();
+    });
+
+    SERVER.on('uncaughtException', function(req, res, route, err, next) {
```
misterdjules: In the documentation, I would suggest not calling that callback `next`, so that users don't confuse this callback with middlewares' next callbacks and their specific semantics/API.

test/server.test.js
```diff
@@ -2051,6 +2051,34 @@ test(
...
+        next();
+    });
+
+    SERVER.get('/foobar', function(req, res, next) {
+        // fast client times out at 500ms, wait for 800ms which should cause
+        // client to timeout
+        setTimeout(function() {
+            throw new Error('foo');
+        }, 800);
```
misterdjules: Can we rely on a 'timeout' event on `FAST_CLIENT` instead of a timer duration to make sure that we throw after the client times out and abort the request? Or maybe even relying on an `abort` event on the client would be better? Or even better on the 'abort' event of the request object?

Otherwise it seems this test could be broken easily by changes in the client's implementation details.

lib/server.js
```diff
@@ -1405,7 +1405,16 @@ Server.prototype._routeErrorResponse = function _routeErrorResponse(
...
         self.listenerCount('uncaughtException') > 1
     ) {
-        self.emit('uncaughtException', req, res, req.route, err);
+        self.emit(
+            'uncaughtException',
+            req,
+            res,
+            req.route,
+            err,
+            function uncaughtExceptionCompleted() {
```
hekike: Please put a comment about the why here and describe what will happen with and without domains.

test/server.test.js
cprussin: Could do, I mostly copied this pattern from existing tests so I'm not sure if you'd prefer consistency with those tests over stability for this test?

rajatkumar: I prefer stability here. Your stable test can act as a good example of how to correctly write tests and that pattern can then be used for future tests. I think changing the other tests to follow your pattern of stable tests can also be a good first issue for new contributors.

lib/server.js
rajatkumar: +1

misterdjules: @hekike @rajatkumar Added some comments here, let me know if that addresses your feedback!

test/server.test.js
misterdjules: @cprussin @rajatkumar I refactored the test so that it doesn't rely on any arbitrary timeout, but instead relies on the actual state of the request/response lifecycle. Let me know if that looks good to you!

misterdjules:
@cprussin @rajatkumar @hekike Updated this PR with a few more commits that I believe address all the comments in this PR. Please take another look!

Issue #1768, opened 2019-04-06
throwing error in route handler after request aborted leaks inflight requests


misterdjules:
- [x] Used appropriate template for the issue type
- [x] Searched both open and closed issues for duplicates of this issue
- [x] Title adequately and _concisely_ reflects the feature or the bug

<!--

# Bug Report

## Restify Version

Latest master.

## Node.js Version

10.15.1.

## Expected behaviour

> This section details what you expected restify to do based on the code that you wrote

When throwing an error in a route handler and *not* listening on the `'uncaughtException'` event, the `'after'` event is emitted.

This is similar to the issue described at https://github.com/restify/node-restify/issues/1765, except for the fact that in this case the application's code is *not* setting an event listener for the `'uncaughtException'` event.

## Actual behaviour

When throwing an error in a route handler and not listening on the `'uncaughtException'` event, the `'after'` event is *not* emitted.

## Repro case

The following test added to test/server.test.js fails:

```
diff --git a/test/server.test.js b/test/server.test.js
index ba19cbb..31e580c 100644
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -2051,6 +2051,40 @@ test(
     }
 );
 
+// eslint-disable-next-line max-len
+test("should emit 'after' on uncaughtException after response closed without custom uncaughtException listener", function(t) {
+    var ERR_MSG = 'foo';
+    var gotAfter = false;
+    var gotReqCallback = false;
+
+    SERVER.on('after', function(req, res, route, err) {
+        gotAfter = true;
+        t.ok(err);
+        t.equal(req.connectionState(), 'close');
+        t.equal(res.statusCode, 444);
+        t.equal(err.name, 'Error');
+        t.equal(err.message, ERR_MSG);
+        if (gotReqCallback) {
+            t.end();
+        }
+    });
+
+    SERVER.get('/foobar', function(req, res, next) {
+        res.on('close', function onResClose() {
+            throw new Error(ERR_MSG);
+        });
+    });
+
+    FAST_CLIENT.get('/foobar', function(err, _, res) {
+        gotReqCallback = true;
+        t.ok(err);
+        t.equal(err.name, 'RequestTimeoutError');
+        if (gotAfter) {
+            t.end();
+        }
+    });
+});
+
 test('should increment/decrement inflight request count', function(t) {
     SERVER.get('/foo', function(req, res, next) {
         t.equal(SERVER.inflightRequests(), 1);
```

## Cause

[...]

PR #1769, opened 2019-04-06
breaking: emit after on uncaught exception with no custom handler

misterdjules:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #1768

When throwing an error in a route handler and not listening on the `'uncaughtException'` event, the `'after'` event is *not* emitted.

# Changes

The changes in this  PR:
* always emit the `'uncaughtException'` event, even when the application's code does not add a listener for it. This allows the internal `'uncaughtException'` event handler to run. I'm not sure this is a good thing.
* check whether the response is already flushed before sending a default error response. If it is, it forces finishing the request/response lifecycle, as nothing else would do that otherwise.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,6 +2,7 @@
 
 'use strict';
 
+var _ = require('lodash');
 var assert = require('assert-plus');
 var errors = require('restify-errors');
 
@@ -90,8 +91,13 @@ function createServer(options) {
             req,
             res,
             route,
-            e
+            e,
+            cb
         ) {
+            // If there _is_ at least one user-defined 'uncaughtException' event
+            // listener set on the server instance, do not perform the default
+            // handling mechanism for this uncaught exception, and instead leave
+            // that to the user-defined handlers.
             if (
                 this.listeners('uncaughtException').length > 1 ||
                 res.headersSent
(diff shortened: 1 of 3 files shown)


DonutEspresso:
Ah, looks like this needs to be updated to handle the new callback based interface.

misterdjules:
@DonutEspresso 

> Ah, looks like this needs to be updated to handle the new callback based interface.

Done, can you please take another look?

DonutEspresso:
Looks like some formatting failures w/ prettier but otherwise LGTM!

misterdjules:
> Looks like some formatting failures w/ prettier but otherwise LGTM!

🤦‍♂ I ran make test instead of make prepush. Fixed, thank you!

ghermeto:
LGTM. @restify/current-core, can I merge or any objections?

mmarchini:
No objections on my side

ghermeto (approved):

ghermeto:
@wesleytodd @mmarchini:

@misterdjules said this is ok to be merged, but I think we need to run the tests again (just to be safe), only that this is from his fork... any ideas?

misterdjules:
Closing as part of cleaning up my list of inflight PRs, and because I don't think this work is relevant anymore.

PR #1772, opened 2019-04-12
fix(test): make upgrade test pass

DonutEspresso:
## Pre-Submission Checklist

## Issues

Two issues. 

1) Upgrade tests are failing locally (and likely on CI), but we didn't know about it. Travis isn't running all of the nodeunit tests for some reason (??). We need to investigate.

2) Something has changed in recent Node 10 release, perhaps a lifecycle ordering change or something else like we saw previously. The upgrade test is currently set up in a way that requires proper response lifecycle cleanup to exit properly. This doesn't seem to be happening, as calling `res.resume()` on an upgraded request isn't triggering the correct behavior to allow the server to "close/resume" the response from restify, which leaves the client hanging. 

# Changes

I updated the test to exit the test earlier and not depend on proper response cleanup, as I suspect in the field we wouldn't really want to be encouraging the kind of scenarios exhibited by this test anyway. This kind of usage should be fatal. 

That said, it's probably worth digging further into what's causing this breakage between Node8 => Node10. There may be other issues we uncover along the way.

### DIFF
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -34,7 +34,7 @@ var FAST_CLIENT;
 var SERVER;
 
 if (SKIP_IP_V6) {
-    console.warning('IPv6 tests are skipped: No IPv6 network is available');
+    console.warn('IPv6 tests are skipped: No IPv6 network is available');
 }
 
 ///--- Tests
--- a/test/upgrade.test.js
+++ b/test/upgrade.test.js
@@ -160,7 +160,8 @@ test('GET without upgrade headers', function(t) {
 
 test('Dueling upgrade and response handling 1', function(t) {
     var done = finish_latch(t, {
-        'expected requestUpgrade error': 1
+        'expected requestUpgrade error': 1,
+        'client response': 1
     });
 
     SERVER.get('/attach', function(req, res, next) {
@@ -173,9 +174,7 @@ test('Dueling upgrade and response handling 1', function(t) {
         }
 
         try {
-            var upg = res.claimUpgrade();
-            // TODO we never reach this destroy() call. do we still need it?
-            upg.socket.destroy();
+            res.claimUpgrade();
         } catch (ex) {
             done('expected requestUpgrade error');
         }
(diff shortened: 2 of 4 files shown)


test/server.test.js
```diff
@@ -34,7 +34,7 @@ var FAST_CLIENT;
 var SERVER;
 
 if (SKIP_IP_V6) {
-    console.warning('IPv6 tests are skipped: No IPv6 network is available');
+    console.warn('IPv6 tests are skipped: No IPv6 network is available');
```
DonutEspresso: @hekike :joy:

DonutEspresso: This is why Travis was skipping the rest of the tests. Not sure why this wouldn't cause an error code though.

hekike: ooups, VW

DonutEspresso:
Looks like this commit in 10.10.0 is likely related:
https://github.com/nodejs/node/blob/v10.10.0/lib/_stream_readable.js#L864

It seems calling resume won't do anything unless I have a `data` listener attached to the stream. I haven't dug too deeply yet into why this change was made. I'll make the requisite changes in this repo here to get us back into a good state.

test/upgrade.test.js
```diff
@@ -174,8 +174,7 @@ test('Dueling upgrade and response handling 1', function(t) {
         }
 
         try {
-            var upg = res.claimUpgrade();
-            upg.socket.destroy();
```
DonutEspresso: This was removed because the `claimUpgrade()` call is what throws, so this line is never executed.

DonutEspresso:
> 1) Upgrade tests are failing locally (and likely on CI), but we didn't know about it. Travis isn't running all of the nodeunit tests for some reason (??). We need to investigate. 

This is because of a typo `console.warning` in a nodeunit test that caused the test runner to bail with an exit code of 0. As nodeunit is deprecated, I'm not sure if it's worth the time to investigate this more deeply. We should prioritize migrating those tests onto a new test runner.
 
> 2) Something has changed in recent Node 10 release, perhaps a lifecycle ordering change or something else like we saw previously. The upgrade test is currently set up in a way that requires proper response lifecycle cleanup to exit properly. ~This doesn't seem to be happening, as calling `res.resume()`~ on an upgraded request isn't triggering the correct behavior to allow the server to "close/resume" the response from restify, which leaves the client hanging.

The test did not have a listener for the `data` event. Without it, it appears that calling `resume()` was not doing anything. Worth digging more to see if it's a problem with the way the test is setup, vs an unintentional change in a non breaking release of Node (10.10.0).

Issue #1774, opened 2019-04-12
Restify hangs when ending route chain without closing response


cprussin:
- [x] Used appropriate template for the issue type
- [x] Searched both open and closed issues for duplicates of this issue
- [x] Title adequately and _concisely_ reflects the feature or the bug

# Bug Report

## Restify Version

Tested on 8.3.0.

## Node.js Version

Tested on v8.15.1.

## Expected behaviour

Restify should not hang (in express this same case automatically 404s)

## Actual behaviour

Restify hangs

## Repro case

```js
const restify = require('restify');
const server = restify.createServer();
server.get('/', (req, res, next) => next()); // also could be server.get('/', () => {});
server.listen(8080, () => console.log('Server up on port 8080'));
```

## Cause

Restify does not detect cases where the middleware chain is completed and no response has been flushed.

## Are you willing and able to fix this?

Yes

retrohacker:
Comparing this to express' behavior:

Express uses https://github.com/pillarjs/finalhandler at the end of the middleware chain. It does a check to see if headers have started being flushed yet as a heuristic.

This snippet 404s:
```js
require('express')().get('/', function(req, res, next) { next(); }).listen(8080);
```

This snippet hangs:
```js
require('express')().get('/', function(req, res, next) { res.flushHeaders(); next(); }).listen(8080);
```

It also requires calling the continuation, otherwise it hangs:

```js
require('express')().get('/', function(req, res, next) {}).listen(8080);
```

ghermeto:
merged :)

PR #1777, opened 2019-04-20
feat: send 500s for unhandled requests

DonutEspresso:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #1774 

# Changes

If we reach the end of a handler chain and the headers haven't been sent yet, send a 500 back to the client.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -1352,7 +1352,10 @@ Server.prototype._finishReqResCycle = function _finishReqResCycle(
     var self = this;
     var route = req.route; // can be undefined when 404 or error
 
-    if (res._finished) {
+    // if the returned err value was a string, then we're handling the
+    // next('foo') case where we redirect to another middleware stack. don't
+    // do anything here because we're not done yet.
+    if (res._finished || _.isString(err)) {
         return;
     }
 
@@ -1366,6 +1369,22 @@ Server.prototype._finishReqResCycle = function _finishReqResCycle(
         var finalErr = err || res.err;
         req.emit('restifyDone', route, finalErr);
         self.emit('after', req, res, route, finalErr);
+    } else if (
+        res._handlersFinished === true &&
+        res.headersSent === false &&
+        !res.err
+    ) {
+        // if we reached the end of the handler chain and headers haven't been
+        // sent AND there isn't an existing res.err (e.g., req abort/close),
+        // it's possible it's a user error and a response was never written.
+        // send a 500.
+        res.send(
+            new errors.InternalServerError(
+                'reached the end of the handler chain without ' +
+                    'writing a response!'
+            )
+        );
+        return;
     } else {
         // Store error for when the response is flushed and we actually emit the
         // 'after' event. The "err" object passed to this method takes
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -2497,6 +2497,48 @@ test('should emit error with multiple next calls with strictNext', function(t) {
     });
 });
 
+test(
+    'should send 500 if we reached the end of handler chain w/o sending ' +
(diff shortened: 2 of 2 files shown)


lib/server.js
```diff
@@ -1352,7 +1352,10 @@ Server.prototype._finishReqResCycle = function _finishReqResCycle(
     var self = this;
     var route = req.route; // can be undefined when 404 or error
 
-    if (res._finished) {
+    // if the returned err value was a string, then we're handling the
+    // next('foo') case where we redirect to another middleware stack. don't
+    // do anything here because we're not done yet.
+    if (_.isString(err)) {
```
DonutEspresso: Probably this check should not happen in the `finishReqResCycle` method itself, but this method is called from multiple places so rather than gate the check we'll just bail here. I think the old check here worked because we didn't have to handle this use case, but now we have to be more explicit.

lib/server.js
```diff
@@ -1366,6 +1369,22 @@ Server.prototype._finishReqResCycle = function _finishReqResCycle(
...
+        res.headersSent === false &&
+        !res.err
+    ) {
+        // if we reached the end of the handler chain and headers haven't been
+        // sent AND there isn't an existing res.err (e.g., req abort/close),
+        // it's possible it's a user error and a response was never written.
+        // send a 500.
+        res.send(
+            new errors.InternalServerError(
+                'reached the end of the handler chain without ' +
```
retrohacker: Doesn't this get sent back to the client as:

```
{ "code": "Internal Server Error", "message": "reached the end of the handler chain without writing a response!" }
```

Have we thought about how much this reveals about the server?

lib/server.js
```diff
@@ -1366,6 +1369,22 @@ Server.prototype._finishReqResCycle = function _finishReqResCycle(
...
+    } else if (
+        res._handlersFinished === true &&
+        res.headersSent === false &&
+        !res.err
+    ) {
+        // if we reached the end of the handler chain and headers haven't been
+        // sent AND there isn't an existing res.err (e.g., req abort/close),
+        // it's possible it's a user error and a response was never written.
+        // send a 500.
+        res.send(
```
retrohacker: Should we also do the stuff from the previous if block here? Decrement inflight requests, emitting after, etc?

It seems like we are asserting along this path that this is an error, why wouldn't we clean up?

DonutEspresso: Calling `res.send` will actually loop back to this method again, It's probably more accurate to rename `finishReqResCycle` to `attemptFinishReqResCycle`, as we call it optimistically from various points in the code base and let this method handle the state reconciliation.

lib/server.js
DonutEspresso: Good point. I don't feel super strongly about this, and its worth noting we already send back pretty specific error messages for various other scenarios (bad redirects, bad formatters, etc). The messages are scrubbable using the error events API but makes me think we could do a better job w/ standardized error codes for restify.

DonutEspresso:
The current behavior is no response; I'd be surprised if folks built their systems around a server that didn't respond but I guess anything is possible! This is one of those scenarios where I'm actually ok with this not being breaking, but would be curious to get other's thoughts.

DonutEspresso:
cc @cprussin

cprussin:
Sweet!  Thanks for the heads up

lib/server.js
ghermeto: I don't think it reveals anything that shouldn't be fixed. Maybe a more generic message like "request finished without response", or "sever didn't send response" is enough.

DonutEspresso:
@ghermeto merge away!

PR #1786, opened 2019-05-15
feat: adds support to http problem details - rfc7807

ghermeto:
## Pre-Submission Checklist

# Changes

> What does this PR do?

adds supports to handle error objects conforming with the [http
problem details spec (rfc7807)](https://tools.ietf.org/html/rfc7807).
 
The spec defines a `status` field (instead of `statusCode`) and a 
`application/problem+json` content type. Code will prefer the current 
behavior of sending `application/json` if `statusCode` is present.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -389,7 +389,7 @@ function patch(Response) {
         // If the body is an error object and we were not given a status code,
         // try to derive it from the error object, otherwise default to 500
         if (!code && body instanceof Error) {
-            code = body.statusCode || 500;
+            code = body.statusCode || body.status || 500;
         }
 
         // Set sane defaults for optional arguments if they were not provided
@@ -441,7 +441,12 @@ function patch(Response) {
             // Set Content-Type to application/json when
             // res.send is called with an Object instead of calling res.json
             if (!type && typeof body === 'object' && !Buffer.isBuffer(body)) {
-                type = 'application/json';
+                // if Error that conforms with rfc7807 use the correct type
+                if (body instanceof Error && body.status && !body.statusCode) {
+                    type = 'application/problem+json';
+                } else {
+                    type = 'application/json';
+                }
             }
 
             // Derive type if not provided by the user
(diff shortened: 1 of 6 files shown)


lib/server.js
```diff
@@ -1435,7 +1435,7 @@ Server.prototype._routeErrorResponse = function _routeErrorResponse(
         }
 
         // only automatically send errors that are known (e.g., restify-errors)
-        if (err instanceof Error && _.isNumber(err.statusCode)) {
+        if (err instanceof Error && _.isNumber(err.statusCode || err.status)) {
```
hekike: when is this `status`? I 
`restify-errors` does `statusCode`
https://github.com/restify/errors/blob/master/lib/baseClasses/HttpError.js#L54

lib/response.js
```diff
@@ -389,7 +389,7 @@ function patch(Response) {
         // If the body is an error object and we were not given a status code,
         // try to derive it from the error object, otherwise default to 500
         if (!code && body instanceof Error) {
-            code = body.statusCode || 500;
+            code = body.statusCode || body.status || 500;
```
hekike: when is this `status`?

lib/server.js
ghermeto: exactly, and `restify-errors` doesn't conform to rfc7807, which defines `status`. The problem is that it assumes that users are indeed using `restify-errors` and don't give them the ability to follow the standard.

This PR keeps `statusCode` for backward compatibility with `restify-errors` but allows restify to return the correct content type `application/problem+json` when `status` is used and no `statusCode` is present

ghermeto:
```js
class NotFound extends Error {
  constructor(detail) {
    this.title = 'Not Found';
    this.type = 'https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404';
    this.status = 404;
    this.detail = detail || 'Resource not found';
  }
  
  toJSON() {
    return this;
  }
}
```

this will conform with rfc7807 and will hit that code with `status`

ghermeto: You know what? It doesn't really need to "have a status property instead of a statusCode" to conform to the rfc. I was mainly using `status` property as a way to identify that was an error message complying with the rfc and sending the correct content-type.

But we can have a `statusCode` property in the error object and on the `toJSON` method that I serialize it to `status` instead, but I would need to find a way to set the `application/problem+json` content type and still keep backward compatibility. Any thoughts on this?

PR #1789, opened 2019-06-04
chore(package): update

hekike:
Update dependencies.


hekike:
It solves all the non dev deps.

PR #1794, opened 2019-06-24
Bugfix + Improvement

carlosascari:
Currently there is a hyphen "-" that is added between the method & path, however it is removed by the next line with the regex replace that replaces/removes all non-word characters:

```
    if (!name) {
        name = method + '-' + path;  // adds hypen
        name = name.replace(/\W/g, '').toLowerCase(); // removes hyphen
    }
```
I don't think you should force users to create a new object in order to allow them to name a route when they can specify the route explicitly. The improvement I added basically allows anyone to use `next('<method> <path>');` to call an existing route within another route without having to name a route explicitly.

Currently this works:

```
server.get('/user', (req, res, next) => {
  res.json({ user: 'ascari' });
  next();
});

server.get('/user', (req, res, next) => {
  next('getuser');
});
```

The improvement makes the default/auto-name of a route dev-friendly

```
server.get('/user', (req, res, next) => {
  res.json({ user: 'ascari' });
  next();
});

server.get('/user', (req, res, next) => {
  next('get /user');
});
```

This way, for non-dupe routes, you can call another existing route without having to name it.

## Pre-Submission Checklist

## Issues

Closes:

* Issue #
* Issue #
* Issue #

# Changes

> What does this PR do?

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -357,8 +357,7 @@ Router.prototype.defaultRoute = function defaultRoute(req, res, next) {
 Router.prototype._getRouteName = function _getRouteName(name, method, path) {
     // Generate name
     if (!name) {
-        name = method + '-' + path;
-        name = name.replace(/\W/g, '').toLowerCase();
+        name = (method + ' ' + path).toLowerCase();
     }
 
     // Avoid name conflict: GH-401


ghermeto:
I'm now closing this PR based on:
- the **submission checklist was not completed** 
- there are no tests
- the PR is over 18 months old

If this change interests someone, please: 
- link to an issue
- add comprehensive tests

PR #1798, opened 2019-07-26
feat(chain): schedule handlers to the next tick

hekike:
Schedule handlers to nextTick for cleaner Flamegraphs.
Shouldn't have an effect on performance:

```
? Do you want to compare HEAD with the stable release (7.2.0)? Yes
? How many connections do you need? 100
? How many pipelining do you need? 10
? How long does it take? 30
---- response-json ----
⠋ Started head/response-json
response-json throughput:
{
    "significant": "",
    "equal": true
}

---- response-text ----
response-text throughput:
{
    "significant": "",
    "equal": true
}

---- router-heavy ----
router-heavy throughput:
{
    "significant": "***",
    "equal": true
}

---- middleware ----
middleware throughput:
{
    "significant": "",
    "equal": true
}

```

### DIFF
--- a/lib/chain.js
+++ b/lib/chain.js
@@ -161,7 +161,9 @@ function call(handler, err, req, res, _next) {
         return;
     } else if (!hasError && arity < 4) {
         // request-handling middleware
-        handler(req, res, next);
+        process.nextTick(function nextTick() {
+            handler(req, res, next);
+        });
         return;
     }
 
--- a/test/chain.test.js
+++ b/test/chain.test.js
@@ -1,6 +1,7 @@
 'use strict';
 /* eslint-disable func-names */
 
+var domain = require('domain');
 var Chain = require('../lib/chain');
 
 if (require.cache[__dirname + '/lib/helper.js']) {
@@ -184,7 +185,7 @@ test('onceNext prevents double next calls', function(t) {
 });
 
 test('throws error for double next calls in strictNext mode', function(t) {
-    var doneCalled = 0;
+    t.expect(1);
     var chain = new Chain({
         strictNext: true
     });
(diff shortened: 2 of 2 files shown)


lib/chain.js
```diff
@@ -161,7 +161,9 @@ function call(handler, err, req, res, _next) {
         return;
     } else if (!hasError && arity < 4) {
         // request-handling middleware
-        handler(req, res, next);
+        process.nextTick(function nextTick() {
```
kenashcraft: Have you run the benchmarks with this PR?  Consider this implementation instead:
process.nextTick(handler, req, res, next);
Looking at the implementation of `nextTick` with Julien, it isn't clear which would be faster.

DonutEspresso: Is this enough to reset the FG stack? Or do we need setImmediate?

hekike: yes I run, looks like no effect

hekike: The goal would be cleaner flamegraphs not faster execution.

hekike: cc @misterdjules

hekike:
@DonutEspresso why do you think it's a breaking change? It doesn't affect any of the public APIs or execution order.

DonutEspresso:
If it's nextTick it's probably not breaking - if it's setImmediate then I'm thinking in a worst case scenario this could cause the queueing behavior to change, e.g., the number of inflight requests may increase and latency gets amortized across the requests in the queue.

lib/chain.js
misterdjules: What do we mean by "FG stack"? If the question is whether the `handler` call's stack frame would not have the previous handler as a parent stack frame, the answer is yes, it would not have the previous handler as a parent stack frame.

misterdjules: Ah Peter clarified that FG is for Flame Graphs :)

PR #1800, opened 2019-08-14
Use restify-errors 8.0.1

eltimn:
Yarn audit still shows restify-errors using lodash <4.17.12 which has a vuln.

I would have thought that your last release (8.4.0) would have used 8.0.1 because of the ^. But, I admit to not fully understanding how that works.

I didn't open an issue, as it seemed small enough to not warrant one, but I will if you'd like.

## Pre-Submission Checklist

## Issues

Closes:

* Issue #
* Issue #
* Issue #

# Changes

> What does this PR do?

Use restify-errors 8.0.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
     "once": "^1.4.0",
     "pidusage": "^2.0.17",
     "qs": "^6.7.0",
-    "restify-errors": "^8.0.0",
+    "restify-errors": "^8.0.1",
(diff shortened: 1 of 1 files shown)


mmarchini:
Thank you for your PR and sorry for the delay. Minimal restify-errors version has been bumped to 8.0.2 on the latest version, so the security warning should be fixed. Let me know if that's not the case.

PR #1807, opened 2019-11-08
Gracefully handle formatters throwing restify errors.

CrabDude:
Summary: Don't always blowup (uncaughtException) when we can avoid it, like in scenarios where a JSON payload does not properly parse or serialize.

Details:

When a formatter throws an error that is an instanceof restify-errors' `HttpError` or `RestError`, handle it gracefully by catching & passing to `formatterErrors`.

### DIFF
--- a/lib/formatters/json.js
+++ b/lib/formatters/json.js
@@ -23,7 +23,10 @@ function formatJSON(req, res, body) {
         try {
             data = JSON.stringify(body);
         } catch (e) {
-            throw new errors.InternalServerError(e);
+            throw new errors.InternalServerError(
+                { cause: e, info: { formatter: 'json' } },
+                'could not format response body'
+            );
         }
     }
 
--- a/lib/formatters/json.js
+++ b/lib/formatters/json.js
@@ -2,6 +2,8 @@
 
 'use strict';
 
+var errors = require('restify-errors');
+
 ///--- Exports
 
 /**
@@ -16,7 +18,15 @@
  * @returns  {String} data
  */
 function formatJSON(req, res, body) {
-    var data = body ? JSON.stringify(body) : 'null';
+    var data = 'null';
+    if (body) {
+        try {
+            data = JSON.stringify(body);
+        } catch (e) {
+            throw new errors.InternalServerError(e);
+        }
+    }
+
     // Setting the content-length header is not a formatting feature and should
     // be separated into another module
     res.setHeader('Content-Length', Buffer.byteLength(data));
--- a/lib/response.js
+++ b/lib/response.js
@@ -506,7 +506,22 @@ function patch(Response) {
                 // Finally, invoke the formatter and flush the request with it's
                 // results
 
-                return flush(self, formatter(self.req, self, body));
+                var formattedBody;
+                try {
+                    formattedBody = formatter(self.req, self, body);
+                } catch (e) {
+                    if (
+                        e instanceof errors.RestError ||
+                        e instanceof errors.HttpError
+                    ) {
+                        e.message =
+                            'could not format response body: ' + e.message;
+                        var res = formatterError(self, e);
+                        return res;
+                    }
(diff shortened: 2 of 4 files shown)


ghermeto:
Can you give example of use cases where this new behavior is more desirable than the old one?

CrabDude:
@ghermeto `JSON.stringify` fails when a body payload has either circular references or is too large. There are no easy/obvious ways to check for either.

With the deprecation of `domains` & `handleUncaughtExceptions`, the current default behavior is to bubble up to `process.on('uncaughtException')`, which prescribes a process restart, or at the very least loses error context. Converting these exceptions to errors allows restify to serve  responses (500s) in these cases.

CrabDude:
Removed an erroneous change to `lib/formatters/json.js`'s `formatJSON`'s JSDoc return type.

test/formatter.test.js
```diff
@@ -130,6 +134,30 @@ test('GH-845: sync formatter should blow up', function(t) {
     );
 });
 
+test('sync formatter should handle expected errors gracefully', function(t) {
+    var hasUncaughtException = false;
+    SERVER.on('uncaughtException', function(req, res, route, err) {
+        hasUncaughtException = true;
```
hekike: nit: How about adding an `assert.throw('Should not reach')` instead of the boolean logic which is more fragile to execution order.

lib/formatters/json.js
```diff
@@ -16,7 +18,15 @@
  * @returns  {String} data
  */
 function formatJSON(req, res, body) {
-    var data = body ? JSON.stringify(body) : 'null';
```
hekike: Could we add a test case for invalid JSON parsing?

lib/response.js
```diff
@@ -506,7 +506,22 @@ function patch(Response) {
...
-                return flush(self, formatter(self.req, self, body));
+                var formattedBody;
+                try {
+                    formattedBody = formatter(self.req, self, body);
+                } catch (e) {
+                    if (
+                        e instanceof errors.RestError ||
+                        e instanceof errors.HttpError
+                    ) {
+                        e.message =
```
DonutEspresso: @hekike do you recall if we are using n/verror in restify core? Should we wrap this instead of modifying the existing message?

hekike: NError via restify-errors https://github.com/restify/node-restify/blob/master/package.json#L111

hekike: I think if you wrap you loose original statusCode and body of the error. Which is maybe fine here, just wanted to call out.

DonutEspresso: Hmm, actually - thoughts on removing the try/catch in the formatter itself, but retaining the try/catch here? The InternalServerError wrapping could happen here and we can use NError to set the new error message too. I could see an argument for try/catch in the formatter _if_ the formatter had different types of errors, but I don't think we have a use case for that yet.

CrabDude: The reasoning behind the try/catch in the formatter is opt-in/consent to an error being catchable. While exceptions in formatters should semantically always be catchable due to an expectation/reasonable assumption of statelessness, no such guarantee exists unfortunately. Coupling these w/n `node-restify` I think is perfectly reasonable, but since 3rd party formatters are supported, the formatter error/exception handling I think needs this 2-layer approach since otherwise true exceptions from 3rd party formatters will not properly be treated as exceptions.

Given the above, expecting error wrapping in the formatter seems the most pragmatic approach for differentiating catchable errors vs true exceptions, and given this expectation of wrapping, additional wrapping w/n `response.js` seems superfluous.

CrabDude: Resolving with @hekike's recommendation to not mutate the error:
https://github.com/restify/node-restify/pull/1807/commits/84e456a066543731dd6e01bccc92f0b5af6cbe25

hekike:
Released as `restify@8.4.1`

PR #1808, opened 2019-11-25
feat(chain): use nextTick instead of setImmediate

mmarchini:
BREAKING CHANGE: Using setImmediate has more overhead than nextTick on
Node.js v12. Benchmarks show improvements of >20% when changing
setImmediate to nextTick:

```
  response-json throughput:
  {
      "significant": "***",
      "equal": false,
      "wins": "head",
      "diff": "28.62%"
  }

  ---- response-text ----
  response-text throughput:
  {
      "significant": "***",
      "equal": false,
      "wins": "head",
      "diff": "43.45%"
  }

  ---- router-heavy ----
  router-heavy throughput:
  {
      "significant": "***",
      "equal": false,
      "wins": "head",
      "diff": "30.02%"
  }

  ---- middleware ----
  middleware throughput:
  {
      "significant": "***",
      "equal": false,
      "wins": "head",
      "diff": "20.78%"
  }
```

This changes the order some events are processed (nextTick is processed
earlier than setImmediate), thus this can be considered a breaking
change as some edge cases will notice this (as we can see in the test
changes).

## Pre-Submission Checklist

### DIFF
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -1820,7 +1820,7 @@ test('calling next(false) should early exit from use handlers', function(t) {
 
     SERVER.on('after', function() {
         steps++;
-        t.equal(steps, 2);
+        t.equal(steps, 1);
         t.end();
     });
 
@@ -2111,7 +2111,7 @@ test('should increment/decrement inflight request count', function(t) {
     CLIENT.get('/foo', function(err, _, res) {
         t.ifError(err);
         t.equal(res.statusCode, 200);
-        t.equal(SERVER.inflightRequests(), 1);
+        t.equal(SERVER.inflightRequests(), 0);
     });
 });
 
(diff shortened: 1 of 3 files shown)


test/server.test.js
```diff
@@ -2111,7 +2111,7 @@ test('should increment/decrement inflight request count', function(t) {
     CLIENT.get('/foo', function(err, _, res) {
         t.ifError(err);
         t.equal(res.statusCode, 200);
-        t.equal(SERVER.inflightRequests(), 1);
+        t.equal(SERVER.inflightRequests(), 0);
```
retrohacker: I _think_ this is fine. Is this a result of `CLIENT` and `SERVER` sharing an event loop? Looking at this change, I don't see how it would cause a problem with our inflight request throttling. I could see the interaction between open request throttling and inflight request throttling changing with this PR, though I don't see a use case for both of those being turned on at the same time.

hekike:
We discussed that we will treat it as implementation details and making a nonbreaking release.

hekike:
Thanks @mmarchini!
released as `restify@8.5.0`

PR #1814, opened 2020-01-28
Preserve original error when there is no fallback formatter

CrabDude:
## Pre-Submission Checklist

## Issues

There is an edge case where `strictFormatters=true`, `content-type` is set without a supported formatter, and an error is passed. Right now the error will fall through to a raw serialization (`res.write`), which fails to serialize the error object.

# Changes

> What does this PR do?

Ensures the original error is preserved with an `InternalServerError`, instead of falling through to `res.write`.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -529,6 +529,19 @@ function patch(Response) {
                 }
                 return flush(self, formattedBody);
             }
+            // Preserve original error when
+            // Content-type is set, _strictFormatters=false, no formatter found
+            if (body instanceof Error) {
+                return formatterError(
+                    self,
+                    new errors.InternalServerError(
+                        { cause: body, info: { 'content-type': type } },
+                        'could not format Error for content-type: "' +
+                            type +
+                            '"'
+                    )
+                );
+            }
         }
 
         return flush(self, body);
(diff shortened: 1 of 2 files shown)


lib/response.js
```diff
@@ -529,6 +529,19 @@ function patch(Response) {
                 }
                 return flush(self, formattedBody);
             }
+            // Preserve original error when
+            // Content-type is set, _strictFormatters=false, no formatter found
+            if (body instanceof Error) {
```
DonutEspresso: Couldn’t this be something other than an error? Maybe check against what we assert in `flush()`?

retrohacker: I don't understand @DonutEspresso, isn't this PR specifically for preserving errors?

DonutEspresso: It's been awhile since I looked at this - I _think_ my question was if it was possible to get to this point in the function where body isn't an Error but just a POJO. IOW, does this new `if` clause cause a silent failure/hang where there's no formatter, and the body is not an Error?

DonutEspresso: Maybe there's already a test for this in which case carry on!

Issue #1819, opened 2020-03-16
multipartBodyParser crash if mapFiles is true


guoyibj:
- [ ] Used appropriate template for the issue type
- [ ] Searched both open and closed issues for duplicates of this issue
- [ ] Title adequately and _concisely_ reflects the feature or the bug

# Bug Report

## Restify Version
8.5.1
## Node.js Version
v10.15.1
## Expected behaviour
No crash

## Actual behaviour

Crash stack:
NodeError: The "path" argument must be one of type string, Buffer, or URL. Received type undefined
    at Object.readFile (fs.js:297:3)
    at forEach (/Users/guoyi/git/testServer/node_modules/restify/lib/plugins/multipartBodyParser.js:120:28)
    at Array.forEach (<anonymous>)
    at parse (/Users/guoyi/git/testServer/node_modules/restify/lib/plugins/multipartBodyParser.js:115:40)
    at IncomingForm.<anonymous> (/Users/guoyi/git/testServer/node_modules/formidable/lib/incoming_form.js:107:9)
    at IncomingForm.emit (events.js:189:13)
    at IncomingForm._maybeEnd (/Users/guoyi/git/testServer/node_modules/formidable/lib/incoming_form.js:557:8)
    at /Users/guoyi/git/testServer/node_modules/formidable/lib/incoming_form.js:238:12
    at WriteStream.<anonymous> (/Users/guoyi/git/testServer/node_modules/formidable/lib/file.js:79:5)
    at Object.onceWrapper (events.js:277:13)

### server side:
const restify = require('restify');

const server = restify.createServer({
  name: 'Test Server',
  maxParamLength: 1024,
});

server.use(restify.plugins.queryParser({ mapParams: false }));
server.use(restify.plugins.bodyParser({
  mapParams: true,
  mapFiles: true,
  multiples: true,
}));

server.post("/files/upload", async (request, response) => {
  return response.status(200).send(requst.url);
});

server.listen(3000, () => console.log(`Server started on port 3000`));

### client side:
async function addAttachment(folder, files) {
  const url = `http://localhost:3000/files/upload`;

  const headers = {
    host: hostname,
    'contentType': 'multipart/form-data',
  };

  const formData = {
    folder,
    files: files.map((file) => fs.createReadStream(file)),
  };

  return new Promise((resolve, reject) => {
    request.post({
      url,
      headers,
      formData,
    }, (err, httpResponse, body) => {
      if (err) {
        return reject(err);
      }
[...]

PR #1820, opened 2020-03-16
Fix a crash in multipartBodyParser if mapping files array #1819

guoyibj:
## Pre-Submission Checklist

- [*] Opened an issue discussing these changes before opening the PR
- [*] Ran the linter and tests via `make prepush`
- [*] Included comprehensive and convincing tests for changes

## Issues

Closes:

* Issue #1819 

Crash in multipartBodyParser if mapping files array

# Changes

Fix the crash issue and add test code.

### DIFF
--- a/lib/plugins/multipartBodyParser.js
+++ b/lib/plugins/multipartBodyParser.js
@@ -116,6 +116,30 @@ function multipartBodyParser(options) {
                         if (req.params[f] && !override) {
                             return;
                         }
+                        if (Array.isArray(files[f])) {
+                            req.params[f] = [];
+                            files[f].forEach(function forEach2(file) {
+                                barrier.start('fs' + file.name);
+                                fs.readFile(file.path, function readFile(
+                                    ex,
+                                    data
+                                ) {
+                                    barrier.done('fs' + file.name);
+                                    if (ex) {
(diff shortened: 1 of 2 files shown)


lib/plugins/multipartBodyParser.js
```diff
@@ -116,6 +116,30 @@ function multipartBodyParser(options) {
...
+                                            )
+                                        );
+                                    }
+                                    req.params[f].push(data);
+                                    return true;
+                                });
+                            });
+                        }
+
                         barrier.start('fs' + f);
```
retrohacker, line 143: Thanks for the contribution @guoyibj!

I'm not super familiar with this piece of code. But it looks like we are trying to flatten the array here. This if condition doesn't look like it returns and the following barrier is still run against `f`. Is that intentional?

Would it be possible to flatten `files` first? I.E. convert `[f1, f2, [f3, [f4]]]` to `[f1, f2, f3, f4]` to avoid the conditional logic here?

guoyibj:
Test script in my MR is not suitable, I attached the reproducible projects in the ticket https://github.com/restify/node-restify/issues/1819, can you please download testServer.zip and testClient.zip and try them?

PR #1823, opened 2020-03-18
feat(deps): replace cover/istanbul with nyc

mmarchini:
`istanbul` is not supported anymore, and the last version of `cover` was
released seven years ago. `cover` also depends on a vulnerable package
(which makes `npm audit` noisy). Since `make cover` was not working
anymore, replacing it with `nyc` (`istanbul`'s successor) should be
straightforward.

Signed-off-by: Matheus Marchini <mmarchini@netflix.com>

## Pre-Submission Checklist


wesleytodd:
I only ended up here because of the release please PR, so sorry if this is bringing up something unnecessarily, but have you considered using `c8`?  I have started using it on a few projects and it is working great for me.

https://www.npmjs.com/package/c8

mmarchini:
Haven't looked into it yet. What made you change? How hard was the change?

I think nyc is working for us right now, but I'm open to changing if it becomes a problem at some point.

wesleytodd:
It was super simple to change, just swap in `c8` where you call `nyc`. The nice thing about `c8` is that it uses [v8's internal coverage api](https://nodejs.org/dist/latest-v10.x/docs/api/cli.html#cli_node_v8_coverage_dir). Maybe @bcoe could elaborate on other benefits?

PR #1827, opened 2020-04-09
removed useless methods

anshymanel:
those `.filter` methods are useless, they don't change anything, the main logic is within the `.map` method


DonutEspresso:
We do assert on arrayOfString, so it's possible this is a non-issue. If we had some tests we could probably verify whether it actually changes anything.

ghermeto:
True! It might be a non-issue.

@shymanel would you be able to add some tests?

3imed-jaberi (approved):

ghermeto:
account that filed the PR has been deleted

PR #1833, opened 2020-06-27
9.X - async/await

ghermeto:
## Pre-Submission Checklist

# Changes

> What does this PR do?

_This PR has two **commits**:_
1. drops support to node 8.x (EOL);
2. adds support for async/await functions on `pre`, `use` and route handlers;

**_Example of async handling:_**

```js
    const server = restify.createServer();
    
    server.pre(async function (req, res) {
        await something();
        doSomethingElse();
    });

    server.get('/async', async function (req, res) {
        const body = await thatThing();
        res.send(body);
    });
```

Basically, the `Chain` call will check if the handler return is a `Promise` (by checking the existence of `.then`, as Falcor Router does it) and resolves by calling `next` or rejects by calling `next with error`.

.

### DIFF
--- a/docs/guides/8to9guide.md
+++ b/docs/guides/8to9guide.md
@@ -1,0 +1,96 @@
+---
+title: restify 8.x to 9.x migration guide
+permalink: /docs/8to9/
+---
+
+## Introduction
+
+restify `9.x` comes with `async/await` support!
+
+## Breaking Changes
+
+### Drops support for Node.js `8.x`
+
+Restify requires Node.js version `>=10.0.0`. 
+
+### Async/await support
+
+`async/await` basic support for `.pre()`, `.use()` and route handlers.
+
+#### Example
+
+```js
+const restify = require('restify');
+
+const server = restify.createServer({});
+
+server.use(async (req, res) => {
+  req.something = await doSomethingAsync();
+});
+
+server.get('/params', async (req, res) => {
+  const value = await asyncOperation(req.something);
+   res.send(value);
+});
+```
+
+#### Middleware API (`.pre()` and `.use()`)
+
+```js
+server.use(async (req, res) => {
+  req.something = await doSomethingAsync();
+});
+```
+- `fn.length === 2` (arity 2);
+- `fn instanceof AsyncFunction`;
+- if the async function resolves, it calls `next()`;
+- any value returned by the async function will be discarded;
+- if it rejects with an `Error` instance it calls `next(err)`;
+- if it rejects with anything else it wraps in a `AsyncError` and calls `next(err)`;
+
+#### Route handler API
+
+```js
+server.get('/something', async (req, res) => {
+  const someData = await fetchSomeDataAsync();
+  res.send({ data: someData });
+});
+```
+- `fn.length === 2` (arity 2);
+- `fn instanceof AsyncFunction`;
+- if the async function resolves without a value, it calls `next()`;
+- if the async function resolves with a string value, it calls `next(string)` (re-routes*);
+- if the async function resolves with a value other than string, it calls `next(any)`;
+- if it rejects with an `Error` instance it calls `next(err)`;
+- if it rejects with anything else it wraps in a `AsyncError` and calls `next(err)` (error-handing**);
+
+##### (*) Note about re-routing:
+The `8.x` API allows re-routing when calling `next()` with a string value. If the string matches a valid route,
+it will re-route to the given handler. The same is valid for resolving a async function. If the value returned by
+the async function is a string, it will try to re-route to the given handler.
+
+##### (**) Note about error handling:
+Although it is recommended to always reject with an instance of Error, in a async function it is possible to
+throw or reject without returning an `Error` instance or even anything at all. In such cases, the value rejected
+will be wrapped on a `AsyncError`.
+
+### Handler arity check
+Handlers expecting 2 or fewer parameters added to a `.pre()`, `.use()` or route chain must be async functions, as:
+
+```js
+server.use(async (req, res) => {
+  req.something = await doSomethingAsync();
+});
+```
+
+Handlers expecting more than 2 parameters shouldn't be async functions, as:
+
+````js
+// This middleware will be rejected and restify will throw
+server.use(async (req, res, next) => {
+  doSomethingAsync(function callback(val) {
+    req.something = val;
+    next();
+  });
+});
+````
--- a/lib/chain.js
+++ b/lib/chain.js
@@ -2,6 +2,7 @@
 
 var assert = require('assert-plus');
 var once = require('once');
+var customErrorTypes = require('./errorTypes');
 
 module.exports = Chain;
 
@@ -71,6 +72,15 @@ Chain.prototype.getHandlers = function getHandlers() {
  * @returns {undefined} no return value
  */
 Chain.prototype.add = function add(handler) {
+    assert.func(handler);
+    if (handler.length <= 2) {
+        // arity <= 2, must be AsyncFunction
+        assert.equal(handler.constructor.name, 'AsyncFunction');
+    } else {
+        // otherwise shouldn't be AsyncFunction
+        assert.notEqual(handler.constructor.name, 'AsyncFunction');
+    }
+
     // _name is assigned in the server and router
     handler._name = handler._name || handler.name;
 
@@ -144,7 +154,6 @@ Chain.prototype.run = function run(req, res, done) {
  */
 function call(handler, err, req, res, _next) {
     var arity = handler.length;
-    var error = err;
     var hasError = err === false || Boolean(err);
 
     // Meassure handler timings
@@ -157,6 +166,32 @@ function call(handler, err, req, res, _next) {
         _next(nextErr, req, res);
     }
 
+    function resolve(value) {
+        if (value && req.log) {
+            // logs resolved value
+            req.log.debug({ value }, 'Async handler resolved with a value');
+        }
+
+        return next(value);
+    }
+
+    function reject(error) {
+        if (!(error instanceof Error)) {
+            error = new customErrorTypes.AsyncError(
+                {
+                    info: {
+                        cause: error,
+                        handler: handler._name,
+                        method: req.method,
+                        path: req.path ? req.path() : undefined
+                    }
+                },
+                'Async middleware rejected without an error'
+            );
+        }
+        return next(error);
+    }
+
     if (hasError && arity === 4) {
         // error-handling middleware
         handler(err, req, res, next);
@@ -164,12 +199,14 @@ function call(handler, err, req, res, _next) {
     } else if (!hasError && arity < 4) {
         // request-handling middleware
         process.nextTick(function nextTick() {
-            handler(req, res, next);
+            const result = handler(req, res, next);
+            if (result && typeof result.then === 'function') {
+                result.then(resolve, reject);
+            }
         });
         return;
     }
 
     // continue
-    next(error, req, res);
-    return;
+    next(err);
 }
--- a/lib/errorTypes.js
+++ b/lib/errorTypes.js
@@ -5,5 +5,6 @@ var errors = require('restify-errors');
 // This allows Restify to work with restify-errors v6+
 module.exports = {
     RequestCloseError: errors.makeConstructor('RequestCloseError'),
-    RouteMissingError: errors.makeConstructor('RouteMissingError')
+    RouteMissingError: errors.makeConstructor('RouteMissingError'),
+    AsyncError: errors.makeConstructor('AsyncError')
 };
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "report-latency": "./bin/report-latency"
   },
   "engines": {
-    "node": ">=8.3.0"
+    "node": ">=10.21.0"
   },
   "dependencies": {
     "assert-plus": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "report-latency": "./bin/report-latency"
   },
   "engines": {
-    "node": ">=10.21.0"
+    "node": ">=10.0.0"
   },
   "dependencies": {
     "assert-plus": "^1.0.0",
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -127,7 +127,7 @@ if (!process.env.NO_LINT) {
 // stylistic.
 if (!process.env.NO_STYLE) {
     // Global
-    config.rules['max-len'] = [ERROR, { code: 80 }];
+    config.rules['max-len'] = [ERROR, { code: 80, ignoreComments: true }];
 
     // Prettier
     config.extends.push('prettier');
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,7 +9,10 @@ var config = {
     env: {
         browser: false,
         node: true,
-        es6: false
+        es6: true
+    },
+    parserOptions: {
+        ecmaVersion: 2018
     },
     rules: {}
 };
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-   - '8'
    - '10'
    - "lts/*"                # Active LTS release
    - "node"                 # Latest stable release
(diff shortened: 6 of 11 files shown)


ghermeto:
cc/ @josephharrington

ghermeto:
ops... forgot to add examples on the docs (and build the docs)...

ghermeto:
doh! the 3rd commit with async/await docs made this commit huge... if you can go into each individual commit to add your comments, I appreciate  :)

ghermeto:
cc /@cprussin

DonutEspresso:
Nit: For future, would recommend separate, smaller PRs for easier reviewing/accounting. There's a lot going on here (style changes, docs, Node 8.x EOL, etc) and its a bit hard to parse it all at once. 

A few questions that might influence the direction of the PR:
- Do we plan on supporting regular functions that return promises? Depending on the answer may have further implications to discuss, it doesn't look like we make a distinction at the moment.
- Do we plan on supporting mixed mode middleware? e.g., some callback, some async/await. I think the answer is yes, but good to be explicit.  
- I think the rejection contract could benefit from a bit more clarity/discussion. I want to say we should only allow rejecting with Error objects, but we can't enforce that at server startup. What are some of the ecosystem best practices here? 
- Async/await is likely to introduce some race conditions. e.g., imagine a `res.write` call in a loop. It looks like [fastify has some clear rules](https://www.fastify.io/docs/latest/Routes/#async-await) around this to ensure things don't go off the rails. Anything that applies to us or that we can learn from?

ghermeto:
> Nit: For future, would recommend separate, smaller PRs for easier reviewing/accounting. There's a lot going on here (style changes, docs, Node 8.x EOL, etc) and its a bit hard to parse it all at once.

Yes, I broke them into different commits, but you are right. After I added the documentation commit  I realized it should have waited. I won't send multiple commits next time. 

> A few questions that might influence the direction of the PR:
> 
> * Do we plan on supporting regular functions that return promises? Depending on the answer may have further implications to discuss, it doesn't look like we make a distinction at the moment.

I followed the ecosystem. I guess we could check `handler.contructor.name === 'AsyncFunction'`, but it is [not recommended](https://github.com/tc39/ecmascript-asyncawait/issues/78#issuecomment-164425925). For instance, fastify [only checks the constructor name](https://github.com/fastify/fastify/blob/bf18412f2ba838fc06e3162a13ebcf88b5a283be/fastify.js#L398-L408) to verify the arity of the handler when inserted, but later they just [do this same check](https://github.com/fastify/fastify/blob/bf18412f2ba838fc06e3162a13ebcf88b5a283be/lib/hooks.js#L160) for `then`. I wouldn't oppose check for the constructor & arity, but I believe is unnecessary, since the result **_must be treated_** like any Promise.

> * Do we plan on supporting mixed mode middleware? e.g., some callback, some async/await. I think the answer is yes, but good to be explicit.

yes, this allows mixed mode. Like the other frameworks that support both models.

> * I think the rejection contract could benefit from a bit more clarity/discussion. I want to say we should only allow rejecting with Error objects, but we can't enforce that at server startup. What are some of the ecosystem best practices here?

I used the [same approach as fastify](https://github.com/fastify/fastify/blob/bf18412f2ba838fc06e3162a13ebcf88b5a283be/lib/hooks.js#L169-L175). You are right, we can't enforce it and it deviates from the current behavior. I don't think changing the error handling behavior should be coupled with this PR. 

What I can do is check for if  `err instanceof Error` or `err instanceof VError` and, if not, I can force wrap it under an `AsyncHandlerRejection` (or `AsyncHandlerError`), but that will differ from the callback behavior that allows user to throw anything.

> * Async/await is likely to introduce some race conditions. e.g., imagine a `res.write` call in a setInterval loop. It looks like [fastify has some clear rules](https://www.fastify.io/docs/latest/Routes/#async-await) around this to ensure things don't go off the rails. Anything that applies to us or that we can learn from?

The biggest problem with fastify happens because they allow the user to both return the function with the body or use `reply.send`. It provides a nicer DevEx but comes with trade-offs. You can't do this in restify. You still need to call `res.send`, but the `setInterval` is an interesting case.

We should recommend using `async/await` syntax on a `promise` or `async/await` context. If you use any code that relies on callbacks and still want to use async handlers, you should convert these to promises outside the handler.  Or better yet, avoid using async handlers. I guess that is something to append to the documentation. Maybe on a follow-up PR to add more docs around it?

ghermeto:
@DonutEspresso Started issue #1837 to discuss this further.

docs/_api/formatters.md
```diff
@@ -50,7 +50,7 @@ Format a response for being sent over the wire
 
 Type: [Function][9]
 
-**Parameters**
+#### Parameters
```
hekike, line 53: Could you please generate docs as a separate commit outside of this PR?

lib/chain.js
```diff
@@ -157,19 +157,38 @@ function call(handler, err, req, res, _next) {
...
+    }
+
+    function reject(error) {
+        if (!error) {
+            error = new errors.VError(
+                {
+                    name: 'AsyncHandlerRejection',
+                    info: { handler: handler._name }
+                },
+                `Async middleware ${handler._name} rejected without an error`
```
hekike, line 171: Please avoid dynamic input in messages.

lib/chain.js
```diff
@@ -157,19 +157,38 @@ function call(handler, err, req, res, _next) {
...
+            );
+        }
+        next(error);
```
hekike, line 174: nit: add return to be consistent and highlight this is the end

hekike, line 174: I think it's expected with async-await but this will catch syntax errors etc. 
Maybe explain in the 9.x migration guide.

lib/chain.js
```diff
@@ -157,19 +157,38 @@ function call(handler, err, req, res, _next) {
...
+    function resolve() {
+        next();
+    }
+
+    function reject(error) {
+        if (!error) {
+            error = new errors.VError(
+                {
+                    name: 'AsyncHandlerRejection',
+                    info: { handler: handler._name }
```
hekike, line 169: Please add `cause`

package.json
```diff
@@ -90,7 +90,7 @@
     "report-latency": "./bin/report-latency"
   },
   "engines": {
-    "node": ">=8.3.0"
+    "node": ">=10.21.0"
```
hekike, line 93: Please add a migration guide like [6to7](https://github.com/restify/node-restify/blob/master/docs/guides/6to7guide.md) and highlight Node version bump.

hekike, line 93: I think we need to change Travis CI minimum node test version as well

lib/chain.js
```diff
@@ -157,19 +157,38 @@ function call(handler, err, req, res, _next) {
...
 
+    function resolve() {
+        next();
+    }
+
+    function reject(error) {
+        if (!error) {
+            error = new errors.VError(
+                {
+                    name: 'AsyncHandlerRejection',
```
hekike, line 168: Please define new error types in: https://github.com/restify/node-restify/blob/master/lib/errorTypes.js

mmarchini:
@ghermeto if you want some help I can break this PR in three and force-push only one commit here. It might be easier for folks to review.

lib/chain.js
ghermeto: what should add here as a `cause`? This error is created whenever the user rejects an async function without passing it anything:

```js
server.get('/params', async (req, res) => {
  return Promise.reject();
});
```
Do you have suggestions for `cause`?

ghermeto: You know... just got a mistake in my code... I wasn't wrapping synthetic errors because I thought Restify would throw anything, but actually, if the value passed to `next()` is a string, it will try to route it, for instance:

```js
const server = restify.createServer({});

server.use(restify.plugins.queryParser({mapParams:true}));

server.get('/route', (req, res, next) => {
   next('getresult');
});

server.get('/string', (req, res, next) => {
   next('invalid');
});

server.get('/number', (req, res, next) => {
   next(1);
});

server.get('/error', (req, res, next) => {
   next(new Error(''));
});

server.get('/result', (req, res, next) => {
   res.send(req.query);
   next();
});

server.listen(8081, 'localhost', function () {
   console.info('server started');
});

```

The result of `curl http://localhost:8081/error?id=1` is:
```json
{"code":"Internal","message":"Error"}
```
The result of `curl http://localhost:8081/number?id=1` is:
```json
{"code":"Internal","message":"1"}
```
The result of `curl http://localhost:8081/string?id=1` is:
```json
{"code":"Internal","message":"RouteMissingError: Route by name doesn't exist"}
```
The result of `curl http://localhost:8081/route?id=1` is:
```json
{"id":"1"}
```

I will update the code to always wrap rejections and update #1837 .

lib/chain.js
ghermeto: Should the migration guide be part of a separate PR, with the docs?

package.json
ghermeto: Should the migration guide be a separate PR?

lib/chain.js
cprussin: FWIW trying to route on strings is IMHO an awful barnacle of Restify, it's worth considering deprecating it--it's essentially a GOTO with all the pain that comes along.

ghermeto: wrapping all non-errors in `AsyncHandlerError` actually turned out pretty nice, because if nothing is specified, the cause is `undefined`.

ghermeto: and I just discovered `cause` must be an instance of `Error`, so I'm moving the `cause` to inside the `info`.

package.json
hekike: yes, pleace

ghermeto:
thank you @mmarchini. I force pushed removing the generated docs and added some changes Peter requested. 
I kept the `jsdoc` updates as it made sense to be on the same PR, now I'm missing the migration guide.

DonutEspresso:
> I followed the ecosystem. I guess we could check `handler.contructor.name === 'AsyncFunction'`, but it is [not recommended](https://github.com/tc39/ecmascript-asyncawait/issues/78#issuecomment-164425925). 

Yeah, that makes sense. My concern is that this opens up a new footgun with promise usage. The following handler will just hang forever:

```
function(req, res) {
  return FuncThatWasSupposedToReturnPromiseButDidnt();
}
```

I haven't used promises enough to know what's the best way to handle this. Is it common to do runtime assertions on thenables? I'd much rather the server blew up than fail silently. One way to enforce this is by saying async functions should only have an arity of 2, and we don't pass next to those functions. 

> > * I think the rejection contract could benefit from a bit more clarity/discussion. I want to say we should only allow rejecting with Error objects, but we can't enforce that at server startup. What are some of the ecosystem best practices here?
> 
> I used the [same approach as fastify](https://github.com/fastify/fastify/blob/bf18412f2ba838fc06e3162a13ebcf88b5a283be/lib/hooks.js#L169-L175). You are right, we can't enforce it and it deviates from the current behavior. I don't think changing the error handling behavior should be coupled with this PR. Right now, an `async` error will be treated exactly like a callback error.
> 
> What I can do is check for if `err instanceof Error` or `err instanceof VError` and, if not, I can force wrap it under an `AsyncHandlerRejection` (or `AsyncHandlerError`), but that will differ from the callback behavior that allows user to throw anything.

Sorry! I wasn't very clear, I was suggesting that we formalize the next() contract, and the promise pathways that lead up to that. Essentially, there's two ways to cede control back to restify through next today, and this PR adds a variation for each of those contracts: 

* next()
  * With async functions, a promise that resolves no value
* next('fooRoute') 
  * With async functions, a promise that resolves 'fooRoute'
* next(new Error('boom'))
  * With async functions, a promise that rejection with an Error object

I think there's a good argument to be made that anything that is not one of those three should cause a runtime exception. But I feel less strongly than I do about at least formalizing and documenting it.

DonutEspresso:
> What I can do is check for if `err instanceof Error` or `err instanceof VError` and, if not, I can force wrap it under an `AsyncHandlerRejection` (or `AsyncHandlerError`), but that will differ from the callback behavior that allows user to throw anything.

What do you mean by `callback behavior that allows user to throw anything`?

DonutEspresso:
Oops, sorry, maybe I should have left my comments in #1837!

ghermeto:
I will reply to your comments on #1837

docs/guides/8to9guide.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Introduction
+
+restify `9.x` comes with `async/await` support!
+
+## Breaking Changes
+
+### Drops support for Node.js `8.x`
+
+Restify requires Node.js version `>=10.21.0`.
```
daviande, line 14: Is there anything in particular for 10.21.0 instead of just 10.x?

ghermeto: I guess not... I will update it

docs/guides/8to9guide.md
```diff
@@ -0,0 +1,94 @@
...
+
+```js
+server.use(async (req, res) => {
+  req.something = await doSomethingAsync();
+});
+```
+- `fn.length === 2` (arity 2);
+- `fn instanceof AsyncFunction`;
+- if the async function resolves, it calls `next()`;
+- any value resolved will be discarded;
```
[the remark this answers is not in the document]
ghermeto: do you think it would read better "any value returned by the async function?". I will make the change.

docs/guides/8to9guide.md
```diff
@@ -0,0 +1,94 @@
...
+- if it rejects with an `Error` instance it calls `next(err)`;
+- if it rejects with anything else it wraps in a `AsyncHandlerError` and calls `next(err)`;
+
+#### Route handler API
+
+```js
+server.use(async (req, res) => {
+  req.something = await doSomethingAsync();
+});
+```
```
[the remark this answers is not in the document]
ghermeto: OMG...

docs/guides/8to9guide.md
```diff
@@ -0,0 +1,94 @@
...
+- if the async function resolves without a value, it calls `next()`;
+- if the async function resolves with a string value, it calls `next(string)` (re-routes*);
+- if the async function resolves with a value other than string, it calls `next(any)`;
+- if it rejects with an `Error` instance it calls `next(err)`;
+- if it rejects with anything else it wraps in a `AsyncHandlerError` and calls `next(err)` (error-handing**);
+
+##### (*) Note about re-routing:
+The `8.x` API allows re-routing when calling `next()` with a string value. If the string matches a valid route,
+it will re-route to that handler. The same is valid for resolving a async function. Whatever value it returns,
+will be sent to `next()` and the behavior will be exactly the same.
```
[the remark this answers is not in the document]
ghermeto: rephrased

lib/errorTypes.js
```diff
@@ -5,5 +5,6 @@ var errors = require('restify-errors');
 // This allows Restify to work with restify-errors v6+
 module.exports = {
     RequestCloseError: errors.makeConstructor('RequestCloseError'),
-    RouteMissingError: errors.makeConstructor('RouteMissingError')
+    RouteMissingError: errors.makeConstructor('RouteMissingError'),
+    AsyncHandlerError: errors.makeConstructor('AsyncHandlerError')
```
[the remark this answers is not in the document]
ghermeto: renamed

docs/guides/8to9guide.md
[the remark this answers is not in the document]
ghermeto: fixed

docs/guides/8to9guide.md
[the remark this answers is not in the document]
ghermeto: done

docs/guides/8to9guide.md
ghermeto: done

ghermeto:
I spoke with @hekike and he said I can merge this. 

Just want to make sure @mmarchini (who had comments) and others can take a look at it before I do.

cc/ @m5m1th @josephharrington

ghermeto:
ok, since no new objection was raised, and all comments were addressed, I'm merging this code. We can always submit fixes before we make a new release.

nikeee:
What happens if the handler calls `next()` besides being async? For example:

```js
server.get('/async', async function (req, res) {
    const body = await thatThing();
    res.send(body);
    next();
});
```

Is this valid?

Issue #1845, opened 2020-07-10
Deprecate re-routing through string parameter to `next`


ghermeto:
## Today

The current `next` API for route handlers is the following:

**next()**
- goes to the next handler in the chain (if any)

**next('fooRoute')**
- [re-routes](https://github.com/restify/node-restify/blob/12be9e243a407eaf7a30cbb16e399ee2a46dec93/lib/server.js#L1245-L1264) to the route chain that resolves for  'fooRoute'

**next(1) (or next({}))**
- sends an [InternalServerError](https://github.com/restify/node-restify/blob/12be9e243a407eaf7a30cbb16e399ee2a46dec93/lib/server.js#L1510-L1515)

**next(new Error('boom'))**
- [returns the error](https://github.com/restify/node-restify/blob/12be9e243a407eaf7a30cbb16e399ee2a46dec93/lib/server.js#L1504-L1515) (if it has the `statusCode` property), otherwise converts it into a `String` and wraps in an `InternalServerError` (same as passing it a number as the first parameter)

## Proposal

Deprecate the API `next('fooRoute')` and its async/await equivalent. Instead, use an explicit re-routing mechanism similar to what was [proposed](https://github.com/restify/node-restify/issues/1837#issuecomment-654528925) by @mmarchini:

```js
  server.get('/one', async (req, res, next) => {
    req.routeTo('getTwo');
    next();
  });

  server.get('/two', async (req, res, next) => {
    res.send('two');
    next();
  });
  
```

The method `.routeTo()` will set a re-route flag, that will be read after the handler finished and it would re-route at that point without relying on the error chain.

Since setting the flag is synchronous, this approach also works for async functions:

```js
  server.get('/one', async (req, res) => {
    req.routeTo('getTwo');
  });

  server.get('/two', async (req, res) => {
    res.send('two');
  });

```

This proposal keeps the functionality intact while removing a potential problem for the async/await DevEx.

Thoughts?

cc/ @cprussin

cprussin:
I don't see any good reason to support re-route. `routeTo` is just `goto` in disguise.

Other than that :+1:

ghermeto:
actually this seems to be a non-documented feature. I can't find documentation of the re-routing in the official docs...

ghermeto:
cc/ @restify/current-core

mmarchini:
IMO we can start by deprecating re-routing, and later if we want to explore alternatives we have that option too.

[...]

PR #1847, opened 2020-07-11
feat: remove re-routing from handler

ghermeto:
**BREAKING CHANGE**: deprecates and removes re-routing when passing a `string` parameter to `next()`

## Pre-Submission Checklist

## Issues

* Issue #1845

# Changes

Assumes the standard node.js callback API: `cb(err, value)` and treats `next()` as a callback. This way:

1. deprecates and removes re-routing;
2. `next('foo')` will be considered an error;
3. the returned value from an async function is ignored (as it would for `next(null, 'foo')`);

### DIFF
--- a/lib/chain.js
+++ b/lib/chain.js
@@ -169,10 +169,13 @@ function call(handler, err, req, res, _next) {
     function resolve(value) {
         if (value && req.log) {
             // logs resolved value
-            req.log.debug({ value }, 'Async handler resolved with a value');
+            req.log.warn(
+                { value },
+                'Discarded returned value from async handler'
+            );
         }
 
-        return next(value);
+        return next();
     }
 
     function reject(error) {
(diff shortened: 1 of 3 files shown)


mmarchini:
> next('foo') will be considered an error;

The standard Node.js callback API accepts any value type for the first argument. Should we do the same?

ghermeto:
> > next('foo') will be considered an error;
> 
> The standard Node.js callback API accepts any value type for the first argument. Should we do the same?

Yes, actually right now it does return an `InternalServerError` if you give call it `next(1)` or `next({})`. With strings, it was checking for that on the error handling and re-routing.

pinko-fowle:
There is [response.redirect](http://restify.com/docs/response-api/#redirect) for those migrating off this.

pinko-fowle:
There's no replacement api is there? A `request.forward` or some such?

We're wondering what our migration forward will look like.

Issue #1851, opened 2020-07-22
Plugin to make res.status() chainable.


june07:
- [x] Used appropriate template for the issue type
- [x] Searched both open and closed issues for duplicates of this issue
- [x] Title adequately and _concisely_ reflects the feature or the bug

# Feature Request

## Use Case
Working on a multi service system where Express.js is used for some services and Restify for others.  The middleware is written such that it expects res.status to be chainable, so when passing in Restify objects, the middleware errors:

`TypeError: res.status(...).send is not a function`

I've written a plugin which provides this:
https://github.com/june07/node-restify/commit/0ec0d88347d6179bf9a5b8385ae0f3fad939ae4d

Following the proper flow to see if I can open a PR for this.

ghermeto:
I don't oppose this change because we already return `this` on many of the `Response` methods, so it would be consistent. However,  this is definitely a breaking change and it would have to go out on 9.x.

For now, I believe the following would work for both Express and Restify:

```js
res.status(...);
res.send(...);
```

@restify/current-core any other thoughts?

june07:
@ghermeto Thank you.  It's been almost 3 weeks without any further input from @restify/current-core so I've submitted a PR.  Also you are absolutely right with the suggestion below, if changing the code was feasible.  Otherwise when the code is upstream, or owned by someone else, this PR should help.

> ```js
> res.status(...);
> res.send(...);
> ```

PR #1854, opened 2020-08-11
Added chainable status plugin.

june07:
Plugin to make res.status chainable.

## Pre-Submission Checklist

## Issues

Closes:

* Issue #1851 

# Changes

This PR provides plugin code including tests.

### DIFF
--- a/lib/plugins/chainableStatus.js
+++ b/lib/plugins/chainableStatus.js
@@ -1,0 +1,62 @@
+
+
+/**
+ * MIT License
+ *
+ * Copyright (c) 2020 June07
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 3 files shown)


lib/plugins/chainableStatus.js
```diff
@@ -0,0 +1,61 @@
+/**
```
retrohacker, line 1: This software is packaged and distributed as a whole. I'm not sure we want to go down the path of separate license claims for individual contributions.

I am not a lawyer though, I've just never seen a license preamble in a source file that didn't match the projects. Not sure what the implications are.

retrohacker:
The code looks good to me. Before we merge, I'd like for other maintainers to review the license comment. Once that happens, there is no need to wait for a follow up from me. Go ahead and merge!

lib/plugins/chainableStatus.js
ghermeto: I think this can be resolved by replacing the license by a link to:
https://github.com/restify/node-restify/blob/master/LICENSE

ghermeto:
@wesleytodd, @mmarchini instead of a plugin, should we just make this change in the core for 9.x?

wesleytodd:
I think it is a more fluid and user friendly api, and one which I will be pushing for as a part of the new core high level http interface.  If we go that route in core (which I cannot promise at this time), then this change would align well.

june07:
In the meantime are we going to merge this?

retrohacker:
@june07 can you update the copyright claim per feedback? I think that is blocking the merge.

cc:// @ghermeto @wesleytodd is your 9.x suggestion a blocker for this?

ghermeto:
> cc:// @ghermeto @wesleytodd is your 9.x suggestion a blocker for this?

@june07 can you instead of a plugin make this change directly to core? We plan on releasing 9.x next month, so it would be a very short-lived plugin.

Issue #1604, opened 2018-02-11
Restify.plugins.serverStatic option "appendRequestPath" is bugged


keithlee96:
- [ X ] Used appropriate template for the issue type
- [ X ] Searched both open and closed issues for duplicates of this issue
- [ X ] Title adequately and _concisely_ reflects the feature or the bug

# Bug Report

Restify.plugins.serverStatic option "appendRequestPath" is not working properly.

## Restify Version
v6.3.4

## Node.js Version
v8.91

## Expected behaviour
I expected appendRequestPath = false; to make it so that the request path is not appended to the file path. (As stated in the documentation at http://restify.com/docs/plugins-api/#servestatic)

## Actual behaviour
The request path was appended to the file path, despite the fact that appendRequestPath was set to false.

## Repro case

index.js
```javascript
var restify = require('restify');

const server = restify.createServer({
    name: 'myapp',
    version: '1.0.0'
  });

server.get('/endpoint', restify.plugins.serveStatic({
    directory: __dirname, // dirname is /Users/keithlee96/testFolder/test-api
    default: 'index.html',
    appendRequestPath: false
}));

server.listen(9090, function(){
    console.log('%s listening at %s', server.name, server.url);
});
```
index.html was in the same folder.

I ran the file with node index.js
When I made a get request to localhost:9090/endpoint in postman, I got the error:
```json
{
    "code": "ResourceNotFound",
    "message": "/endpoint; caused by Error: ENOENT: no such file or directory, stat '/Users/keithlee96/testFolder/test-api/endpoint'"
}
```

Expected the targeted file path to be: '/Users/keithlee96/testFolder/test-api/index.html'

## Are you willing and able to fix this?

No, I'm no familiar with typescript, so I don't know how to fix the error myself. I just want it to behave as the documentation http://restify.com/docs/plugins-api/#servestatic says it should. Any help on this would be greatly appreciated.

PavelPolyakov:
I agree, just met the same unexpected behaviour.
I assume it's because of [`dirBasename === reqpathBasename`](https://github.com/restify/node-restify/blob/master/lib/plugins/static.js#L177) condition, which forces you to name the route folder name as it is in the directory which you try to serve.

Why it is there?

I assume it has little sense.

Regards,

retrohacker:
Hey @keithlee96,

First, thank you for the thorough bug report with detailed examples.

[...]

PR #1860, opened 2020-10-26
fix appendRequestPath = false scenarios

jmbeltran07:
## Pre-Submission Checklist

## Issues

Closes: 

* Issue # https://github.com/restify/node-restify/issues/1604

Restify.plugins.serverStatic option "appendRequestPath" is not working properly.
According to documentation, when appendRequestPath = false, the file we are requesting should be served directly from `directory`, this wasn't the case.

# Changes

With this PR when the user is setting appendRequestPath = false, the server will serve the requested resource specified in the endpoint (or the default static resource) which is located directly on `directory`.
Now the behavior adequates with what's described on http://restify.com/docs/plugins-api/#servestatic

Also we are modifying the function `testNoAppendPath` used by the tests which was passing as false positive so it catches this errors. And we are adding more tests to increase the coverage of Restify.plugins.serverStatic function.

### DIFF
--- a/lib/plugins/static.js
+++ b/lib/plugins/static.js
@@ -87,6 +87,7 @@ function serveStatic(options) {
     assert.optionalObject(opts.match, 'options.match');
     assert.optionalString(opts.charSet, 'options.charSet');
     assert.optionalString(opts.file, 'options.file');
+    assert.optionalString(opts.file, 'options.default');
     assert.bool(opts.appendRequestPath, 'options.appendRequestPath');
 
     var p = path.normalize(opts.directory).replace(/\\/g, '/');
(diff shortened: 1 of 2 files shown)


lib/plugins/static.js
```diff
@@ -170,19 +171,19 @@ function serveStatic(options) {
...
+            if (path.extname(req.path()) === '') {
                 file = opts.directory;
             } else {
-                file = path.join(
-                    opts.directory,
-                    decodeURIComponent(path.basename(req.path()))
-                );
+                var fileName = decodeURIComponent(path.basename(req.path()));
+                // test for requested file, if it doesn't exist we send the directory as path since appendRequestPath = false
+                if (!fs.existsSync(path.join(opts.directory, fileName))) {
```
retrohacker, line 179: nit: mind breaking these nested function invocations up into multiple lines w/ variables indicating what each path is supposed to represent?

retrohacker (approved):
Thanks for this thoughtful contribution :raised_hands:

lib/plugins/static.js
ghermeto: +1

jmbeltran07: Sure! let me work on it and update the PR!

rajatkumar:
Just for completeness, we also happen to have another flavor of serving static files with restify which has it's API syntax (and functionality) similar to that of expressjs. 

We had seen a lot of issues with `restify.plugins.serveStatic` and hence at some point we built `restify.plugins.serveStaticFiles`. 

Here is a small sample: https://codesandbox.io/s/gallant-raman-9x9jk?file=/app.js

PR #1878, opened 2021-07-08
chore: remove travis and update github ci

ghermeto:
removing Travis in favor of Github ci.

### Why?

IMO Github CI:
- runs faster than Travis (Travis sometimes queues for a while)
- has a much better UI  (checks tab within the PR, independent checks in the conversation)
- Travis recently migrated out of travis-ci.org and Github CI is already integrated for a while. 

Results with Github CI are consistent, so there is no good reason to have multiple CIs.


3imed-jaberi (approved):
I think it's okay for now to failed 3 test with Node.js@16.x.x

soffieprincess02 (approved):

josephharrington (approved):

PR #1890, opened 2022-01-18
Bump formidable to v2

nikeee:
According to the [version notes](https://github.com/node-formidable/formidable/blob/master/VERSION_NOTES.md), it should be compatible with v1.


acommodari:
Will this be merged in for a future major version? or even `formidable v3`? would love to help bump up some dependencies

PR #1892, opened 2022-01-31
fix: use close event on response instead of socket

### LINKED ISSUE #1883: MaxListenerExceededWarning with restify 8.6.0
When I updated to latest version of Restify. I started to get memory leak warnings.

**Restify Version**: 8.6.0
**Node.js Version**: 14.18.1

## Expected behaviour
No warnings

## Actual behaviour
```bash
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close listeners added to [Socket]. Use emitter.setMaxListeners() to increase limit
    at _addListener (events.js:475:17)
    at Socket.addListener (events.js:497:10)
    at Socket.Readable.on (internal/streams/readable.js:853:35)
    at Socket.socketListenerWrap [as on] (_http_server.js:950:54)
    at Socket.once (events.js:541:8)
    at readBody (/code/node_modules/restify/lib/plugins/bodyReader.js:189:20)
    at nextTick (/code/node_modules/restify/lib/chain.js:167:13)
    at processTicksAndRejections (internal/process/task_queues.js:77:11)
```

## Repro case
I don't know how to minimally reproduce it. I am running with a fairly large server in prod.

## Cause
I suspect that this is the cause: https://github.com/restify/node-restify/pull/1880

No, I am not able to fix it beyond just reverting the change in the pull request referenced above.

josephharrington:
In #1880, we switched from using the close event on `req` to close on
`req.socket`. This addressed the intended issue but can trigger frequent
warnings when keep-alive is used due to a listener being added for each
request on the same socket.

By using the close event on `res` instead, we address both the issue of
event ordering in Node.js >= 16 that the original change was targeting
and the event emitter warning leak.

Fixes #1883.

### DIFF
--- a/lib/plugins/bodyReader.js
+++ b/lib/plugins/bodyReader.js
@@ -186,7 +186,7 @@ function bodyReader(options) {
         // add 'close and 'aborted' event handlers so that requests (and their
         // corresponding memory) don't leak if client stops sending data half
         // way through a POST request
-        req.socket.once('close', next);
+        res.once('close', next);
         req.once('aborted', next);
         req.resume();
     }
--- a/test/plugins/bodyReader.test.js
+++ b/test/plugins/bodyReader.test.js
@@ -18,25 +18,25 @@ var CLIENT;
 var PORT;
 
 describe('body reader', function() {
+    beforeEach(function(done) {
+        SERVER = restify.createServer({
+            dtrace: helper.dtrace,
+            log: helper.getLog('server')
+        });
+
+        SERVER.listen(0, '127.0.0.1', function() {
+            PORT = SERVER.address().port;
+
+            done();
+        });
+    });
+
+    afterEach(function(done) {
+        CLIENT.close();
+        SERVER.close(done);
+    });
+
     describe('gzip content encoding', function() {
-        beforeEach(function(done) {
-            SERVER = restify.createServer({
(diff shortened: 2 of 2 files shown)


wesleytodd:
Am I missing something here?  IIUC a client stopping in the middle of a `POST` (as the comment mentions), it could be that the `res` isn't closed, thus this might re-introduce the leak?

josephharrington:
> Am I missing something here? IIUC a client stopping in the middle of a `POST` (as the comment mentions), it could be that the `res` isn't closed, thus this might re-introduce the leak?

I believe `res` is closed whenever `req.socket` is closed. The reason it's preferable to `req`'s close event is that it is not affected by the [autodestroy behavior change](https://github.com/nodejs/node/pull/33035#issuecomment-751503021) in Node.js 16.

I validated that the `res` close event reliably fires by manual testing of malformed and hanging requests, but if there's a specific flow you have in mind that would leave `res` open I am happy to try it (and add a test for it).

> EDIT: I wonder if keeping both the socket and res close events would work?

Then we'd still have the event emitter leak warning, yes?

mmarchini (approved):
this should be fine and unless someone can come up with a scenario where this would leak memory/resources we should merge it (since it's better than the current state of things)

PR #1895, opened 2022-02-17
fix: don't create empty clientError listener for http.Server

nigelis:
## Pre-Submission Checklist

- [x] Opened an issue discussing these changes before opening the PR #1894
- [x] Ran the linter and tests via `make prepush`
- [ ] Included comprehensive and convincing tests for changes

## Issues

Closes:

* Issue #1894

# Changes

> What does this PR do?

- NOT create empty `clientError` listener for `http.Server`.

PR #1897, opened 2022-03-23
fix: Re-add support for clientError listeners

josephharrington:
In the recent fix #1895, the `clientError` event was removed from the
list of events proxied from the underlying server object to the restify
Server. That fix was needed because merely adding a listener to the
underlying server for the `clientError` event prevents the default
behavior of returning an appropriate error code and closing the socket.
So by setting up a listener to proxy the event we were preventing that
default behavior, leading to hanging/leaked requests.

But! We should still support proxying `clientError` events if restify
users have added a listener for it. This enables logging or other custom
behavior. Restify users adding a `clientError` listener take on the
responsibility of sending an appropriate error and closing the socket,
exactly as noted in the Node.js docs for the `clientError` event.
https://nodejs.org/docs/latest-v16.x/api/http.html#event-clienterror

This change ensures we add a listener for `clientError` to the
underlying server object if and only if a listener for `clientError` has
been added to the restify server.

### DIFF
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -33,6 +33,8 @@ var CLIENT;
 var FAST_CLIENT;
 var SERVER;
 
+var NODE_MAJOR_VERSION = process.versions.node.split('.')[0];
+
 if (SKIP_IP_V6) {
     console.warn('IPv6 tests are skipped: No IPv6 network is available');
 }
@@ -2846,7 +2848,7 @@ test('Server returns 400 on invalid method', function(t) {
     }).end();
 });
 
-test('Server returns 431 when header size is too large', function(t) {
+test('Server returns 4xx when header size is too large', function(t) {
     SERVER.get('/jellybeans', function echoId(req, res, next) {
         res.send();
         next();
@@ -2863,8 +2865,13 @@ test('Server returns 431 when header size is too large', function(t) {
         }
     };
     http.request(opts, function(res) {
-        t.equal(res.statusCode, 431);
-        t.equal(res.statusMessage, 'Request Header Fields Too Large');
+        if (NODE_MAJOR_VERSION > '10') {
+            t.equal(res.statusCode, 431);
+            t.equal(res.statusMessage, 'Request Header Fields Too Large');
+        } else {
+            t.equal(res.statusCode, 400);
+            t.equal(res.statusMessage, 'Bad Request');
+        }
         res.on('data', function() {
             t.fail('Data was sent on 431 error');
         });
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -2820,3 +2820,128 @@ test('async handler should discard value', function(t) {
         t.end();
     });
 });
+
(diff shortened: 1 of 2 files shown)


test/server.test.js
```diff
@@ -2820,3 +2820,128 @@ test('async handler should discard value', function(t) {
...
+        t.equal(res.statusCode, 418);
+        t.equal(res.statusMessage, "I'm a teapot");
+        res.on('data', function() {});
+        res.on('end', function() {
+            t.end();
+        });
+    }).end();
+});
+
+test('Server correctly handles multiple clientError listeners', function(t) {
```
wesleytodd, line 2911: Not sure about this. So it is an event emitter, so obviously two handlers is technically valid, but the use case here is almost certainly to be sending a response in some cases, which shifts the burden onto each handler to make sure it properly checks before sending the response and closing the socket.  If a user does this in multiple handlers they could get some pretty unexpected error results if they are not careful.

Could a better user experience from restify be to error or warn when they try to add a second handler?

wesleytodd (approved):

test/server.test.js
josephharrington: @wesleytodd, I agree generally. This is more of a contrived test rather than a valid usage example. I added this test to make sure that the proxying is done correctly / doesn't emit the proxied event multiple times if multiple listeners are added.

My goal with this change was to restore the ability to add a `clientError` listener after #1895 removed it, so am matching the original behavior here. I don't think folks would do this very often if at all. One actual use case might be to do the response in one handler and log in another. I'm not opposed to adding a warning but feels beyond the scope of this change.

wesleytodd: cool, yeah makes sense. I just wanted to bring up this as I have thought before that sometimes these kind of api's from node core have foot guns, and I think frameworks are generally good places to shore up those gaps.  I for sure don't think this one is a common issue worth spending much time on. So 👍 to the change!

DonutEspresso:
Naive Q - would it make sense for restify to replicate the default behavior when you attach a listener? Or is this a pretty standard Node.js expectation? It just seems strange that adding a listener would actually _change_ behavior.

josephharrington:
> Naive Q - would it make sense for restify to replicate the default behavior when you attach a listener? Or is this a pretty standard Node.js expectation? It just seems strange that adding a listener would actually _change_ behavior.

@DonutEspresso [Here is the default handler logic](https://github.com/nodejs/node/blob/3f3fa6d990b69080d88cda5eae44dd8f59ca9583/lib/_http_server.js#L658-L695) that is executed when no listeners are registered for `clientError`. I agree it feels strange that adding a listener changes behavior. In my opinion the default behavior should be disabled via an explicit option instead. But anyway. 😅 

We _could_ copy that default behavior into restify, but that feels like coupling more tightly with the underlying implementation than just conditionally proxying the event. If that default implementation in Node.js ever changed, we'd likely need to change it in restify as well anyway.

Issue #1888, opened 2021-11-17
restify cannot be imported with latest node v18.0.0 nightly: `Request.prototype.closed` patch clashes with new Readable.closed getter


trentm:
- [x] Used appropriate template for the issue type
- [x] Searched both open and closed issues for duplicates of this issue
- [x] Title adequately and _concisely_ reflects the feature or the bug

**Restify Version**: 8.6.0
**Node.js Version**: v18.0.0-nightly2021111694fa781580 (from https://nodejs.org/download/nightly/v18.0.0-nightly2021111694fa781580/)

## Expected behaviour

`require('restify')` does not throw

## Actual behaviour

```
% node
Welcome to Node.js v18.0.0-nightly2021111694fa781580.
Type ".help" for more information.
> require('restify')
Uncaught TypeError: Cannot set property closed of #<Readable> which has only a getter
    at patch (.../node_modules/restify/lib/request.js:848:30)
    at Object.<anonymous> (.../node_modules/restify/lib/server.js:33:1)
    at Module._compile (node:internal/modules/cjs/loader:1097:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1149:10)
    at Module.load (node:internal/modules/cjs/loader:975:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:999:19)
    at require (node:internal/modules/cjs/helpers:102:18)
>
```

## Repro case

```js
require('restify')
```

## Cause

Node's master branch recently added a `closed` property to `streams.Readable.prototype` which only has a getter:
https://github.com/nodejs/node/pull/40696/files#diff-040c1f5a53844e600d40b33c4624f1fe39fcf2f8d62c76ca3fc5ea5442231469

This is incompatible with Restify's patching of `http.IncomingMessage` (which is a Readable) to add its own `closed` property: https://github.com/restify/node-restify/blob/71c7f4965342c13cac55847f87149cc34c1ad566/lib/request.js#L839-L851

## Are you willing and able to fix this?

Probably not. I do not use restify currently. I came across this in testing we do on [our APM agent's instrumentation of restify](https://github.com/elastic/apm-agent-nodejs/blob/master/lib/instrumentation/modules/restify.js). We periodically test against the latest node nightly build.

ronilan:
Bump.
Node 18 is now (April 22) current.

acommodari:
[...]

PR #1906, opened 2022-05-08
fix: support node v18

ghermeto:
* fix Request.prototype.closed() clash with new Reader.closed (fixes GH-1888)
* fix IPv6 url by supporting new server.address().family format
* add node 18 to GH Actions

## What?

Backward compatible change to support Node v18.

## Why?

Node v18 breaking changes:

- Added a `Reader.closed` getter that clashes with the `Request.prototype.closed()` patch.
- Changed the data type of `server.address().family` from `string` (`IPv6`) to a `number` (`6`)

## Drawbacks

I don't particularly like this change because it will be unexpected to people using Node18 that `closed` will be a method and not a getter. I would prefer to limit Restify 8.x to Node 17 and make a breaking change for 9.x by changing the method name to `isClosed()` or better yet, drop the method in favor of patching a getter for node v10~v17.

@restify/current-core please advise on the preferred action plan.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -837,6 +837,16 @@ function patch(Request) {
     };
 
     /**
+     * Node v18 introduced a `closed` getter to Readable stream which breaks the API
+     * @see https://github.com/restify/node-restify/issues/1888
+     */
+    Object.defineProperty(Request.prototype, 'closed', {
+        get: function getter(x) {
+            return closed;
+        }
+    });
+
+    /**
      * Returns true when connection state is "close"
      *
      * @private
@@ -845,10 +855,14 @@ function patch(Request) {
      * @function closed
      * @returns {Boolean} is closed
      */
-    Request.prototype.closed = function closed() {
+    function closed() {
         var self = this;
-        return self.connectionState() === 'close';
-    };
+        if (this._readableState && 'close' in this._readableState) {
+            return this._readableState.closed;
+        } else {
+            return self.connectionState() === 'close';
+        }
+    }
 
     /**
      * Returns the route object to which the current request was matched to.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,7 @@ jobs:
           - 12.x
           - 14.x
           - 16.x
+          - 18.x
     runs-on: ${{matrix.os}}
     steps:
       - uses: actions/checkout@v2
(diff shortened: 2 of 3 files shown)


ghermeto:
@DonutEspresso thoughts?

petershaw:
Is there is a roadmap for reviewing and merging node18 support? How can I help?

ghermeto:
cc/ @mmarchini @wesleytodd

Can you advise what is the preferred path to add support to Node 18?

ben:
FWIW here's another ping. We'd love to be on node 18, but can't start up without this patch.

lib/request.js
```diff
@@ -836,6 +836,16 @@ function patch(Request) {
         return self._connectionState;
     };
 
+    /**
+     * Node v18 introduced a `closed` getter to Readable stream which breaks the API
+     * @see https://github.com/restify/node-restify/issues/1888
+     */
+    Object.defineProperty(Request.prototype, 'closed', {
+        get: function getter(x) {
```
meteorlxy, line 844: Is the `x` arg redundant?

sternam: If this code continues to remain necessary (see my overall review comment) @ghermeto Please remove the `x`. @meteorlxy is correct that it is unnecessary.

sternam (changes requested):
One minor comment if we stay with this approach, but at a higher level, why isn't it preferable to address the problem by refactoring restify to just use the `closed` getter directly instead of relying on Request having a `closed` property?  Then this change would be unnecessary.

lib/request.js
```diff
@@ -845,10 +855,14 @@ function patch(Request) {
      * @function closed
      * @returns {Boolean} is closed
      */
-    Request.prototype.closed = function closed() {
+    function closed() {
         var self = this;
-        return self.connectionState() === 'close';
-    };
+        if (this._readableState && 'close' in this._readableState) {
```
mmarchini, line 860: typo

```suggestion
        if (this._readableState && 'closed' in this._readableState) {
```

and tests are failing with typo fixed. I'm trying to figure out if there's a way to do this without monkey patching.

mmarchini:
One alternative we're considering is to just remove the `closed` monkey patch and let restify codebase as well as users rely on `connectionState` instead: https://github.com/restify/node-restify/pull/1929. We might want to add a deprecation warning first though

mmarchini:
Closing since we're moving forward with #1929

PR #1936, opened 2022-12-13
CHANGELOG: put latest version on top

mtmail:
v10 description was listed below v9, usually the latest releases are on top


mtmail:
replaced by https://github.com/restify/node-restify/pull/1955

Issue #1935, opened 2022-12-08
New async/await handler support breaks `next(false)` functionality in current async handlers


gmahomarf:
- [x] Used appropriate template for the issue type
- [x] Searched both open and closed issues for duplicates of this issue
- [x] Title adequately and _concisely_ reflects the feature or the bug

**Restify Version**: 10.0.0
**Node.js Version**: 16.8.1

## Expected behaviour
Given a handler that does async work, I should be able to call `next(false);` and have the chain stop processing there.

## Actual behaviour
The handler arity checks prevent me from having handlers that use next and are async

## Repro case
Code similar to this is used in one of our projects using restify v8. It breaks when trying to update to v10:
```.js
server.use(async (req, res, next) => {
  const result = await someAsyncWork();
  if (shouldStop(result)) {
    res.send({something: 'here'});
    next(false);
    return;
  }

  // ... more work
  next();
});
```

I am aware I could make my handler synchronous, then do `someAsyncWork().then(result => {...})` but `async/await` syntax was chosen for cleanliness and readability.

## Cause
https://github.com/restify/node-restify/blob/2053ef6a7e16d380a4e33d40059ea987c7373e4c/lib/chain.js#L77-L101

## Are you willing and able to fix this?
This probably requires reworking the async chain stuff, so no.

phil-warner:
Agree. This is making me consider switching framework.

kolbma:
Kind of this also broke the __express middleware__ support, although there has been no usage of async in own code.  
Have tried to change everything in my code to async, but then also there is this compatibility problem with chained __express middleware__.

ghermeto:
Calling `next` **is not allowed** when using `async` functions. That said, I agree Restify must provide _**a way to stop chain processing**_, similarly to `next(false);`.

Right now, all values are being discarded when returning a `AsyncFunction`. Maybe `false` should be passed through.
https://github.com/restify/node-restify/blob/adf24c1046022b1ed2a284556ff0cc9b1777ade2/lib/chain.js#L188-L198

gmahomarf:
> Calling `next` **is not allowed** when using `async` functions. That said, I agree Restify must provide _**a way to stop chain processing**_, similarly to `next(false);`.

Yes, I'm well aware it's **not allowed**. That's the whole point of this issue, as is the inability to stop chain processing.

cjroebuck:
[...]

PR #1937, opened 2022-12-16
fix: destroyed async/await handler

Misyuari:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #1935  

# Changes

Fix unstopped chain-handler when response already destroyed.

### DIFF
--- a/lib/chain.js
+++ b/lib/chain.js
@@ -140,7 +140,7 @@ Chain.prototype.run = function run(req, res, done) {
         var handler = self._stack[index++];
 
         // all done or request closed
-        if (!handler || req.connectionState() === 'close') {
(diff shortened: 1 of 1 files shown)


gmahomarf:
I'm not sure how this fixes #1935. The issue there is that one is no longer allowed to have `async` handlers which also call `next`, due to handler arity checks.

Misyuari:
> I'm not sure how this fixes #1935. The issue there is that one is no longer allowed to have `async` handlers which also call `next`, due to handler arity checks.

oh that was my fault, looks like it's a different issue, I've created a new issue

ghermeto (changes requested):
can you add tests?

Misyuari:
> can you add tests?

done, test added

PR #1942, opened 2023-02-08
stop handling `then` when `next` is called

twksos:
## Pre-Submission Checklist

- [v] Opened an issue discussing these changes before opening the PR
- [v] Ran the linter and tests via `make prepush`
- [v] Included comprehensive and convincing tests for changes

## Issues

Closes:

* Issue #1935

currently, the async function handling does not meet user's expectation.
a function returning Promise can still have the next param and:

- it could get called twice.
- if user send out an error through the next, the promise could still resolve.

# Changes

add a local variable to check whether next is called and wrap the next function.

> What does this PR do?

- in `chain.js` prevent further Promise handling if next is called.
- add tests

### DIFF
--- a/lib/chain.js
+++ b/lib/chain.js
@@ -221,8 +221,12 @@ function call(handler, err, req, res, _next) {
     } else if (!hasError && arity < 4) {
         // request-handling middleware
         process.nextTick(function nextTick() {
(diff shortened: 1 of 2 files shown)


twksos:
Hi @mmarchini, could you review this PR when you have time?

PR #1976, opened 2024-08-07
feat: qs -> neoqs

PuruVJ:
# Changes

Replaces `qs` dependency with `neoqs`

Result:
CleanShot 2024-08-07 at 19 34 35

Graph:

CleanShot 2024-08-07 at 19 35 04

This entire branch will be removed with this PR


wesleytodd:
Hey, me again! I happen to also be on the team at Netflix who maintains restify and have exactly the same concerns here that I presented in `body-parser`. I will not restate them here, but [folks can go read them if they like](https://github.com/expressjs/body-parser/pull/530).

kolbma:
There is the change for usage in lib/plugins/formBodyParser.js missing.

```
diff --git a/lib/plugins/formBodyParser.js b/lib/plugins/formBodyParser.js
index 45731e8..ed19bc5 100644
--- a/lib/plugins/formBodyParser.js
+++ b/lib/plugins/formBodyParser.js
@@ -3,7 +3,7 @@
 'use strict';
 
 var assert = require('assert-plus');
-var querystring = require('qs');
+var qs = require('neoqs/legacy');
 
 var bodyReader = require('./bodyReader');
@@ -42,7 +42,7 @@ function urlEncodedBodyParser(options) {
         }
 
         try {
-            var params = querystring.parse(req.body);
+            var params = qs.parse(req.body);
 
             if (opts.mapParams === true) {
                 var keys = Object.keys(params);
```

PR #1996, opened 2026-06-24
feat: replace deprecated url.resolve() and url.parse() with WHATWG URL

dianager:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #
* Issue #
* Issue #

# Changes

## Deprecated API Fixes
- `url.parse()` / `url.resolve()` — removed in Node.js 24.
- `req.getUrl()` now returns a WHATWG `URL` (semver-major)

## Breaking changes
1. `req.getUrl()` shape changed — no longer a legacy `url.parse()` object. `.query`, `.path`, `.auth`, `.slashes` are gone (silently `undefined`, no error). Use `.searchParams`, `.pathname` + `.search`, `.username`/`.password` instead. Note `.auth`/`.slashes` were already unreliable on the old object for normal (origin-form) requests, so this mainly affects `.query`/`.path` consumers. Also: `.port`/`.host` no longer include an explicitly-sent default port (`:80`/`:443`) the way `url.parse()` did.
2. `req.href()`/`getHref()` is now absolute, not relative (e.g. `http://host/foo?a=1` instead of `/foo?a=1`). 
3. Breaks any code comparing it to a relative path. This also changes the base value used internally by `res.redirect()`, which was rewritten to build the `Location` header via `URL`/`URLSearchParams` instead of `url.format`/`mergeQs` — behavior is preserved (see new redirect tests) but the implementation is new.
4. `req.path()`/`getPath()` now normalizes the path — collapses `../`/`./` segments (converting `\` to `/` was already the case pre-migration, not new). A literal `/foo/../bar` now returns `/bar`. This changes what `lib/plugins/static.js` sees, since it builds file paths via `decodeURIComponent(req.path())` — normalization now runs before decoding, changing traversal-resistance characteristics for that plugin (likely a hardening, but currently untested; worth a dedicated test before relying on it).
5. Malformed Host header can now throw a `TypeError` during routing (previously it never mattered for relative URLs). A **missing** Host header does not throw — it falls back to `localhost`. Highest-severity item — recommend adding Host-header validation middleware before upgrading.
6. `OPTIONS *` requests: `req.getUrl().pathname` now returns `/*` instead of `*` (a bare `*` can't be represented as a URL). Anything checking `req.getUrl().pathname === '*'` should switch to `req.url === '*'`.
7. req.getQuery() - WHATWG URL version auto-encodes special characters (unicode, < > " ') in the raw query string. If an app compares, hashes, or logs the raw query string as text (not parsed), it will now see a different string for the same request. before: req.query() → "a=café",  with this change: req.query() → "a=caf%C3%A9"

 
## Also changed (adjacent cleanup)
- `req.isSecure()` now reads `this.socket.encrypted` instead of `this.connection.encrypted` (same swap in `audit.js`/`throttle.js`). Not URL-parsing related, but feeds into `getUrl()`'s protocol resolution — tests/code that stub `req.connection` instead of `req.socket` will silently misbehave.

Action for consumers: grep for `.getUrl(`, `.href(`, `.getHref(`, `.path(`, `.getPath(`, `.query(`, `.getQuery(` ; update stale `url.Url` type declarations to `URL`; add Host-header validation; re-run tests (this tends to surface as silently wrong values, not crashes).


lib/request.js
```diff
@@ -74,7 +73,7 @@ function patch(Request) {
 
         var protocol = this.isSecure() ? 'https://' : 'http://';
         var hostname = this.headers.host;
-        return url.resolve(protocol + hostname + this.path() + '/', path);
+        return new URL(path, protocol + hostname + this.path() + '/').href;
```
wesleytodd: `new URL` is really slow (relativly). We should cache this call like we do for `getURL`: https://github.com/restify/node-restify/pull/1996/changes#diff-2871801a54601791d9e1004f04b2df5d7adc881560285ff3fa6bee7bf9697857R434

Not saying it needs to be done exactly like either of those, but this is also what `express` does for solving this: https://github.com/pillarjs/parseurl/blob/master/index.js#L43;L54

lib/request.js
```diff
@@ -420,11 +428,25 @@ function patch(Request) {
...
      * @instance
      * @function getUrl
-     * @returns  {Object} url
+     * @returns  {URL} url
      */
     Request.prototype.getUrl = function getUrl() {
         if (this._cacheURL !== this.url) {
-            this._url = url.parse(this.url);
+            var protocol = this.isSecure() ? 'https://' : 'http://';
+            var base = protocol + (this.headers.host || 'localhost');
```
wesleytodd, line 460: This logic is basically the same from `absoluteUri`, if we add the caching to that call I think we can consolidate some of this logic.

lib/response.js
```diff
@@ -768,32 +767,34 @@ function patch(Response) {
...
+            // opt.hostname is documented/used as a bare host (e.g.
+            // 'www.foo.com'), never a full URL, so it can't be fed into
+            // `new URL()` directly - that throws, since a bare host has no
+            // scheme for the URL parser to key off. so when it's set, start
+            // from an empty URL and let it be applied as a plain `.hostname`
+            // assignment below. otherwise, seed finalUri from the current
+            // request (req.href() is always absolute) so we're redirecting
+            // within the same host by default.
+            finalUri = opt.hostname
+                ? new URL('http://localhost/')
```
wesleytodd, line 779: I don't think this is the right way to go about this. While `new URL` does require a host, defaulting to `localhost` is almost certainly incorrect behavior for this api. My guess is that a better api would be to:

1. require a hostname
2. use the incoming request host as the fallback

I am going to move on from this for now because I don't have enough time to finish a review and also dig deep on this one, but we for sure need to address this part before merging.

wesleytodd (changes requested):
I am going to drop this here because I have held this up by inaction too long already, but I am pretty sure that the redirect changes can be *way* smaller and thus make this PR easier to undertand and ensure it doesn't introduce new security or behavior regressions. I would suggest taking a look at what fastify (uses URL) and express (still uses url.parse)) are doing for this and see if there is something you can copy directly.

lib/response.js
dianager: Hey @wesleytodd .

1. "Require a hostname" — I don't think it will work.
This branch only runs when opt.hostname is already set (opt.hostname ? ... : ...), so there's nothing to "require" — it's optional by design, and that's intentional: res.redirect({pathname: '/other'}) with no hostname is a valid, common call that must keep working.

2. "Use the incoming request host as the fallback" — No.
Basing the new URL on the current request would carry over the current request's query string and hash to the new hostname unless something explicitly strips them — leaking data (e.g. ?token=...) to a different host on a plain hostname-only redirect. That's worse than the current behavior.


I invoke **new URL()** with opt.hostname and then manually set **finalUri.hostname** as **opt.hostname**. If I call _new URL(opt.hostname)_ it will fail. According to the comments ot.hostname has the format www.foo.com. new URL treats this as invalid host. This workaround helps to build final url correctly

wesleytodd: For number 2 I was thinking of *just* using the `hostname` from the incoming request. The assumption here is that the developer, if they allowed or passed in a path only string, that it would mean they intend to redirect *within the current domain/host*. Does that make sense?

dianager: currentFullPath will contain `hostname` from the incoming request. `var currentFullPath = req.href();`
Now href is absolute path and will contain the host.

wesleytodd (approved):
As long as you and @jcreighton are good with the one comment I had on the fallback, I don't want to block on this any longer since I don't have enough time right now to dig in more deeply. Approving to make sure I am not holding anything up for lack of time to review.

PR #2001, opened 2026-06-24
feat: replace module.parent with explicit module param in test helper

dianager:
## Pre-Submission Checklist

## Issues

Closes:

* Issue #
* Issue #
* Issue #

# Changes

## Deprecated API Fixes
- `module.parent` — removed in Node.js 26. Replaced with `require.main === module`.
  Affected: `benchmark/benchmarks/*.js`, `test/lib/helper.js`.

### DIFF
--- a/examples/todoapp/lib/server.js
+++ b/examples/todoapp/lib/server.js
@@ -278,7 +278,10 @@ function listTodos(req, res, next) {
  */
 function putTodo(req, res, next) {
     if (!req.body.task) {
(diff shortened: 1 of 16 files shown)


wesleytodd (approved):
Might be worth calling out that this commit also includes a bunch of unrelated lint fixes. Not a problem, but sometimes it is nice to add that to the commit message in case folks are doing a `git blame` in the future.

PR #2005, opened 2026-08-10
fix: npm publish workflow

Senseye:
Adds `npm-publish.yml` workflow for enabling npm trusted publishing.
Updated `replease-please` to latest version.
Updated docs.

Notes:

- [x] "Publish" environment created and configured
- [x] Trusted Publisher configs on npm side

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -9,6 +9,8 @@ permissions:
   contents: read
 
 concurrency:
+  # tag_name grouping - queues actual releases for the same tag
+  # run_id - fallback value required for group as tag_name might not always be present (manual triggers)
   group: npm-publish-${{ github.event.release.tag_name || github.run_id }}
   cancel-in-progress: false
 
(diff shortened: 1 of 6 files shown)


.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,84 @@
...
+    name: publish
+    needs: validate
+    if: github.event_name == 'release'
+    runs-on: ubuntu-latest
+    environment: Publish
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - uses: actions/checkout@v7
```
dianager, line 75: @Senseye  Do you have the chance to verify that this yml works?

Senseye: Good question @dianager 
I can confirm the validate step works - just tested it via https://github.com/nektos/act

And `publish` is just standard stuff. It should work fine.

dianager: I will approve this. But would be great it @wesleytodd  also take a look

dianager (approved):

